PR #71, opened 2023-02-15
Julian Beldotti

JCBeldo:


JCBeldo:
My atttempt

lib/facility.rb
```diff
@@ -1,14 +1,72 @@
...
+  def register_vehicle(car)
+    if @services.include?('Vehicle Registration')
+      car.registration_date = Date.today
+      car.plate_type
+      collect_fees(car)
+      @registered_vehicles.push(car)
+    end
+  end
+
+  def administer_written_test(registrant)
```
dcoleman21, line 43: For this method, the written test has to be a service provided and a registrant has to be at least 16 and a registrant has to have a permit. Here you are using the `||` operator and I think you would want to use the `&&` operator that checks the above mentioned conditions. So something like: 

```
if @services.include?('Written Test') && registrant.permit == true && registrant.age >= 16
else
  false
end
```

lib/facility.rb
```diff
@@ -1,14 +1,72 @@
...
+    if @services.include?('Written Test') == false
+      false
+    elsif registrant.age < 16 || registrant.permit? != true
+      false
+    else
+      registrant.license_data[:written] = true
+    end
+  end
+  
+  def administer_road_test(registrant)
```
dcoleman21, line 53: Look at my comment for `administer_written_test`. I think this method could follow that same sort of logic flow.

lib/facility.rb
```diff
@@ -1,14 +1,72 @@
...
+    if @services.include?('Road Test') == false
+      false
+    elsif registrant.license_data[:written] != true
+      false
+    else
+      registrant.license_data[:license] = true
+    end
+  end
+  
+  def renew_drivers_license(registrant)
```
dcoleman21, line 63: Look at my comment for `administer_written_test`. I think this method could follow that same sort of logic flow.

lib/registrant.rb
```diff
@@ -0,0 +1,21 @@
+class Registrant
+  attr_accessor :permit
+  attr_reader   :name,
+                :age,
+                :license_data
+              
+  def initialize(name, age, permit = false)
+  @name = name
```
dcoleman21, line 8: Watch your indentation here.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,25 @@
...
+  attr_accessor :registration_date
+  attr_reader   :vehicles
+
+  def initialize
+    @vehicles = []
+  end
+  
+  def create_vehicles(registrations)
+    registrations.map do |vehicle|
+      @vehicles << vehicle = Vehicle.new(registrations)
```
dcoleman21, line 11: Looks like you were in the process of building this out. Nice use of `.map`. So you probably won't want to create the `vehicle` variable here. You can instead just shovel `Vehicle.new` into your `@vehicles` array. 
So if you were to look at your `vehicle_spec.rb` you will see that when a new `vehicle` object is created, it does indeed only take 1 argument but that argument is a hash. So you will want to replicate that in this method. So something like so:

```
registrations.map do |vehicle|
    @vehicles << Vehicle.new({
        engine: all_electric,
        make: vehicle[:make],
        model: vehicle[:model],
        vin: vehicle[:vin_1_10],
        year: vehicle[:model_year]
    })
```

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,25 @@
...
+
+require 'pry'; binding.pry
```
dcoleman21, line 13: make sure to at least comment these out before submitting.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,25 @@
...
+  
+  def create_vehicles(registrations)
+    registrations.map do |vehicle|
+      @vehicles << vehicle = Vehicle.new(registrations)
+
+require 'pry'; binding.pry  
+    end
+  end
+
+  def all_electric
```
dcoleman21, line 17: Nice helper for your above method

lib/vehicle.rb
```diff
@@ -17,9 +18,18 @@ def initialize(vehicle_details)
 
   def antique?
     Date.today.year - @year > 25
-  end
-
+
```
dcoleman21, line 21: You are missing the `end` to your `def antique?` method

spec/facility_spec.rb
```diff
@@ -23,4 +25,229 @@
...
+      expect(facility_2).to be_a(Facility)
+    end
+
+    it 'can register a vehicle' do
+      facility_1 = Facility.new({name: 'Albany DMV Office', address: '2242 Santiam Hwy SE Albany OR 97321', phone: '541-967-2014' })
+      facility_2 = Facility.new({name: 'Ashland DMV Office', address: '600 Tolman Creek Rd Ashland OR 97520', phone: '541-776-6092' })
+      cruz = Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice} )
+      bolt = Vehicle.new({vin: '987654321abcdefgh', year: 2019, make: 'Chevrolet', model: 'Bolt', engine: :ev} )
+      camaro = Vehicle.new({vin: '1a2b3c4d5e6f', year: 1969, make: 'Chevrolet', model: 'Camaro', engine: :ice} )
+
```
dcoleman21, line 44: This test is named `can register a vehicle` but I don't see anywhere in this test your `register_vehicle` method that show that a facility can register a vehicle and then check its `registered_vehicles`.

spec/facility_spec.rb
```diff
@@ -23,4 +25,229 @@
...
+    end
+    
+    it 'adds a new service "vehicle registration"' do
+      facility_1 = Facility.new({name: 'Albany DMV Office', address: '2242 Santiam Hwy SE Albany OR 97321', phone: '541-967-2014' })
+      
+      expect(facility_1.add_service('Vehicle Registration')).to eq(["Vehicle Registration"])
+      expect(facility_1.services).to eq(['Vehicle Registration'])
+    end
+    
+    it 'can collect fees' do
```
dcoleman21, line 57: So in order to test this more robustly, you would need to start with your `collected_fees` as zero, then register a vehicle and check `collected_fees` again.

spec/facility_spec.rb
```diff
@@ -23,4 +25,229 @@
...
+
+      expect(facility_1.register_vehicle(camaro)).to eq([camaro])
+      expect(facility_1.registered_vehicles).to eq([camaro])
+      expect(camaro.plate_type).to eq(:antique)
+      expect(facility_1.collected_fees).to eq(25)
+    end
+  end
+
+  describe "registrants can apply for services" do
+    it "shows license data for registrant_1" do
```
dcoleman21, line 124: We are in the `facility_spec` so you will want most of your `expect statements` to be checking `facility vs registrant`. You will want to check your` registrants` in the `registrant_spec`.

spec/facility_spec.rb
```diff
@@ -1,4 +1,6 @@
 require 'spec_helper'
+require './lib/facility'
+require './lib/vehicle'
 
 RSpec.describe Facility do
   before(:each) do
```
dcoleman21, line 6: I would review all of your test in this spec file. Since we are in the `facility_spec` we want our `expect statements` to check for `facility` things. All of your `expect statements` for `registrant` will want to be in the `registrant_spec`. This iteration would require you to be making tests in both files.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+    
+    it 'earned a permit?' do
+      registrant_2 = Registrant.new('Penny', 15)
+      registrant_2.earn_permit
+      
+      expect(registrant_2.permit?).to eq(true)
+    end
+  end
+end
```
dcoleman21, line 61: All of your above tests are great for Iteration 1. However, when we get to Iteration 2, where a facility is providing `written, road, and renew` you will need/want to add tests for updating a `registrants` license_data, permit based on whether or not they completed the services mentioned. This is how we use integration and unit tests. This may also require you to build additional methods to support these changes.

dcoleman21 (commented):
Overall nice job working through Iteration 2. I have left some comments for you to review. Please let me know if you have any questions. 
I would recommend continuing to work on organizing your tests better. In your facility spec your have some tests that aren't testing what you are saying they are testing. You will also need to test any/all helper methods that you create (those are considered unit tests).

PR #130, opened 2023-04-04
Michael Callahan

calforcal:


lib/registrant.rb
```diff
@@ -0,0 +1,22 @@
...
+  attr_reader :name, :age, :permit_status, :license_data
+
+  def initialize(name, age, permit_status = false)
+    @name = name
+    @age = age
+    @permit_status = permit_status
+    @license_data = {
+      :written => false,
+      :license => false,
+      :renewed => false
```
epintozzi, line 11:
```suggestion
      written: false,
      license: false,
      renewed: false
```

This syntax is a bit more conventional for ruby hashes

epintozzi (commented):
Great work! Overall your testing looks good and thorough, so I have no individual comments there on purpose.

PR #131, opened 2023-04-04
Steven Stackhouse

stackmm:


lib/vehicle_factory.rb
```diff
@@ -0,0 +1,15 @@
...
+
+class VehicleFactory
+  def create_vehicles(data)
+    data.map do |car|
+      Vehicle.new({
+        vin: car[:vin_1_10],
+        year: car[:model_year].to_i,
+        make: car[:make],
+        model: car[:model],
+        engine: :ev
```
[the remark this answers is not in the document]
KatBrandt: The other thing to think about here, is a lot of the keys from the `car` hash you have in your block variable, already have the keys that your Vehicle class is looking for to initialize. You could just send in Vehicle.new(car) and then the Vehicle class would use what you have in initialize to pull out the info it needed. This wouldn't work for the vin since the car hash has the key as :vin_1_10, but you could add in the :vin key you needed:
car[:vin] = car[:vin_1_10] which would add that key with the correct value, and then you just send all that data, as is, to the Vehicle class. Hoping this all makes sense, reach out if it doesn't.

lib/registrant.rb
```diff
@@ -0,0 +1,35 @@
+class Registrant
+  attr_reader :name, :age, :permit, :license_data
```
KatBrandt, line 2: Careful here, only make attr_readers for the methods you want to expose. We don't need the method `#permit` since we have `#permit?`

lib/registrant.rb
```diff
@@ -0,0 +1,35 @@
...
+  
+  def permit?
+    @permit
+  end
+
+  def earn_permit
+    @permit = true
+  end
+
+  def written?
```
KatBrandt, line 23: Interesting choice to break these apart into their own methods. I like it! I will peek in the tests to make sure all of these helper methods have their own tests. :)

spec/registrant_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    end
+
+    it 'leaves permit status unchanged if previously earned' do
+      expect(@registrant_1.permit?).to be true
+      @registrant_1.earn_permit
+      expect(@registrant_1.permit?).to be true
+    end
+  end
+
+  describe '#written?' do
```
KatBrandt, line 64: woohoo!!! Good job adding tests for all your helper methods. You don't need to test both instances though, especially since they're both false, testing one is fine.

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    it 'can initialize' do
+      factory = FacilityFactory.new
+      expect(factory).to be_an_instance_of(FacilityFactory)
+    end
+  end
+
+  describe '#create_facility_oregon' do
+    it 'can create facilities from Oregon DMV office data' do
+      expect(@oregon_facilities).to be_a(Array)
+      expect(@oregon_facilities.count).to eq(59)
```
KatBrandt, line 23: For now. This would break if the API we're pulling the data from added a facility. Makes our tests a bit brittle because they are dependent on data we don't control.

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    end
+
+    it 'can find the name of DMV offices in Oregon' do
+      expect(@oregon_facilities[0].name).to eq("Albany DMV Office")
+      expect(@oregon_facilities[@oregon_facilities.count - 1].name).to eq("La Pine DMV Office")
```
KatBrandt, line 28: These tests are also very brittle because as soon as the api makes changes, these would all fail. We want to be testing the code we write, independent of the actual data which we don't have control over. It's a tough balance!

spec/facility_spec.rb
```diff
@@ -2,25 +2,201 @@
...
+
   describe '#initialize' do
     it 'can initialize' do
-      expect(@facility).to be_an_instance_of(Facility)
-      expect(@facility.name).to eq('Albany DMV Office')
-      expect(@facility.address).to eq('2242 Santiam Hwy SE Albany OR 97321')
-      expect(@facility.phone).to eq('541-967-2014')
-      expect(@facility.services).to eq([])
+      expect(@facility_1).to be_an_instance_of(Facility)
+      expect(@facility_2).to be_an_instance_of(Facility)
```
KatBrandt, line 18: Redundant testing. Once we know the .new works for one instance, we can safely assume it works for all instances.

KatBrandt: All these attrbitutes can also be tested for just one instance.

spec/facility_spec.rb
```diff
@@ -2,25 +2,201 @@
...
-      @facility.add_service('Renew Drivers License')
-      @facility.add_service('Vehicle Registration')
-      expect(@facility.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
+      expect(@facility_1.services).to eq([])
+      @facility_1.add_service('New Drivers License')
+      expect(@facility_1.services).to eq(['New Drivers License'])
+      @facility_1.add_service('Renew Drivers License')
+      expect(@facility_1.services).to eq(['New Drivers License', 'Renew Drivers License'])
+      @facility_1.add_service('Vehicle Registration')
+      expect(@facility_1.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
```
KatBrandt, line 38: Good job testing that the initial value was an empty array, then adding to it, then testing it again to ensure it changed.

spec/facility_spec.rb
```diff
@@ -2,25 +2,201 @@
...
+      expect(@camaro.registration_date).to eq(Date.today)
+      expect(@camaro.plate_type).to eq(:antique)
+      expect(@facility_1.registered_vehicles).to eq([@cruz, @camaro])
+      expect(@facility_1.collected_fees).to eq(125)
+
+      @facility_1.register_vehicle(@bolt)
+      expect(@bolt.registration_date).to eq(Date.today)
+      expect(@bolt.plate_type).to eq(:ev)
+      expect(@facility_1.registered_vehicles).to eq([@cruz, @camaro, @bolt])
+      expect(@facility_1.collected_fees).to eq(325)
```
KatBrandt, line 65: This is good testing because each of these cars has a different plate type so you want to test each situation

spec/facility_spec.rb
```diff
@@ -2,25 +2,201 @@
...
+    it 'does nothing if there is no Vehicle Registration service available' do
+      expect(@facility_1.services.include?('Vehicle Registration')).to be false
+      expect(@facility_1.registered_vehicles).to eq([])
+      @facility_1.register_vehicle(@cruz)
+      expect(@facility_1.registered_vehicles).to eq([])
+    end
+  end
+
+  describe '#collect_registration_fee' do
+    it 'can collect registration fee' do
```
KatBrandt, line 77: Great job with this unit testing.

spec/facility_spec.rb
```diff
@@ -2,25 +2,201 @@
...
+      expect(@facility_1.collected_fees).to eq(100)
+      @facility_1.collect_registration_fee(@bolt)
+      expect(@facility_1.collected_fees).to eq(300)
+      @facility_1.collect_registration_fee(@camaro)
+      expect(@facility_1.collected_fees).to eq(325)
+    end
+  end
+
+  describe '#administer_written_test' do
+    it 'can administer written test if age >= 16 and has a permit' do
```
KatBrandt, line 90: Your test names are wonderful! Short, succinct, yet very accurate of what the test is looking for

spec/registrant_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    end
+
+    it 'can set age' do
+      expect(@registrant_1.age).to eq(18)
+      expect(@registrant_2.age).to eq(15)
+    end
+
+    it 'can set permit status' do
+      expect(@registrant_1.permit).to be true
+      expect(@registrant_2.permit).to be false
```
KatBrandt, line 27: For this one I _would_ test this for both registrants since we're expecting different outcomes, one with the default value and one being set.

KatBrandt (commented):
Overall you're doing a great job! Your tests are thorough and well documented (labeling), looks like you have a good grasp of enumerables and flow control. I would like to see you use more branches in your development though. One branch per larger feature you're working on

PR #132, opened 2023-04-04
Austin Woods

boomclear:


README.md
```diff
@@ -1,3 +1,15 @@
 # The DMV
 
 This is the starter repo for the BE Mod1 DMV project.
+
+Describe the steps you took to dig in to this code base. What was your process? If you don’t feel you had a process, define one that you might like to try next time.
+ 
+The first step I took was to look at the code and see what was going on. I then compared the code to the tests to get a sense of what was intentioned. After that I ran the tests to confirm that the existing code worked. I then fixed the code for the tests to pass.
```
KatBrandt, line 7: This is such a great strategy. The tests act as documentation for the code so that's a great place to start.

README.md
```diff
@@ -1,3 +1,15 @@
...
+ 
+The first step I took was to look at the code and see what was going on. I then compared the code to the tests to get a sense of what was intentioned. After that I ran the tests to confirm that the existing code worked. I then fixed the code for the tests to pass. 
+
+What was hard about working with code you did not write?
+
+It was formatted in a different way that I personally code with. It was also different because I didn't know the original intention behind the code so finding its purpose took longer.
+
+What was easier than you expected about jumping in to an unfamiliar codebase? What made it easy? If nothing felt easy, what would’ve helped you feel more comfortable more quickly?
+
+Being able to fix the problems was easier than I originally expected it to be. I know this is because there were good tests already in place but those made it easier to edit the code and fix it.
```
KatBrandt, line 15: Shout out to having good tests! :)

lib/dmv.rb
```diff
@@ -9,8 +9,8 @@ def add_facility(facility)
   end
 
   def facilities_offering_service(service)
-    @facilities.find do |facility|
-      facility.services.include?(service)
+      @facilities.find_all do |facility|
```
KatBrandt, line 12: Great use of find_all

lib/facility.rb
```diff
@@ -1,14 +1,61 @@
...
+    @registered_vehicles = []
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def administer_written_test(person)
+    if @services.include?("Written Test") && person.permit? && person.age >= 16
```
KatBrandt, line 18: Great use of your boolean logic here

lib/facility.rb
```diff
@@ -1,14 +1,61 @@
...
+      false
+    end
+  end
+
+  def register_vehicle(car)
+    if @services.include?("Vehicle Registration")
+      @registered_vehicles << car
+      car.registration_date = Date.today
+      car.plate_type!
+      if car.plate_type == :regular
```
KatBrandt, line 49: In the future, consider moving this logic into its own helper method. Maybe it's a method called `collect_fees` or something, that the register vehicle method could call. When thinking about Single Responsibility Principle, we want each method to do one thing. Currently, register_vehicle is registering a vehicle, and changing the plate, and collecting fees. If we can break these sections of logic into their own helper methods that register_vehicle calls, then we've separated out the responsibilities.

lib/facility_factory.rb
```diff
@@ -0,0 +1,42 @@
+class FacilityFactory
+
+  def initialize
```
KatBrandt, line 3: If nothing is in the initialize method, you can omit it.

lib/facility_factory.rb
```diff
@@ -0,0 +1,42 @@
...
+  def create_facilities(facilities)
+    if facilities[0].keys.include?(:location_1)
+      facilities.map do |facility|
+        location = JSON.parse(facility[:location_1][:human_address].gsub('\"', ''))
+        facility_details = {
+          name: facility[:title],
+          address: location.values.join(' '),
+          phone: facility[:phone_number],
+        }
+        Facility.new(facility_details)
```
KatBrandt, line 15: However, what you could store in the initalize method is an attribute `@or_facilities` and another for each set, that way these methods only have to run once and then all the data is stored in that attribute, instead of having these enumerables run everytime we need the data set. 

I would also have separated each of your else blocks into separate methods. Maybe a or_facilities method and a wa_facilities method. And then your if/else blocks could just call those methods depending on which logic path the code went down.

KatBrandt: Great job working with all the different data sets here!

lib/vehicle.rb
```diff
@@ -6,20 +6,37 @@ class Vehicle
...
+
+  def plate_type!
+    if antique?
+      @plate_type = :antique
+    elsif regular?
+      @plate_type = :regular
+    else
+      @plate_type = :ev
+    end
+  end
```
KatBrandt, line 41: Great job with this method. I like your use of helper methods in the if conditionals.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,16 @@
...
+    vehicles.map do |vehicle|
+      vehicle_details = {
+        vin: vehicle[:vin_1_10],
+        year: vehicle[:model_year],
+        make: vehicle[:make],
+        model: vehicle[:model],
+        engine: :ev,
+        county: vehicle[:county]
+      }
+      Vehicle.new(vehicle_details)
```
KatBrandt, line 13: Same thing here as the FacilitiesFactory. Would be helpful to have an instance variable declared in initialize that you could feed all these instances into so you only have to run this map once. Great job with the map!

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  it 'creates Facilities' do
+    factory = FacilityFactory.new
+
+    or_dmv_facilities = DmvDataService.new.or_dmv_office_locations
+    or_facilities = factory.create_facilities(or_dmv_facilities)
+
+    expect(or_facilities).to be_an(Array)
+    expect(or_facilities[0]).to be_a(Facility)
```
KatBrandt, line 19: Yup, great job with this testing. We can't test the content because that could change since we don't own or control that data.

spec/facility_spec.rb
```diff
@@ -23,4 +26,155 @@
...
+    facility_1 = Facility.new({name: 'Albany DMV Office', address: '2242 Santiam Hwy SE Albany OR 97321', phone: '541-967-2014' })
+    facility_2 = Facility.new({name: 'Ashland DMV Office', address: '600 Tolman Creek Rd Ashland OR 97520', phone: '541-776-6092' })
+
+    cruz = Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice} )
+    bolt = Vehicle.new({vin: '987654321abcdefgh', year: 2019, make: 'Chevrolet', model: 'Bolt', engine: :ev} )
+    camaro = Vehicle.new({vin: '1a2b3c4d5e6f', year: 1969, make: 'Chevrolet', model: 'Camaro', engine: :ice} )
+
+    facility_1.add_service('Vehicle Registration')
+    facility_1.register_vehicle(cruz)
+    expect(facility_1.registered_vehicles).to eq [cruz]
```
KatBrandt, line 80: How is this test different than the ones on lines 50-54? I'm not sure we're testing anything new here. Oh... I see, ok, actually, I would have put this assertion on line 80, on line 55 to ensure that the vehicle was registered.

spec/facility_spec.rb
```diff
@@ -23,4 +26,155 @@
...
+    camaro = Vehicle.new({vin: '1a2b3c4d5e6f', year: 1969, make: 'Chevrolet', model: 'Camaro', engine: :ice} )
+
+    facility_1.add_service('Vehicle Registration')
+    facility_1.register_vehicle(cruz)
+    expect(facility_1.collected_fees).to eq 100
+    facility_1.register_vehicle(camaro)
+    facility_1.register_vehicle(bolt)
+
+    expect(facility_1.registered_vehicles).to eq [cruz, camaro, bolt]
+    expect(facility_1.collected_fees).to eq 325
```
KatBrandt, line 98: This is a good flow of these tests. Would just want to start with the initial value in this test as well.

spec/facility_spec.rb
```diff
@@ -23,4 +26,155 @@
...
+      registrant_1 = Registrant.new('Bruce', 18, true )
+      registrant_2 = Registrant.new('Penny', 16 )
+      registrant_3 = Registrant.new('Tucker', 15 )
+      facility_1 = Facility.new({name: 'Albany DMV Office', address: '2242 Santiam Hwy SE Albany OR 97321', phone: '541-967-2014' })
+      facility_2 = Facility.new({name: 'Ashland DMV Office', address: '600 Tolman Creek Rd Ashland OR 97520', phone: '541-776-6092' })
+
+      facility_1.add_service('Written Test')
+      expect(facility_1.administer_written_test(registrant_2)).to eq false
+      registrant_2.earn_permit
+      expect(facility_1.administer_written_test(registrant_2)).to eq true
```
KatBrandt, line 135: Great test showing the initial value and then the changed value.

lib/vehicle.rb
```diff
@@ -6,20 +6,37 @@ class Vehicle
...
 
   def antique?
     Date.today.year - @year > 25
   end
 
+  def regular?
+    @engine == :ice
+  end
+
   def electric_vehicle?
```
KatBrandt, line 29: Good job testing your helper methods as well!

lib/vehicle.rb
```diff
@@ -6,20 +6,37 @@ class Vehicle
...
     @engine == :ev
   end
+
+  def plate_type!
```
KatBrandt, line 33: I don't see a test in the vehicle_spec.rb for this method. Remember that every method should be tested in the test for the class.

KatBrandt (commented):
Overall, great job! In terms of your version control, good job using branches. Moving forward, lets name the branches indicative of the work they are encompassing. Something like "Create Facility Factory" and then all the commits on that branch would related to the tests and code. Nice work!

PR #133, opened 2023-04-04
Nick Tassinari!

NickTassinari:


spec/facility_spec.rb
```diff
@@ -17,10 +17,64 @@
...
+      bolt = Vehicle.new({vin: '987654321abcdefgh', year: 2019, make: 'Chevrolet', model: 'Bolt', engine: :ev} )
+
+      expect(facility_2.registered_vehicles).to eq([])
+      expect(facility_2.services).to eq([])
+  
+      facility_2.register_vehicle(bolt)
+
+      expect(facility_2.registered_vehicles).to eq([])
+      expect(facility_2.collected_fees).to eq(0)
+    end
```
dcoleman21, line 79: So in your `Facility` class you have `administer_written_test`, `administer_road_test`, and `renew_drivers_license` methods. I know you have those tested in your `Registrant` class but they would need to live here because this is where those methods live.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+    expect(registrant_2.name).to eq("Penny")
+    expect(registrant_2.age).to eq(15)
+    expect(registrant_2.permit?).to eq(false)
+    expect(registrant_2.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+    
+    registrant_2.earn_permit 
+    expect(registrant_2.permit?).to eq(true)
+  end
+
+  it 'can issue written test' do
```
dcoleman21, line 35: These are the tests that would live in your `facility_spec` because the methods are written in the `Facility` class.  So we would need to Unit test them there.

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+require './lib/dmv_data_service'
+require 'spec_helper'
+
+RSpec.describe VehicleFactory do 
+
+  it 'exists' do 
+    @factory = VehicleFactory.new 
+    @wa_ev_registrations = DmvDataService.new.wa_ev_registrations
+    
+    expect(@factory).to be_a(VehicleFactory)
```
dcoleman21, line 12: I would include an expect statement for
```ruby
expect(@wa_ev_registrations).to be_an(Array)
```

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    expect(@factory.vehicles).to eq([])
+    @factory.create_vehicles(@wa_ev_registrations)
+    expect(@factory.vehicles.first).to be_a(Vehicle)
+    expect(@factory.vehicles.first.vin).to be_truthy 
+    expect(@factory.vehicles.first.year).to be_truthy 
+    expect(@factory.vehicles.first.make).to be_truthy 
+    expect(@factory.vehicles.first.model).to be_truthy 
+  end
+
+  it "can create many vehicles" do
```
dcoleman21, line 28: I think these two tests are doing essentially the same thing.

dcoleman21 (commented):
Nick, overall nice job! This one was a little tricky in regards to where you might write your tests. You had the tests for the written, road, and renew in your registrant_spec and they should live in your facility_spec because the Facility is where those methods are written. I can show you how you could potentially split them into both the registrant_spec and vehicle_spec. Also you wrote a method for `plate_type` but you didn't write a test for it. 
Keep working on cleaning up your white spaces and making sure your def/ends align. You don't want white space at the beginning or end of your tests. Let me know if you have any questions.

PR #135, opened 2023-04-04
Kim Bergstrom - The DMV Project

kbergstrom78:


README.md
```diff
@@ -1,3 +1,13 @@
-# The DMV
 
-This is the starter repo for the BE Mod1 DMV project.
+# Reflection
```
dcoleman21, line 2: Really great reflections.

lib/facility.rb
```diff
@@ -1,14 +1,77 @@
...
+    @registered_vehicles = []
+    @collected_fees = 0
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+ 
+  def register_vehicle(vehicle)
```
dcoleman21, line 23: Nice flow here, just watch your indentations.

lib/facility.rb
```diff
@@ -1,14 +1,77 @@
...
+  end
+
+
+  def administer_written_test(registrant)
+    if @services.include?('Written Test') && 
+      registrant.permit? &&
+      registrant.age >= 16
+      
+      registrant.license_data[:written] = true
+        return true
```
dcoleman21, line 45: You don't need to use the keyword `return`.

lib/facility.rb
```diff
@@ -1,14 +1,77 @@
...
+      end
```
dcoleman21, line 46: this should be an `else` vs `end`.

lib/facility.rb
```diff
@@ -1,14 +1,77 @@
...
+      false
+
```
dcoleman21, line 48: your are missing two `ends`.

lib/facility.rb
```diff
@@ -1,14 +1,77 @@
...
+      end
+      false
+  
+  def administer_written_test(registrant)
+   if @services.include?('Written Test') && 
+     registrant.permit? &&
+     registrant.age >= 16
+     registrant.license_data[:written] = true
+        return true
+     end
```
dcoleman21, line 55: Make this an `else` and add an `end` to line 56 1/2

lib/facility.rb
```diff
@@ -1,14 +1,77 @@
...
+     false
+  end
+  
+  def administer_road_test(registrant)
+    if @services.include?('Road Test') &&
+      registrant.license_data[:written]
+      registrant.license_data[:license] = true
+      return true
+    end
```
dcoleman21, line 64: Make this and `else` and add and `end` to line 65 1/2

lib/facility.rb
```diff
@@ -1,14 +1,77 @@
...
+    false
+  end
+
+  def renew_drivers_license(registrant)
+    if @services.include?('Renew License') &&
+      registrant.license_data[:license]
+      registrant.license_data[:renewed] = true
+      return true
+    end
```
dcoleman21, line 73: Make this and `else` and add an `end` to line 74 1/2

lib/facility_factory.rb
```diff
@@ -0,0 +1,62 @@
...
+  def initialize
+    @new_facilities = []
+  end
+   
+  
+  def checks_location(location_data)
+      dmv_office_locations(location_data)
+  end
+  
+  def dmv_office_locations(location_data)
```
dcoleman21, line 14: This is an awfully large method. A good refactor would be to see what you can pull out into helper methods to make things more dynamic. What if you needed to pull data from 20 more offices? 

So think about having one method that is responsible for creating new `Facility` objects. Then build helper methods that can format things like name, address, phone_number and use those inside your main method that is creating the new objects. Just remember, you have to test EVERY method you create. 
Currently you are not testing this method. You are only testing `checks_location`.

lib/facility_factory.rb
```diff
@@ -0,0 +1,62 @@
+class FacilityFactory
+  attr_reader :new_facilities
+              :dmv_office_locations
+
+  def initialize
+    @new_facilities = []
```
dcoleman21, line 6: You need to test this. But would you even really need it if you created a method that returned an array of new Facility objects?

lib/registrant.rb
```diff
@@ -0,0 +1,20 @@
...
+    @permit = permit
+    @license_data = {:written=>false, :license=>false, :renewed=>false}
+  end
+
+  def permit?
+    @permit
+  end
+
+  def earn_permit
+    @permit = true
```
dcoleman21, line 19: You could make this even more robust by adding a check `@permit = true if @age >= 15`.

lib/vehicle.rb
```diff
@@ -6,20 +6,35 @@ class Vehicle
...
+                  :plate_type
+ 
   def initialize(vehicle_details)
     @vin = vehicle_details[:vin]
     @year = vehicle_details[:year]
     @make = vehicle_details[:make]
     @model = vehicle_details[:model]
     @engine = vehicle_details[:engine]
+    @registration_date = nil
+    @plate_type = nil
```
dcoleman21, line 20: You never reference this anywhere but here. You built a method called `plate_type` below and that is the more dynamic way to go about it.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,28 @@
+class VehicleFactory
+  attr_reader :created_vehicles 
+  
+  def initialize
+    @created_vehicles = []
+  end
+
+  def create_vehicles(vehicle_details)
```
dcoleman21, line 8: This method should be responsible for creating new `Vehicle` objects. I am assuming that `vehicle_details` is an array of data. So you can just iterate over this and instantiate a new `Vehicle` and pass it an argument that is a hash with the attributes that you for each vehicle object. 
```ruby
def create_vehicles(vehicle_details)
   vehicle_details.map do |vehicle|
     Vehicle.new({
       engine: :ev,
       make: vehicle[:make],
       model: vehicle[:model],
       vin: vehicle[:vin_1_10],
       year: vehicle[:model_year]
     })
end
```
This returns a new array.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,28 @@
...
+    wa_ev_registrations = DmvDataService.new.wa_ev_registrations
+    wa_registrations(vehicle_details)
+    @created_vehicles
+  end 
+  
+   # Helper method
+  def wa_registrations(vehicle_details)
```
dcoleman21, line 15: This is not tested currently. Think about other enumerables that return an array of the same length.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,28 @@
+class VehicleFactory
+  attr_reader :created_vehicles 
+  
+  def initialize
```
dcoleman21, line 4: I understand why you create this `initialize` method because you created an instance variable to hold the newly created objects but it is not necessary. 
```ruby
factory = VehicleFactory.new 
```
Has no arguments. So see my comments below on how you can just use the `create_vehicles` method to create new `Vehicle` objects without having to shovel them into a collector.

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    office = FacilityFactory.new 
+  
+    expect(office).to be_an_instance_of(FacilityFactory)
+  end
+
+  describe 'oregon_offices' do
+    it 'has oregon office info' do
+      office = FacilityFactory.new
+      or_dmv_office_locations = DmvDataService.new.or_dmv_office_locations
+      office = office.checks_location(or_dmv_office_locations)
```
dcoleman21, line 17: Be careful with your variable naming. You have `office` on line 15 and are using it again here on line 17. To make it easier on yourself you could just do 
```ruby
expect(office.checks_locations(or_dmv_office_locations)).to be_an(Array)
```
Also is this method really checking the locations or is it creating new `Facility` objects??

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      office = FacilityFactory.new
+      mo_dmv_office_locations = DmvDataService.new.mo_dmv_office_locations
+      office = office.checks_location(mo_dmv_office_locations)
+
+      expect(office).to be_a(Array)
+      expect(office.first).to be_a(Facility)
+      expect(office.length).to eq(177)
+    end
+  end   
+
```
dcoleman21, line 48: Make sure you get rid of all the extra white space.

spec/facility_spec.rb
```diff
@@ -15,12 +19,257 @@
...
+      facility_2 = Facility.new({name: 'Ashland DMV Office', address: '600 Tolman Creek Rd Ashland OR 97520', phone: '541-776-6092' })
+
+      expect(facility_1).to be_an_instance_of(Facility)
+      expect(facility_1).to be_an_instance_of(Facility)
+
+      cruz = Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice} )
+      bolt = Vehicle.new({vin: '987654321abcdefgh', year: 2019, make: 'Chevrolet', model: 'Bolt', engine: :ev} )
+      camaro = Vehicle.new({vin: '1a2b3c4d5e6f', year: 1969, make: 'Chevrolet', model: 'Camaro', engine: :ice} )
+
+      expect(cruz).to be_an_instance_of(Vehicle)
```
dcoleman21, line 43: You don't have to test these here because they are already being tested in the `Vehicle` class but I appreciate your thoroughness.

spec/facility_spec.rb
```diff
@@ -15,12 +19,257 @@
...
+   it 'can add available services' do
       expect(@facility.services).to eq([])
       @facility.add_service('New Drivers License')
       @facility.add_service('Renew Drivers License')
       @facility.add_service('Vehicle Registration')
       expect(@facility.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
     end
   end
+
+  describe '#vehicle registration' do
```
dcoleman21, line 31: Overall your tests are looking good, just really be mindful of unnecessary white space at the end of your tests and the end of the file.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    registrant_1 = Registrant.new('Bruce', 18, true )
+    registrant_2 = Registrant.new('Penny', 16, false )
+    registrant_3 = Registrant.new('Tucker', 15, false )
+
+    facility_1 = Facility.new({name: 'Albany DMV Office', address: '2242 Santiam Hwy SE Albany OR 97321', phone: '541-967-2014' })
+    facility_2 = Facility.new({name: 'Ashland DMV Office', address: '600 Tolman Creek Rd Ashland OR 97520', phone: '541-776-6092' })
+
+    expect(registrant_1.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+    expect(registrant_1.permit?).to eq(true)
+    expect(facility_1.administer_written_test(registrant_1)).to eq(false)
```
dcoleman21, line 46: You can make this test more robust by adding the service to your facility, then calling `administer_written_test` again (which should now eq(true) and then checking that a registrants `license_data` updates accordingly.

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+      expect(factory).to be_an_instance_of(VehicleFactory)
+    end
+  end
+
+  describe '#ev_registrations' do
+    it 'has data' do
+      factory = VehicleFactory.new
+      wa_ev_registrations = DmvDataService.new.wa_ev_registrations
+      cars = factory.create_vehicles(wa_ev_registrations)
```
dcoleman21, line 20: Again you can create another variable here or just write your expect statements like such
```ruby
expect(factory.create_vehicles(wa_ev_registrations)).to be_an(Array)
```

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  describe '#initialize' do
+    it 'instance of factory' do
+      factory = VehicleFactory.new
+
+      expect(factory).to be_an_instance_of(VehicleFactory)
+    end
+  end
+
+  describe '#ev_registrations' do
+    it 'has data' do
```
dcoleman21, line 17: This test is really testing `#create_vehicles`. See if you can be more explicit with your naming.

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+      factory = VehicleFactory.new
+      wa_ev_registrations = DmvDataService.new.wa_ev_registrations
+      cars = factory.create_vehicles(wa_ev_registrations) 
+
+      expect(cars).to be_a(Array)
```
dcoleman21, line 22: In your class you have a `wa_registrations(vehicle_details)` method. You would need a test for that.

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'spec_helper'
+require './lib/vehicle'
+require './lib/vehicle_factory'
+require './lib/dmv_data_service'
+
+
+RSpec.describe VehicleFactory do
+  describe '#initialize' do
+    it 'instance of factory' do
```
dcoleman21, line 9: You created an `initialize` method in this class that has `created_vehicles` in it. You would need to test that when you call `factory.created_vehicles` it returns something.

dcoleman21 (commented):
Overall nice job with this project. I would really love to see you be more attentive to your indentation and spacing. When I originally ran your tests I had to go into a couple files to add missing ends. This can be avoided by really paying attention to this. We also don't want any extra white space in our code. I left comments for you to review. Please let me know if you have any questions.

PR #137, opened 2023-04-04
K.D. Hubbard

kdhubb:


lib/vehicle_factory.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+
+    def most_registered_county(source)
+      county_count = Hash.new(0)
+      source.each do |vehicle|
+        county_count[vehicle[:county]] += 1
+      end
+      county_count.sort_by do |county, num|
+        num
+      end.reverse[0].first
```
epintozzi, line 59:
```suggestion
county_count.max_by { |k, v| v}.first
```
I think you can use a max_by here as well to make it a little cleaner

lib/facility_factory.rb
```diff
@@ -0,0 +1,46 @@
+class FacilityFactory
+attr_reader :facilities
+
+  def initialize()
+    @facilities = []
+  end
```
epintozzi, line 6: I like your individual source methods. A next step might be to have a method called in initialize (`load_source`?) to determine the source and kick off the correct `add_xx_source` method so you can go right ahead and `create_facilities` when you're ready. Great work!

epintozzi (commented):
Great work! Tests all look reasonable to me, so intentionally not adding extra comments there. Nice job testing your helpers as well 🔥

PR #140, opened 2023-04-04
Lauren Golden

llgolden:


lib/registrant.rb
```diff
@@ -0,0 +1,20 @@
+class Registrant
+  attr_reader :name, :age, :permit
+
```
epintozzi, line 3:
```suggestion
```
We don't need a space here

lib/registrant.rb
```diff
@@ -0,0 +1,20 @@
...
+  attr_accessor :license_data
+
+  def initialize(name, age, permit=false)
+    @name = name 
+    @age = age
+    @permit = permit
+    @license_data = {:written=>false, :license=>false, :renewed=>false}
```
epintozzi, line 10:
```suggestion
    @license_data = {written: false, license: false, renewed: false}
```
This format is slightly more conventional

lib/vehicle.rb
```diff
@@ -1,18 +1,18 @@
 require 'date'
 
 class Vehicle
-  attr_reader :vin,
-              :year,
-              :make,
-              :model,
-              :engine
+  attr_reader :vin, :year, :make, :model, :engine
+
```
epintozzi, line 5:
```suggestion 
```

lib/vehicle.rb
```diff
@@ -1,18 +1,18 @@
...
 
 class Vehicle
-  attr_reader :vin,
-              :year,
-              :make,
-              :model,
-              :engine
+  attr_reader :vin, :year, :make, :model, :engine
+              
+  attr_accessor :registration_date
```
epintozzi, line 6: I'm getting several errors when I run your specs:
```
NoMethodError:
       undefined method `plate_type=' for #<Vehicle:0x00000001510f28f8>
```
Adding `:plate_type` to your accessors should fix that. I'm curious if you also saw this error before submitting or if there was a different way you thought this was being handled?

```suggestion
  attr_accessor :registration_date, :plate_type
```

lib/facility.rb
```diff
@@ -1,14 +1,68 @@
...
+  def add_service(service)
     @services << service
   end
-end
+
+  def register_vehicle(vehicle)
+    if services.include?('Vehicle Registration')
+      @registered_vehicles << vehicle
+      vehicle.registration_date = Date.today
+      fees_and_plates(vehicle)
```
epintozzi, line 21: I like that you've pulled this out to its own method instead of putting all the logic inside `register_vehicle`. That makes it more dynamic later on if fee change, for example.

lib/facility.rb
```diff
@@ -1,14 +1,68 @@
...
+  def administer_written_test(registrant)
+    if services.include?('Written Test')
+      if registrant.permit? && registrant.age >= 16
+        registrant.license_data[:written] = true
+      else 
+        false
+      end
+    else
+      false
+    end
```
epintozzi, line 50: Nested `if`s can get buggy and hard to test every single outcome. You should actually be able to bring all the conditionals up to the first `if` for the same outcome:
```suggestion
    if services.include?('Written Test') && registrant.permit? && registrant.age >= 16
      registrant.license_data[:written] = true
    else
      false
    end
```

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,27 @@
+class VehicleFactory
+              
+  def initialize
+  end
```
epintozzi, line 4: If we aren't doing anything in our initialize method, we can actually just get rid of it! Ruby knows how to initialize if there's nothing else that needs to happen out of the ordinary.
```suggestion
```

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def create_vehicles(vehicle_list)
+    vehicle_list.map do |vehicle_data|
+      Vehicle.new(vehicle_data)
+    end
+  end
+
+  def rename_vehicles(vehicle_list)
```
epintozzi, line 12: I think I might call this method something like "map_vehicle_attributes". You are literally taking attributes from some source and remapping them to match OUR source (the `Vehicle` class attributes). Naming aside, I really like that you've broken this out to its own method.

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,67 @@
+require 'spec_helper'
+
+RSpec.describe VehicleFactory do
+  it "exists" do
+    factory = VehicleFactory.new
+    expect(factory).to be_an_instance_of(VehicleFactory)
+  end
+
+  it "creates a vehicle from a list of one vehicle" do
```
epintozzi, line 9: You were really smart about how you tested this. Start small. We just need to test from one small piece of data that the `create_vehicles` method works. Then you incremented in the next test to have multiple vehicles before finally testing the external data set. Really nice work.

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    wa_ev_registrations = DmvDataService.new.wa_ev_registrations
+    renamed_vehicles = factory.rename_vehicles(wa_ev_registrations)
+    created_vehicles = factory.create_vehicles(renamed_vehicles)
+    
+    expect(created_vehicles[0]).to be_an_instance_of(Vehicle)
+    expect(created_vehicles).to be_an(Array)
+    expect(created_vehicles.count).to eq(1000) 
+  end
+
+  it "can rename vehicle variables" do
```
epintozzi, line 47: Excellent! We always want to test our helpers just like you're doing here 🔥

lib/facility_factory.rb
```diff
@@ -0,0 +1,58 @@
+class FacilityFactory
```
epintozzi, line 1: I like how you've handled formatting the data in this class. The alternative I sometimes see is a method per data source, so something like `def format_oregon_data`, `def format_ny_data`, etc. either works just fine. I could see the way it is set up here getting tedious if you have to handle all 50 states. At some point you'd probably want to switch to formatting per state rather than adding another `elsif` per state.

epintozzi (commented):
Nice work! I had an issue with a couple tests failing that I commented on but otherwise I really like your approach to testing here.

PR #141, opened 2023-04-04
Jeff Nelson

jpnelson85:


lib/vehicle_factory.rb
```diff
@@ -0,0 +1,38 @@
...
+  
+  def initialize
+    @created_vehicles = []
+  end
+
+  def create_vehicles_by_state(state_database)
+    wa_ev_registrations = DmvDataService.new.wa_ev_registrations
+    ny_state_registrations = DmvDataService.new.ny_dmv_office_locations
+
+    if state_database == wa_ev_registrations
```
epintozzi, line 14: I don't love the direct comparison to `DmvDataService.new.wa_ev_registrations` (and same for New York) because that is going to execute the creation of a `DmvDataService` object AND make the API call to make sure the two data sets match. This isn't ideal for two reasons.

1. It's expensive to make API calls. It's really slow. You probably noticed your test suite slowing down as you progressed through the project. It can also literally be expensive. A lot of APIs cost money. You pay every time you access the API's data. That's not the case here, but often is.
2. While not super likely, it's possible that data changes between the time you call your `create_vehicles_by_state` method and lines 14 or 25 execute which means it could actually be Washington EV Registration data but since the data sets don't 100% match, it'll return false.

Instead I would approach this by finding something ABOUT the data you can match. For example is there a key/value pair in the hash that is unique to the Washington data set?

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,38 @@
...
+      wa_ev_registrations.each do |car|
+        @created_vehicles << Vehicle.new({
+            vin: car[:vin_1_10],
+            year: car[:model_year],
+            make: car[:make],
+            model: car[:model],
+            engine: 'ev'
+        })
+      end
+        @created_vehicles
```
epintozzi, line 24: I would also try to refactor out the formatting of the data so this method can focus only on creating vehicle objects. It might look something like this:
```ruby
def create_vehicles_by_state(state_database)
    wa_ev_registrations = DmvDataService.new.wa_ev_registrations
    ny_state_registrations = DmvDataService.new.ny_dmv_office_locations

    if state_database == wa_ev_registrations
      wa_registrations = format_wa_data(state_database)
      wa_registrations.each do |car_data|
        @created_vehicles << Vehicle.new(car_data)
      end
    elsif state_database == ny_state_registrations
      vehicle_data.map do |vehicle|
        @created_vehicles << Vehicle.new({
            vin: vehicle[:vin],
            year: vehicle[:model_year],
            make: vehicle[:make],
            model: nil,
            engine: 'ev'
          })
      end
    end
    @created_vehicles
  end

  def format_wa_data(wa_ev_registrations)
    formatted_car_info = {}
    wa_ev_registrations.each do |car|
      formatted_car_info[:vin] = car[:vin_1_10],
      formatted_car_info[:year] = car[:model_year],
      formatted_car_info[:make] = car[:make],
      formatted_car_info[:model] = car[:model],
      formatted_car_info[:engine] = 'ev'
    end
  end
```
I would do the same for New York too but I'll leave that for you to play with.

lib/facility_factory.rb
```diff
@@ -0,0 +1,40 @@
+class FacilityFactory
```
epintozzi, line 1: This class looks great! One way you could refactor just a little would be to keep only the formatting of the data in the individual state methods and then have one `create_facilities` method responsible for the `Facility.new....` part of the transaction. Overall I like this a lot though and that you've got each state separated out to its own method rather than a giant if statement.

epintozzi (commented):
Overall nice work! Testing looks good. I think you could go a bit deeper with your two factory specs, however those are hard to test well and keep up with the changing nature of external data. You'll learn some more techniques for handling stuff like this in mod 3.

lib/facility.rb
```diff
@@ -1,14 +1,70 @@
...
+      nil
+    end
+  end
+
+  def add_service(service)
     @services << service
   end
+
+  def administer_written_test(registrant)
+    if (registrant.permit?) && (registrant.age > 15) && (@services.include?('Written Test'))
```
epintozzi, line 46: In your submission form you asked:
> Besides refactoring the code, what's the best way to separate multiple lines from one.  There are good examples of what I'm talking about in my Facility class...Line 46, Line 54, and Line 63

I'm not totally sure what you mean by this. Can you say a little more about what you're asking? If you'd like to jump on a call to talk through if that's easily, let me know. Happy to do so.

PR #142, opened 2023-04-04
Ashley Plasket - DMV

aplasket:


lib/facility_factory.rb
```diff
@@ -0,0 +1,60 @@
...
+    if state == :OR
+      oregon_processing(data)
+    elsif state == :NY
+      new_york_processing(data)
+    elsif state == :MO
+      missouri_processing(data)
+    else
+      "that state's data is not valid for processing"
+    end
+  end
```
epintozzi, line 14: I might change your `processing` methods to only do the processing and more the creation of facilities up here. I've only altered Oregon (and added a guard clause), but it would look something like this:
```ruby
def create_facilities(data, state)
    return "that state's data is not valid for processing" unless valid_state?(state)

    all_facilities = if state == :OR
                       oregon_processing(data)
                     elsif state == :NY
                       new_york_processing(data)
                     elsif state == :MO
                       missouri_processing(data)
                     end

    all_facilities.map do |facility_data|
      Facility.new(facility_data)
    end
  end

  def valid_state?(state)
    state == :OR || state == :NY || state == :MO
  end

  def oregon_processing(data)
    data.map do |facility|
      {
        name: facility[:title],
        address: JSON.parse(facility[:location_1][:human_address]).values.join(" "),
        phone: OR_format_phone_number(facility)
      }
    end
  end
```

This way, you're only calling `Facility.new` in one place rather than 3 different places, `create_facilities` then becomes truly responsible for creating facilities and the `processing` methods are left to just process.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,14 @@
+class VehicleFactory
```
epintozzi, line 1: Nothing to add. Looks good!

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+      expect(@oregon).to be_an(Array)
+      expect(@new_york).to be_an(Array)
+      expect(@missouri).to be_an(Array)
+    end
+  end
+  
+  describe '#oregon processing' do
+    it 'creates oregon facility objects' do
+      expect(@oregon.first).to be_a(Facility)
+      expect(@oregon.sample).to be_a(Facility)
```
epintozzi, line 37: Ooh this is interesting! I like the use of `sample` here so you can make sure you aren't getting a weird false positive by only checking `first`.

epintozzi (commented):
Great work!

PR #143, opened 2023-04-04
Will Zale

WZale:
This pull request completes iteration 2. 

The following tests were added to facility_spec.rb:
-register vehicles
  -it can register a vehicle
  -it cannot register a vehicle that is already registered at another facility

-administer tests
  -it can administer a written test
  -it cannot administer a written test for underage registrants to earn a license
  -it can administer a road test
  -it cannot administer a road test to an underage registrant
  -it can renew a license

The following features were added to facility.rb:
-register vehicle
-administer written test
-renew driver's license

The following tests were added to registrant_spec.rb:
-it has an earn permit method

The following feature was added to registrant.rb:
-earn permit


WZale:
Looks good!

PR #144, opened 2023-04-04
Sarah Garlock

sarahgarlock:


README.md
```diff
@@ -1,3 +1,17 @@
 # The DMV
 
-This is the starter repo for the BE Mod1 DMV project.
+## Reflection
```
dcoleman21, line 3: Really great reflections here.

lib/facility.rb
```diff
@@ -1,14 +1,64 @@
...
+    @registered_vehicles = []
+    @collected_fees = 0
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
```
dcoleman21, line 22: In this method, if you had a `plate_type` method in your `Vehicle` class, you could set the plate types there and get rid of the lines that are assigning the types in this method. 
You have `@plate_type` in your `initialize` method in your `Vehicle` class but having it as an instance variable there doesn't make it very dynamic and now puts the responsibility on this method to reassign them.

lib/facility_factory.rb
```diff
@@ -0,0 +1,37 @@
+class FacilityFactory
+
+  def initialize
```
dcoleman21, line 3: You don't need this method. We aren't instantiating a new `FacilityFactory` with any attributes/args.

lib/facility_factory.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def create_facilities(registration_location)
+    registration_location.map do |registration|
+      Facility.new(registration_location)
```
dcoleman21, line 8: So if you put a pry on line 6 1/2, you will see that `registration_location` is a hash but it is the hash that is passed if from 
```ruby    
@facility_1 = Facility.new({name: 'Albany DMV Office', address: '2242 Santiam Hwy SE Albany OR 97321', phone: '541-967-2014'})
```
We are wanting to create entirely new `Facility` objects from the data that we are getting from our external api
```ruby
@or_dmv_office_locations = DmvDataService.new.or_dmv_office_locations
```
So this one method should be able to take in any facilities data and format it the way that we want it. 
So instead of creating methods for each different state (what if we decided we needed all 50 states; that wouldn't be very DRY), you would want to have one method that was dynamic. 
So you may need to create helper methods that format things like name, address, and phone number depending on the state info you are pulling from that you could call in your `create_facilities` method.

lib/registrant.rb
```diff
@@ -0,0 +1,25 @@
...
+      :renewed=>false
+    }
+  end
+
+  def permit?
+    @permit
+  end
+
+  def earn_permit
+    @permit = true
```
dcoleman21, line 23: You could make this more robust by adding a guard to check `@permit = true if @age >= 15`.

lib/vehicle.rb
```diff
@@ -1,18 +1,22 @@
...
+                :plate_type
 
   def initialize(vehicle_details)
     @vin = vehicle_details[:vin]
     @year = vehicle_details[:year]
     @make = vehicle_details[:make]
     @model = vehicle_details[:model]
     @engine = vehicle_details[:engine]
+    @registration_date = nil
+    @plate_type = nil
```
dcoleman21, line 19: I would not have this as an instance variable. It is not dynamic this way and relies on another class to assign it. 
I would create a stand alone method called `plate_type` where you can the type based on the type of vehicle.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,16 @@
+class VehicleFactory
+  def initialize
```
dcoleman21, line 2: You do not need this method. When `VehicleFactory` is instantiated it has no attribute/args.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,16 @@
...
+  end
+ 
+  def create_vehicles(registration_location)
+    registration_location.map do |registration|
+      Vehicle.new(
```
dcoleman21, line 7: You are pulling in the correct data with `registration_location` but when we create a new `Vehicle` object what datatype should the argument be?
```ruby
Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice} )
```
I looked at the `vehicle_spec` to get this. How could you update 
```ruby
Vehicle.new(...)
```
on line 7?

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'spec_helper'
+
+RSpec.describe do
+  before(:each) do
+    @facility_factory = FacilityFactory.new
+    @facility_1 = Facility.new({name: 'Albany DMV Office', address: '2242 Santiam Hwy SE Albany OR 97321', phone: '541-967-2014'})
```
dcoleman21, line 6: You don't need this. The only thing that should be in your setup is
```ruby
@facility_factory = FacilityFactory.new
@or_dmv_office_locations = DmvDataService.new.or_dmv_office_locations
```
If you want to use the `before(:each)` then you will want to put the other states in there as well
```ruby
@new_york_facilities = DmvDataService.new.ny_dmv_office_locations
@missouri_facilities = DmvDataService.new.mo_dmv_office_locations
```
Then you would just write separate tests for each state. 
We don't want to pass `@facility_1` info into line 7. 
So our expect statements should be
```ruby
expect(@facility_factory.create_facilities(@or_dmv_office_locations)).to be_an(Array)
```
Note the expect statement would change based on the state data you are passing to `@facility_factory.create_facilities`.

spec/facility_spec.rb
```diff
@@ -23,4 +34,133 @@
...
-end
+
+  describe '#register vehicle' do
+    it 'may offer vehicle registration' do
+      expect(@facility_1.add_service('Vehicle Registration')).to eq(["Vehicle Registration"])
+      expect(@cruz.registration_date).to be nil
+      expect(@facility_1.registered_vehicles).to eq([])
+      expect(@facility_1.collected_fees).to eq 0
+
+      expect(@facility_1.register_vehicle(@cruz)).to eq([@cruz])
```
dcoleman21, line 45: So `register_vehicle(@cruz)` is a command that should update the `registered_vehicles` method. So you wouldn't want it how you have it written on line 45. I would update your test to read like
```ruby
@facility_1.register_vehicle(@cruz)
expect(@facility_1.registered_vehicles).to eq([@cruz])
```
You would want to update all the tests to emulate this.

spec/facility_spec.rb
```diff
@@ -23,4 +34,133 @@
...
+  end
+
+  # Written test
+
+  it 'can change licence data to true' do
+    expect(@registrant_1.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+    expect(@registrant_1.permit?).to be true
+
+    expect(@facility_1.administer_written_test(@registrant_1)).to be false
+    expect(@facility_1.add_service('Written Test')).to eq(['Written Test'])
```
dcoleman21, line 79: `@facility_1.add_service('Written Test')` is a command that updates the `.services` method.

spec/facility_spec.rb
```diff
@@ -23,4 +34,133 @@
...
+    expect(@facility_1.administer_written_test(@registrant_1)).to be true
+    expect(@registrant_1.license_data).to eq({:written=>true, :license=>false, :renewed=>false})
+
```
dcoleman21, line 82: Make sure you do not have white space at the end of any tests, classes, or methods.

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'spec_helper'
+
+RSpec.describe VehicleFactory do
+  before(:each) do
```
dcoleman21, line 4: You can use this setup structure for you `FacilityFactory` spec.

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    @factory = VehicleFactory.new
+    @wa_ev_registrations = DmvDataService.new.wa_ev_registrations
+  end
+
+  it 'exists' do
+    expect(@factory).to be_an_instance_of VehicleFactory
+  end
+
+  it 'creates a list of vehicles using an API' do
+    @factory.create_vehicles(@wa_ev_registrations)
```
dcoleman21, line 14: I don't think you need this test because it is essentially the same thing that you are testing on line 19

dcoleman21 (commented):
Overall nice job on this project. Things to be mindful of: 
- make sure your indentation/spacing is good. You had a few extra white spaces at the tests and a few methods. Not super bad, just wanted to call your attention to it. 
- When writing your tests, see if you can identify what is a command vs what you want your expect statement to be (see comments in tests).
- For your FacilityFactory, I would look at how you built out your tests for the VehicleFactory. You will be essentially doing the same thing except there will be a few extra tests for each state. 

I left comments throughout your code. Let me know if you have any questions.

PR #146, opened 2023-04-04
Alex Kiselich DMV Project

AlexKiselich:


lib/vehicle.rb
```diff
@@ -22,4 +24,16 @@ def antique?
   def electric_vehicle?
     @engine == :ev
   end
+
+  def plate_type
+    if engine == :ev
+      :ev
+    elsif engine == :ice && antique?
```
epintozzi, line 31:
```suggestion
    elsif antique?
```
I would probably say it's okay to just check `antique?` here since in our imaginary app there's only two engine types, however I can definitely follow the line of thought that it's not necessarily the case.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,24 @@
...
+      vehicle_details = {
+        vin: vehicle[:vin_1_10],
+        year: vehicle[:model_year],
+        make: vehicle[:make],
+        model: vehicle[:model],
+        engine: :ev,
+        registration_date: nil
+      }
+      
+      @vehicles << vehicle = Vehicle.new(vehicle_details)
```
epintozzi, line 18:
```suggestion
      @vehicles << Vehicle.new(vehicle_details)
```
You can actually go ahead and just shovel the vehicle in directly. No need to assign it to a variable if we aren't going to use the variable for anything.

spec/facility_spec.rb
```diff
@@ -24,3 +26,153 @@
...
+      expect(@facility_2.registered_vehicles).to eq([])
+      expect(@facility_2.collected_fees).to eq(0)
+
+    end
+  end
+end
+
+
+
+RSpec.describe Facility do
```
epintozzi, line 94: You don't need to do `RSpec.describe ClassName` except for the very first block. I think you might be trying to separate it out for each type of behavior that you're testing, in this case the Written Test. Instead you can do something like this (nested inside the one original `RSpec.describe` block):
```
describe "written test" do
 ...
end
```

lib/vehicle.rb
```diff
@@ -6,13 +6,15 @@ class Vehicle
...
               :engine
+  attr_accessor :registration_date
 
   def initialize(vehicle_details)
     @vin = vehicle_details[:vin]
     @year = vehicle_details[:year]
     @make = vehicle_details[:make]
     @model = vehicle_details[:model]
     @engine = vehicle_details[:engine]
+    @registration_date = nil
```
epintozzi, line 17: You should update your vehicle_spec to test that this attribute is present as well as testing the `plate_type` method below.

epintozzi (commented):
Nice work!

PR #148, opened 2023-04-04
Andy Stilmock DMV Repo Pull Request

AStilmock:


README.md
```diff
@@ -1,3 +1,17 @@
 # The DMV
 
-This is the starter repo for the BE Mod1 DMV project.
+# BE MOD 1 DMV PROJECT - ANDY STILMOCK
+
+## Digging into the existing codebase
+
+  ### - Describe the steps you took to dig in to this code base. What was your process? If you don’t feel you had a process, define one that you might like to try next time.
+Initially, I began with reading through the iteration 1 directions before diving into the code in order to get a good grip on the expectations and setup for this project. After cloning the repo to my computer and diving into the preexisting files - I just started at the top of the list in the library files, looking at the library file then the corresponding test file in the spec folder. As I went through the files I took some notes on what I was confident I would be able to figure out/complete for the steps listed in the interaction pattern but also, took notes on the lines of code that I did not understand or knew I would need some further questioning and research on to take a look at when I started working through the tests and interaction patterns for the project.
```
KatBrandt, line 8: Love the process! Especially the part about reading the directions to understand the expectations first.

README.md
```diff
@@ -1,3 +1,17 @@
...
+Initially, I began with reading through the iteration 1 directions before diving into the code in order to get a good grip on the expectations and setup for this project. After cloning the repo to my computer and diving into the preexisting files - I just started at the top of the list in the library files, looking at the library file then the corresponding test file in the spec folder. As I went through the files I took some notes on what I was confident I would be able to figure out/complete for the steps listed in the interaction pattern but also, took notes on the lines of code that I did not understand or knew I would need some further questioning and research on to take a look at when I started working through the tests and interaction patterns for the project. 
+
+  ### - What was hard about working with code you did not write?
+Really the only difficulties are the unknown items - methods or tests written in a way I don't understand or have not learned yet     (.parse as an example). However, with this being an expected challenge after completing Turing that would be required when diving into any existing codebase in the future - this is a welcome challenge.
```
KatBrandt, line 11: Great perspective!!! When you get a job, you'll be jumping into a huge existing codebase, so embrace this mindset! :)

lib/facility.rb
```diff
@@ -1,14 +1,146 @@
...
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    if self.services.include?("Vehicle Registration")
+    @registered_vehicles << vehicle
```
KatBrandt, line 24: Make sure to indent the code that's inside the if block.

Alternatively, although this looks great and is very readable, if you wanted to get rid of that if/end block, you could use what we call a guard clause. This is a one line if statement to check a condition before we even run the rest of the method. In this case, it would look like `return nil if !services.include?("Vehicle Registration")`. then you wouldn't need the if/end block.

Also, we don't need the `self` keyword when calling a method that exists within the class we're currently in. It's implied.

lib/facility.rb
```diff
@@ -1,14 +1,146 @@
...
+    change_registration_date(vehicle)
+    make_plate_type(vehicle)
+    collect_fees(vehicle)
```
KatBrandt, line 27: Love how readable this method is with these helper methods!

lib/facility.rb
```diff
@@ -1,14 +1,146 @@
...
+      return nil
+    end
+  end
+
+  def change_registration_date(vehicle)
+    vehicle.registration_date = Date.today
+  end
+    
+  def make_plate_type(vehicle)
+    if vehicle.antique?
```
KatBrandt, line 39: Good job using methods that already exist in your vehicle class.

lib/facility.rb
```diff
@@ -1,14 +1,146 @@
...
+    elsif vehicle.electric_vehicle?
+      @collected_fees += 200
+    else 
+      @collected_fees += 100
+    end
+  end
+
+  def administer_written_test(registrant)
+    if self.services.include?('Written Test')
+      if registrant.permit? && registrant.age >= 16
```
KatBrandt, line 60: This works, but nested conditionals can get unreadable very quickly and can use more processing power. If you included all these conditions on line 59, you could unnest this if block.

KatBrandt: `if services.include?("Written Test") && registrant.permit? && reigstrant.age >= 16`

lib/facility.rb
```diff
@@ -1,14 +1,146 @@
...
+        registrant.license_data[:renewed] = true
+      else
+        false
+      end
+    else
+      false
+    end
+  end 
+
+  def create_facility(state)
```
KatBrandt, line 94: This method and all the following methods, I would abstract into their own class, maybe a FacilityFactory class. Remember SRP, we want the Facility class to just handle facilities, not the creation of multiple facilities. Another class  like a FacilityFactory class could be in charge of creating lots of facilities from json data.

lib/facility.rb
```diff
@@ -1,14 +1,146 @@
...
+  end
+
+
+  def create_facility_oregon(state)
+    oregon = DmvDataService.new.or_dmv_office_locations
+    new_facility_list1 = []
+    oregon.each do |station|
+      facility_or_hash = {name: nil, address: nil, phone: nil}
+      facility_or_hash[:name] = station[:title]
+      facility_or_hash[:address] = JSON.parse(station[:location_1][:human_address]).values.join(" ")
```
KatBrandt, line 111: Parsing right here makes sense because just this one value was still in JSON format

lib/facility.rb
```diff
@@ -1,14 +1,146 @@
...
+      create_facility_ny(state)
+    elsif state == "missouri"
+      create_facility_missouri(state)
+    end
+  end
+
+
+  def create_facility_oregon(state)
+    oregon = DmvDataService.new.or_dmv_office_locations
+    new_facility_list1 = []
```
KatBrandt, line 107: I like that you're storing the facility objects in this array, however, it would make more sense to store this array in state, that way you only have to run this method once to get all the facility objects instead of having to do all this iteration over again.  Having this as state would make more sense in a FacilityFactory class.

lib/facility.rb
```diff
@@ -1,14 +1,146 @@
...
+    oregon.each do |station|
+      facility_or_hash = {name: nil, address: nil, phone: nil}
```
KatBrandt, line 109: We don't need to build a whole new hash here, since `station` already is a hash. You could alter some of the keys to match what your initialize is looking for, but essentially, you could do Facility.new(station) and your initialize is already set up to pull out the keys it needs.

lib/registrant.rb
```diff
@@ -0,0 +1,17 @@
+class Registrant
+  attr_reader :name, :age, :permit, :license_data
```
KatBrandt, line 2: Careful here, we only want to expose the methods that we need. We do not need a method called `permit` since we already have `permit?`

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,19 @@
+class VehicleFactory
+  
+  def initialize
+  end
+
+  def create_vehicles(location)
+    new_car_list = []
```
KatBrandt, line 7: Same thing here, I would want this array held in state, so I could refer back to it instead of having to run this method everytime I want to access these vehicles.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,19 @@
...
+    location.each do |vehicle|
```
KatBrandt, line 8: Good use of #each here.

spec/facility_spec.rb
```diff
@@ -23,4 +38,198 @@
...
+      @facility_1.renew_drivers_license(@registrant_3)
+      expect(@facility_1.renew_drivers_license(@registrant_3)).to eq(false)
+      expect(@registrant_3.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+
+      @facility_2.renew_drivers_license(@registrant_2)
+      expect(@facility_1.renew_drivers_license(@registrant_2)).to eq(true)
+      expect(@registrant_2.license_data).to eq({:written=>true, :license=>true, :renewed=>true})
+    end
+  end
+
```
KatBrandt, line 218: I don't see tests for some of your helper methods like `create_plate_type` that you use in the register_vehicle method. Make sure we are unit testing all of our methods, including helper methods.

spec/facility_spec.rb
```diff
@@ -2,8 +2,23 @@
 
 RSpec.describe Facility do
   before(:each) do
-    @facility = Facility.new({name: 'Albany DMV Office', address: '2242 Santiam Hwy SE Albany OR 97321', phone: '541-967-2014' })
+    @facility_new = Facility.new({name: nil, address: nil, phone: nil})
+    @facility   = Facility.new({name: 'Albany DMV Office', address: '2242 Santiam Hwy SE Albany OR 97321', phone: '541-967-2014'})
+    @facility_1 = Facility.new({name: 'Albany DMV Office', address: '2242 Santiam Hwy SE Albany OR 97321', phone: '541-967-2014'})
```
KatBrandt, line 7: Not sure why we have two facilities with the same information, usually we don't want to do that, but wondering if you had a purpose in mind.

spec/facility_spec.rb
```diff
@@ -23,4 +38,198 @@
...
+  describe "Vehicle Registration" do
+    it "adds service for facility_1" do
+      @facility_1.add_service("Vehicle Registration")
+      expect(@facility_1.services).to eq(["Vehicle Registration"])
+    end
+
+    it "checks cruz registration date" do
+      expect(@cruz.registration_date).to eq(nil)
+    end
+
```
KatBrandt, line 51: Your spacing of your tests looks much better!

spec/facility_spec.rb
```diff
@@ -23,4 +38,198 @@
...
+      @facility_2.renew_drivers_license(@registrant_2)
+      expect(@facility_1.renew_drivers_license(@registrant_2)).to eq(true)
+      expect(@registrant_2.license_data).to eq({:written=>true, :license=>true, :renewed=>true})
+    end
+  end
+
+  describe "creating new facilities from existing data sources" do
+    it "creates oregon facility" do
+      @facility_new.create_facility("oregon")
+      expect(@facility_new.create_facility("oregon")).to be_an(Array)
```
KatBrandt, line 222: Yeah, this is really the only thing we can test for since we don't control the data that's coming in.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+RSpec.describe Registrant do
+  before(:each) do
+    @registrant_1 = Registrant.new('Bruce', 18, true )
+    @registrant_2 = Registrant.new('Penny', 15 )
+  end
+
+  describe "Initialize" do
+    it "exists" do
+      expect(@registrant_1).to be_an_instance_of(Registrant)
+      expect(@registrant_2).to be_an_instance_of(Registrant)
```
KatBrandt, line 12: Redundant testing here. For things like instantiating and checking attributes, we really only have to test it once and then we can assume it will work for all Registrant instances.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      expect(@registrant_1.name).to eq("Bruce")
+      expect(@registrant_1.age).to eq(18)
+      expect(@registrant_1.permit?).to eq(true)
+      expect(@registrant_1.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+    end
+
+    it "checks registrant_2 data" do
+      expect(@registrant_2.name).to eq("Penny")
+      expect(@registrant_2.age).to eq(15)
+      expect(@registrant_2.permit?).to eq(false)
```
KatBrandt, line 27: I _would_ test the permit? method twice though since one instance is using the default value and another is passing in that data.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      expect(@registrant_2.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+    end
+  end
+
+  describe "earn permit method" do
+    it "earns permit" do
+      @registrant_2 = Registrant.new('Penny', 15 )
+      @registrant_2.earn_permit
+      expect(@registrant_2.permit?).to eq(true)
```
KatBrandt, line 36: I like to see the change happening within the test. I would have started with an assertion that .permit? was false for Penny, under line 34, then did the rest of the testing you did to make sure it all changed. Also, you don't need to do line 34 here, you did that in your before(:each) block so you can now just call @registrant_2 here and it will be that object.

KatBrandt (commented):
Good job with commits in branches. Moving forward, lets try naming your branches by the feature they're implementing. Great job!

PR #149, opened 2023-04-04
DMV Submission - Garrett Gregor

garrettgregor:
This is my submission for the DMV project! yayy!!! I felt way better about this than nodes, but got stuck on the beginning of iteration 3 for longer than I was hoping.

After a significant amount of refactoring with helper methods from Dani's help, I was able to test all of my helper methods and ensure that a FacilityFactory could create facilities from either OR, NY or MO data.

In future iterations I would like to be able to normalize all of the strings and not create facilities that have been listed as closed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,3 +83,41 @@
 [14] pry(RSpec::ExampleGroups::DmvDataService)> DmvDataService.new.wa_ev_registrations[5][:make]
 => "VOLKSWAGEN"
 ```
+
+## Iteration 3
+
+- When creating a facility factory, it appears you can use `||` as a way to check for different keys but not `or`
+- Ideally, we would normalize street address and not have it be all caps (an example of how you might go about this is below):
+
+```
+[2] pry(#<RSpec::ExampleGroups::FacilityFactory::Iteration3::DifferentFactories>)> a = "hello world"
+=> "hello world"
+[3] pry(#<RSpec::ExampleGroups::FacilityFactory::Iteration3::DifferentFactories>)> a.capitalize
+=> "Hello world"
+[4] pry(#<RSpec::ExampleGroups::FacilityFactory::Iteration3::DifferentFactories>)> a.capitalize
+=> "Hello world"
+[5] pry(#<RSpec::ExampleGroups::FacilityFactory::Iteration3::DifferentFactories>)> a.split(" ").capitalize.join
+NoMethodError: undefined method `capitalize' for ["hello", "world"]:Array
+from (pry):5:in `block (4 levels) in <top (required)>'
+[6] pry(#<RSpec::ExampleGroups::FacilityFactory::Iteration3::DifferentFactories>)> a.split(" ").each do |caps|
+[6] pry(#<RSpec::ExampleGroups::FacilityFactory::Iteration3::DifferentFactories>)*   caps.capitalize
+[6] pry(#<RSpec::ExampleGroups::FacilityFactory::Iteration3::DifferentFactories>)* end  
+=> ["hello", "world"]
+[7] pry(#<RSpec::ExampleGroups::FacilityFactory::Iteration3::DifferentFactories>)> ret = a.split(" ").each do |caps|
+[7] pry(#<RSpec::ExampleGroups::FacilityFactory::Iteration3::DifferentFactories>)*   caps.capitalize!  
+[7] pry(#<RSpec::ExampleGroups::FacilityFactory::Iteration3::DifferentFactories>)* end  
+=> ["Hello", "World"]
+[8] pry(#<RSpec::ExampleGroups::FacilityFactory::Iteration3::DifferentFactories>)> a
+=> "hello world"
+[9] pry(#<RSpec::ExampleGroups::FacilityFactory::Iteration3::DifferentFactories>)> a.capitalize!
+=> "Hello world"
+[10] pry(#<RSpec::ExampleGroups::FacilityFactory::Iteration3::DifferentFactories>)> ret
+=> ["Hello", "World"]
+[11] pry(#<RSpec::ExampleGroups::FacilityFactory::Iteration3::DifferentFactories>)> ret.join
+=> "HelloWorld"
+[12] pry(#<RSpec::ExampleGroups::FacilityFactory::Iteration3::DifferentFactories>)> ret
+=> ["Hello", "World"]
+[13] pry(#<RSpec::ExampleGroups::FacilityFactory::Iteration3::DifferentFactories>)> ret.join(" ")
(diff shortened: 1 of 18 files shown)


lib/registrant.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def initialize(name, age, permit=false)
+    @name         = name
+    @age          = age
+    @permit       = permit
+    @license_data = {
+      :written  =>  false,
+      :license  =>  false,
+      :renewed  =>  false
+    }
```
epintozzi, line 15:
```suggestion
    @license_data = {
      written:  false,
      license:  false,
      renewed:  false
    }
```
Very minor but this is the preferred syntax for setting symbols as hash keys

lib/facility.rb
```diff
@@ -1,14 +1,64 @@
...
+        @collected_fees += 200
+      else 
+        vehicle.plate_type = :regular
+        @collected_fees += 100
+      end
+      @registered_vehicles << vehicle
+    else
+      nil
+    end
+  end
```
epintozzi, line 39: `register_vehicle` is doing a lot. I might try to pull out the different functions happening here to their own methods. Something like:
```ruby
  def register_vehicle(vehicle)
    if @services.include?("Vehicle Registration")
      collect_fees(vehicle)
      set_plate_type(vehicle)
      set_registration_date(vehicle)
    end
  end

  def collect_fees(vehicle)
    if vehicle.antique?
      @collected_fees += 25
      @collected_fees += 200
    else
      @collected_fees += 100
    end
  end

  def set_registration_date(vehicle)
  end

  def set_plate_type(vehicle)
    if vehicle.antique?
      vehicle.plate_type = :antique
      vehicle.plate_type = :ev
    else
      vehicle.plate_type = :regular
    end
  end
```

lib/facility.rb
```diff
@@ -1,14 +1,64 @@
...
+        vehicle.plate_type = :ev
+        @collected_fees += 200
+      else 
+        vehicle.plate_type = :regular
+        @collected_fees += 100
+      end
+      @registered_vehicles << vehicle
+    else
+      nil
+    end
```
epintozzi, line 38: The `else` is also not strictly necessary. By default this method will return `nil` if the `if` condition is not true.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,20 @@
+class VehicleFactory
+  def initialize
+    
+  end
+
+  def create_vehicles(registrations)
+    created_vehicles = registrations.map do |car|
```
epintozzi, line 7: Unlike `each` (which returns the original collection), `map` will return the new collection so you don't need to set this variable here or explicitly return it on line 18. 
```suggestion
    registrations.map do |car|
```

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,20 @@
...
+        { 
+          vin:    car[:vin_1_10],
+          year:   car[:model_year],
+          make:   car[:make],
+          model:  car[:model],
+          engine: :ev
+        }
+      )
+    end
+    created_vehicles
```
epintozzi, line 18:
```suggestion
```

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,20 @@
+class VehicleFactory
+  def initialize
+    
+  end
```
epintozzi, line 4: If you aren't doing anything in your `initialize` method, you can actually leave it out. Ruby will still be able to instantiate without it.

lib/facility_factory.rb
```diff
@@ -0,0 +1,54 @@
...
+        Facility.new(
+          {
+            name:      facility_details[:name],
+            address:   facility_details[:address],
+            phone:     facility_details[:phone]
+          }
+        )
+      end
+    created_facilities
+  end
```
epintozzi, line 14:
```suggestion
  def create_facilities(state_raw_data)
    state_raw_data.map do |facility|
      facility_details = format_facility(facility)
      Facility.new(
        {
          name:      facility_details[:name],
          address:   facility_details[:address],
          phone:     facility_details[:phone]
        }
      )
    end
  end
```
The indentation is a little off here.

I like that you have your formatting being done as a separate method so that `create_facilities` can be responsible for JUST creating facilities.

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,277 @@
...
+        
+        expect(or_facilities[2].name).to eq("Astoria DMV Office")
+        expect(or_facilities[2].phone).to eq("503-325-3951")
+        expect(or_facilities[2].address).to eq("705 W Marine Dr, Astoria, OR 97103")
+        expect(or_facilities[2].registered_vehicles).to eq([])
+        expect(or_facilities[2].services).to eq([])
+        expect(or_facilities[2].collected_fees).to eq(0)
+      end
+
+      context "helper methods" do
```
epintozzi, line 79: Excellent! Super glad to see you testing your helpers.

README.md
```diff
@@ -1,3 +1,123 @@
...
+ ```
+ ...
+ ```ruby
+[14] pry(RSpec::ExampleGroups::DmvDataService)> DmvDataService.new.wa_ev_registrations[5][:make]
+=> "VOLKSWAGEN"
+```
+
+## Iteration 3
+
+- When creating a facility factory, it appears you can use `||` as a way to check for different keys but not `or`
```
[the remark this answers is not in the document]
garrettgregor: Interesting 🧐 🤔

README.md
```diff
@@ -1,3 +1,123 @@
...
+
+- What was easier than you expected about jumping in to an unfamiliar codebase? What made it easy? If nothing felt easy, what would’ve helped you feel more comfortable more quickly?
+
+  - Similar to above, what was easier about working with unfamiliar codebases was that I could pick some clues from the codebase to figure out something else that was going wrong. For instance, I took some clues from the vehicle class to help figure out the facility class. Working with a codebase was also easier because I had so much of the work done already. It nothing had felt easy, I think that chatting it through with another developer, or the former developer if that was an option, might have made it feel a little more familiar. Having a debugging tool to see what the code was doing was ***invaluable***. I'm sure it would have taken quite a bit longer to solve some of those bugs had it not been for `require 'pry'; binding.pry`.
+
+# Notes to Self
+
+## Iteration 1
+
+- Wtf is `require 'date'` in line 1 of the vehicle class? Is date a standard gem file that's being accessed in `Date.antique?`
```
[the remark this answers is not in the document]
garrettgregor: Definitely have a better understanding of this now! Thanks Erin!

PR #150, opened 2023-04-04
Garrett Varenhorst

GVarenhorst:


README.md
```diff
@@ -1,3 +1,15 @@
 # The DMV
 
 This is the starter repo for the BE Mod1 DMV project.
+Reflections:  
+# Q1-- Digging into code base. What was your process?
+
+Q1 Answer-- First I examined the lib files individually, side by side with it's respective spec file. Without touching the lib file code, I examined what the tests required, and wrote down on paper what I thought would need to be added or changed. Close file, rinse and repeat with the next file, until I had a working tree for reference of each file before actually testing. Then I run the test files and compare with my answers.
```
KatBrandt, line 7: Great job using the test files as a guide. That's exactly what they are, documentation of how the code should behave.

lib/dmv.rb
```diff
@@ -1,16 +1,19 @@
 class Dmv
-
+attr_reader :facilities, :service
   def initialize
     @facilities = []
+    @service = service
```
KatBrandt, line 5: We don't really need this attribute here, wasn't necessary.

lib/dmv.rb
```diff
@@ -1,16 +1,19 @@
...
   end
 
   def facilities_offering_service(service)
-    @facilities.find do |facility|
-      facility.services.include?(service)
+    select_service = []
+    @facilities.each do |facility|
+      select_service << facility if facility.services.include?(service)
     end
+    select_service
```
KatBrandt, line 17: This works. Nice job using the accumulator. Could also reach for the find_all enumerable to get rid of the conditional and the accumulator.

lib/facility.rb
```diff
@@ -1,14 +1,40 @@
...
+                :collected_fees
 
-  def initialize(name, address, phone)
-    @name = name
-    @address = address
-    @phone = phone
+  def initialize(facility_details)
+    @name = facility_details[:name]
+    @address = facility_details[:address]
+    @phone = facility_details[:phone]
```
KatBrandt, line 12: Yes, good job using hash methods to extract this data.

lib/facility.rb
```diff
@@ -1,14 +1,40 @@
...
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+  
+  def register_vehicle(vehicle)
+    if @services.include?('Vehicle Registration')
+      collect_fees(vehicle)
```
KatBrandt, line 24: Nice use of this helper method.

lib/registrant.rb
```diff
@@ -0,0 +1,21 @@
+class Registrant
+attr_reader :name, :age, :permit, :license_data
```
KatBrandt, line 2: Careful here, we don't want to expose methods we don't need. `permit` is not a method we need. We already have `permit?`

lib/registrant.rb
```diff
@@ -0,0 +1,21 @@
...
+    @permit = permit
+    @license_data = {
+      :written=>false,
+      :license=>false,
+      :renewed=>false
+    }
+  end
+
+  def permit?
+    @permit == true
```
KatBrandt, line 15: Since @permit holds either true or false, we don't need to do the comparison with true. We could just have `@permit` on line 15.

lib/vehicle.rb
```diff
@@ -22,4 +27,14 @@ def antique?
...
   end
+
+  def set_plate_type
+    if antique?
+      @plate_type = :antique
+    elsif electric_vehicle?
+      @plate_type = :ev
+    else
+      @plate_type = :regular
+    end
```
KatBrandt, line 38: yup, this method looks good.

spec/facility_spec.rb
```diff
@@ -24,3 +27,61 @@
...
+    @facility_2 = Facility.new({name: 'Ashland DMV Office', address: '600 Tolman Creek Rd Ashland OR 97520', phone: '541-776-6092' })
+    @cruz = Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice} )
+    @bolt = Vehicle.new({vin: '987654321abcdefgh', year: 2019, make: 'Chevrolet', model: 'Bolt', engine: :ev} )
+    @camaro = Vehicle.new({vin: '1a2b3c4d5e6f', year: 1969, make: 'Chevrolet', model: 'Camaro', engine: :ice} ) 
+  end
+  
+  describe '#initial service registration' do
+    it 'can add vehicle registration' do
+      
+      @facility_1.add_service('Vehicle Registration')
```
KatBrandt, line 45: This is good, and we'd want to start this test checking that the `@facility_1.services` starts off as an empty array. Then we do the rest of the test as your wrote to ensure it changes like we'd want.

spec/facility_spec.rb
```diff
@@ -24,3 +27,61 @@
...
+
+      expect(@bolt.registration_date).to eq(Date.today)
+      expect(@bolt.plate_type).to eq(:ev)
+
+      expect(@facility_1.registered_vehicles).to eq([@cruz, @camaro, @bolt]) 
+      expect(@facility_1.collected_fees).to eq(325)
+      expect(@facility_2.registered_vehicles).to eq([])
+      expect(@facility_2.services).to eq([])
+      expect(@facility_2.register_vehicle@bolt).to eq(nil)
+      expect(@facility_2.collected_fees).to eq(0)
```
KatBrandt, line 80: There's a lot going on in this test. In the future, lets try to break our tests apart more so that every method is being tested in it's own `it` block.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      expect(@registrant_1.age).to eq(18)
+      expect(@registrant_2.age).to eq(15)
+    end
+  end 
+    describe '#permit' do
+    it 'does not have a permit by default' do
+      
+
+      expect(@registrant_1.permit?).to eq(true)
+      expect(@registrant_2.permit?).to eq(false)
```
KatBrandt, line 35: Good job testing both of these since one uses the default value and one is given a value.

KatBrandt (commented):
Great start here Garrett. There was some learning in this project on boolean values and flow control that you didn't quite get to. I highly recommend meeting with student support a few times this weekend if you can, maybe to work through m ore of this project to ensure you get all the learning from it.

PR #152, opened 2023-04-04
Jimmy Jurado

jcjurado3:


dcoleman21 (commented):
Overall nice job on this project. I have left comments throughout for you to review. One big thing to really be mindful about is your indentation, spacing, and `ends`. As far as your testing goes, I think you did well. There were a few tests that felt hard to follow because they bounced around a bit. So my suggestion is that when a test is checking multiple objects, break it out into chunks where each chunk is focusing on the functionality of each object. See my comments in your `facility_spec`. Let me know if you have any questions. Keep up the good work.

PR #154, opened 2023-04-04
Kailey Kaes

kaileykaes:


lib/facility.rb
```diff
@@ -1,14 +1,79 @@
...
+        @collected_fees += 200 
+        vehicle.change_plate_type(:ev)
+      else
+        @collected_fees += 100 
+        vehicle.change_plate_type(:regular)
+      end
+      vehicle.change_registration_date
+      @registered_vehicles << vehicle 
+    end
+  end
```
KatBrandt, line 42: This nested flow control is where things can get confusing and slowed down. What if instead of that outer if statement, you used a guard clause? Something like `return nil if !can_perform("Vehicle Registration")`. That would be a single line guard clause to first just check that the facility can do the action. And if it can't, nil will be returned right away and it won't go down the rest of the method. Then you're next if statement can be it's own block and not nested.

lib/facility.rb
```diff
@@ -1,14 +1,79 @@
...
+  def renew_drivers_license(registrant)
+    if can_perform?('Renew License') 
+      registrant.license_data[:license] ? registrant.license_data[:renewed] = true : false
+    else
+      false
+    end
+  end
+
+  #helpers
+  def can_perform?(service)
```
KatBrandt, line 69: Love this helper method! Great keeping the code DRY. We call this an abstraction because you're abstracting the code into this method. Great abstraction.

lib/facility.rb
```diff
@@ -1,14 +1,79 @@
...
+        vehicle.change_plate_type(:regular)
+      end
+      vehicle.change_registration_date
+      @registered_vehicles << vehicle 
+    end
+  end
+
+  def administer_written_test(registrant)
+    if can_perform?('Written Test') 
+      registrant.age >= 16 && registrant.permit? ? registrant.license_data[:written] = true : false
```
KatBrandt, line 46: I do not love turinary statements in Ruby, especially when they start getting longer like this one. I'd rather see this written long-form, or break up the method into different parts so you don't have nested if statements.

lib/facility_factory.rb
```diff
@@ -0,0 +1,54 @@
...
+              :mo_facilities
+  
+  def initialize
+    @or_facilities = []
+    @ny_facilities = []
+    @mo_facilities = []
+  end
+
+  def create_or_facilities(dmv_data)
+    dmv_data.map do |facility|
```
KatBrandt, line 13: Overall nice job here. 
For your enumerable choice, I would have just done #each since you are not caring about the return value of the enumerable action, rather, you are putting all the new objects into the facility arrays so no map is needed.

lib/facility_factory.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def create_ny_facilities(dmv_data)
+    dmv_data.map do |facility|
+      address = [facility[:street_address_line_1], facility[:city], facility[:state], facility[:zip_code]]
+      daily_hours = {
+        monday: "#{facility[:monday_beginning_hours]} - #{facility[:monday_ending_hours]}",
+        tuesday: "#{facility[:tuesday_beginning_hours]} - #{facility[:tuesday_ending_hours]}",
+        wednesday: "#{facility[:wednesday_beginning_hours]} - #{facility[:wednesday_ending_hours]}",
+        thursday: "#{facility[:thursday_beginning_hours]} - #{facility[:thursday_ending_hours]}",
+        friday: "#{facility[:friday_beginning_hours]} - #{facility[:friday_ending_hours]}"
```
KatBrandt, line 31: nice job!

lib/vehicle.rb
```diff
@@ -22,4 +26,12 @@ def antique?
   def electric_vehicle?
     @engine == :ev
   end
+
+  def change_registration_date 
+    @registration_date = Date.today
+  end
+
+  def change_plate_type(plate_type)
```
KatBrandt, line 34: Yes, I like this! Much better than an attr_accessor since now the class itself is in charge of making those attribute changes.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,109 @@
...
+  def create_wa_vehicles(dmv_data)
+    dmv_data.map do |registration|
+      vehicles << Vehicle.new({
+        vin: registration[:vin_1_10], 
+        year: registration[:model_year].to_i, 
+        make: registration[:make],
+        model: registration[:model], 
+        engine: :ev
+      })
+    end.flatten!
```
KatBrandt, line 17: Again, we don't need the map here since you're just shoveling all the vehicle objects into the @vehicles array. To get the proper return value, you could just call `@vehicles` on the last line of this method, outside of the map (except I'd use each) :)

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,109 @@
...
+  
+  #helpers 
+  
+  def engine_type(engine_type)
+    if engine_type == 'GAS' || engine_type == 'DIESEL' || engine_type == 'PROPANE' || engine_type == 'FLEX'
+      :combustion
+    elsif engine_type == 'ELECTRIC'
+      :ev
+    else
+      :wtf
```
KatBrandt, line 28: lol

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,109 @@
...
+        model: registration[:model], 
+        engine: :ev
+      })
+    end.flatten!
+  end
+  
+  #helpers 
+  
+  def engine_type(engine_type)
+    if engine_type == 'GAS' || engine_type == 'DIESEL' || engine_type == 'PROPANE' || engine_type == 'FLEX'
```
KatBrandt, line 23: Nice job here

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,109 @@
...
+  end
+  
+  def most_popular_model
+    model_vehicles = @vehicles.sort_by(&:model)
+    count_by_model = Hash.new(0)
+    model_vehicles.map do |vehicle|
+      count_by_model[vehicle.model] += 1 
+    end
+    most_pop_model = count_by_model.max_by{|model, amount| amount}.first
+    most_pop_model
```
KatBrandt, line 56: This is a very clever solution! You're using all these enumerables nicely and I really like that you built this hash to serve as a tool to help you count objects (remember this for technical challenges in mod 4!). 
There are a few different enumerables you can start to play with that would have helped you with this. One of them is #group_by and the other is #max_by.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,109 @@
...
+  end
+
+  def most_popular_model
+    model_vehicles = @vehicles.sort_by(&:model)
+    count_by_model = Hash.new(0)
+    model_vehicles.map do |vehicle|
+      count_by_model[vehicle.model] += 1 
+    end
+    most_pop_model = count_by_model.max_by{|model, amount| amount}.first
+    most_pop_model
```
KatBrandt, line 76: Why are these two methods in this class twice?

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,109 @@
...
+    make_vehicles.map do |vehicle|
+      count_by_make[vehicle.make] += 1 
+    end
+    most_pop_make = count_by_make.max_by{|make, amount| amount}.first
+    "#{most_pop_make}"
+  end
+
+  def count_by_model_year(yr)
+    count = @vehicles.count{|vehicle| vehicle.year == yr}
+    count
```
KatBrandt, line 101: Nice use of the #count enumerable!!! You shouldn't have to store this in a variable though, the return value is the count.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,109 @@
...
+  end
+
+  def most_registered_county(dmv_data)
+    counter = Hash.new(0)
+    dmv_data.map {|registration| counter[registration[:county]] += 1  }
```
KatBrandt, line 106: Since you're just putting everyting into your accumulator (hash), you wouldn't use a map here, you'd want an #each. 
Great job with this method!

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+
+  describe '#initialize' do 
+    it 'exists' do 
+      expect(@factory).to be_a FacilityFactory
+    end
+
+    it 'begins without facilities of any type' do 
+      expect(@factory.or_facilities).to eq([])
+      expect(@factory.ny_facilities).to eq([])
+      expect(@factory.mo_facilities).to eq([])
```
KatBrandt, line 19: Once you check this for one factory instance, really no need to check for all the factory instances. Over testing.

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+    end
+  end
+
+  describe '#create_or_facilities' do 
+    it 'create facilities from data' do 
+      @factory.create_or_facilities(@oregon_facilities)
+      expect(@factory.or_facilities).to_not be_empty
+      expect(@factory.or_facilities[0]).to be_a Facility
```
KatBrandt, line 27: Awesome test.

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+    end
+
+    it 'facilities instantiate with attributes' do 
+      @factory.create_or_facilities(@oregon_facilities)
+      expect(@factory.or_facilities[0].name).to eq('Albany DMV Office')
+      expect(@factory.or_facilities[0].address).to eq('2242 Santiam Hwy SE Albany OR 97321')
+      expect(@factory.or_facilities[0].phone).to eq('541-967-2014')
+    end
+  end
+  describe '#create_or_facilities' do
```
KatBrandt, line 37: repeated test?

spec/facility_spec.rb
```diff
@@ -2,25 +2,192 @@
...
-      expect(@facility.services).to eq([])
-      @facility.add_service('New Drivers License')
-      @facility.add_service('Renew Drivers License')
-      @facility.add_service('Vehicle Registration')
-      expect(@facility.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
+      expect(@facility_1.services).to eq([])
+      @facility_1.add_service('New Drivers License')
+      @facility_1.add_service('Renew Drivers License')
+      @facility_1.add_service('Vehicle Registration')
+      expect(@facility_1.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
```
KatBrandt, line 61: Great job testing the initial value and then testing again after making changes.

spec/facility_spec.rb
```diff
@@ -2,25 +2,192 @@
...
+      expect(@facility_1.registered_vehicles).to eq([@cruz])
+      expect(@facility_1.collected_fees).to eq(100)
+      #camaro
+      @facility_1.register_vehicle(@camaro)
+      expect(@camaro.registration_date).to be_a Date
+      expect(@camaro.plate_type).to eq(:antique)
+      #bolt
+      @facility_1.register_vehicle(@bolt)
+      expect(@bolt.registration_date).to be_a Date
+      expect(@bolt.plate_type).to eq(:ev)
```
KatBrandt, line 84: Ok, I can see why you tested all three vehicles since they're each creating a different plate type.

spec/facility_spec.rb
```diff
@@ -2,25 +2,192 @@
...
+      expect(@facility_1.registered_vehicles).to eq([@cruz, @camaro, @bolt])
+      expect(@facility_1.collected_fees).to eq(325)
+    end
+
+    it 'no register if no can register' do
```
KatBrandt, line 89: I'm not sure what this test name is referring to.

spec/facility_spec.rb
```diff
@@ -2,25 +2,192 @@
...
+    end
+    
+    it 'no written tests for the very young' do 
+      @facility_1.add_service('Written Test')
+      expect(@registrant_3.age).to eq(15)
+      expect(@registrant_3.permit?).to be false
+      expect(@facility_1.administer_written_test(@registrant_3)).to be false
+      @registrant_3.earn_permit
+      expect(@facility_1.administer_written_test(@registrant_3)).to be false
+      expect(@registrant_3.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
```
KatBrandt, line 126: Great job testing all these different conditions of the written test.

spec/facility_spec.rb
```diff
@@ -2,25 +2,192 @@
...
+      @facility_1.administer_written_test(@registrant_3)
+      @facility_1.administer_road_test(@registrant_3)
+      expect(@registrant_3.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+      expect(@facility_1.renew_drivers_license(@registrant_3)).to be false
+      expect(@registrant_3.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+    end
+  end
+
+  describe 'helpers' do 
+    it 'can_perform?' do
```
KatBrandt, line 187: Yay! Good job testing your helper methods.

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+
+RSpec.describe VehicleFactory do
+  before(:each) do
+    @factory = VehicleFactory.new
+    @wa_ev_registrations = DmvDataService.new.wa_ev_registrations
+    @ny_registrations = DmvDataService.new.ny_registrations
+  end
+
+  describe '#initialize' do 
+    xit 'exists' do
```
KatBrandt, line 11: Why are all these tests skipped?

KatBrandt (commented):
Overall, really great job with this!!! Seems like you are understanding enumerables, flow control, and working with classes and objects, hashes, arrays, etc.

PR #156, opened 2023-04-04
Jeffrey Redish

Jredish11:


lib/facility.rb
```diff
@@ -1,14 +1,72 @@
...
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    return nil unless services.include?("Vehicle Registration")
+    vehicle.registration_date = Date.today
```
dcoleman21, line 30: I totally think this could live here. One other thing to think about is could this live in the `Vehicle` class as a helper method that checks if it a vehicle has been registered and then set the date? But nice job!

lib/facility.rb
```diff
@@ -1,14 +1,72 @@
...
+    if vehicle.antique?
+      @collected_fees += 25
+      vehicle.plate_type = :antique
```
dcoleman21, line 33: I would probably have a stand alone method in your `Vehicle` class that assigns the plate type. This would help DRY up this method. So essentially you could
```
if vehicle.antique?
   @collected_fees += 25
   @collected_fees += 200
else
   @collected_fees += 100
end
```

lib/facility_factory.rb
```diff
@@ -0,0 +1,18 @@
+class FacilityFactory
+  def initialize
```
dcoleman21, line 2: I don't think that you even need this `initialize` method. As it is, when a `FacilityFactory` is created it has no attributes. So if you were to remove this everything should still pass.

lib/facility_factory.rb
```diff
@@ -0,0 +1,18 @@
...
+    
+  end
+
+  def create_facility(dmv)
```
dcoleman21, line 6: I think that this method is a great starting point. And yes, using these `||` conditions can feel a bit clunky especially if we were to add more datasets. So one thing that you can consider is creating some helper methods in here that take care of grabbing the data that you want based on how it is formatted. So an example could be:
```
def format_address(data)
   if data[:website] && data[:website].include?('oregon')
      address = JSON.parse(data[:location_1][:human_address], symbolize_names: true)
      "#{address[:address]}, #{address[:city]}, #{address[:state]} #{address[:zip]}"
    elsif data[:state] && data[:state] == 'NY'
      "#{data[:street_address_line_1]}, #{data[:city]}, #{data[:state]} #{data[:zip_code]}"
    else data[:state] && data[:state] == 'missouri'
      "#{data[:address1]}, #{data[:city]}, #{data[:state]}, #{data[:zipcode]}"
    end
end
```
This would help with formatting the addresses based on how they come in via the external data source. You could create additional helpers for name, phone_number. You could then create another helper method that calls on all of the other helpers:
```
def format_facility_data(facility)
    data = {
      name: format_name(facility),
      address: format_address(facility),
      phone: format_phone_number(facility)
    }
end
```
You would then call this method inside of your `create_facilities` method. Let me know if you have questions.

lib/vehicle.rb
```diff
@@ -5,14 +5,18 @@ class Vehicle
...
+  
+  attr_accessor :registration_date, :plate_type
 
   def initialize(vehicle_details)
     @vin = vehicle_details[:vin]
     @year = vehicle_details[:year]
     @make = vehicle_details[:make]
     @model = vehicle_details[:model]
     @engine = vehicle_details[:engine]
+    @plate_type = vehicle_details[:plate_type]
```
dcoleman21, line 18: I wouldn't put this as a static instance variable inside here. Instead make it its own method that is responsible for setting the plate type based on its age (you can still use the `antique?` and `electric_vehicle?` methods inside.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,37 @@
...
+      Vehicle.new({
+        engine: :ev,
+        make:  car[:make],
+        model: car[:model],
+        vin: car[:vin_1_10],
+        year: car[:model_year]
+        })
+    end
+  end
+
```
dcoleman21, line 17: Really nice job using some of the other enumerables that we don't necessarily instruct on!

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+RSpec.describe FacilityFactory do
+  describe "Iteration 3" do
+    it "exists" do
+      facility = FacilityFactory.new
+      
+      expect(facility).to be_a(FacilityFactory)
+    end
+
+    it "creates new facilities from oregon data set" do
```
dcoleman21, line 11: These tests are really good. Just be careful expecting them to eq the exact values. Since we are pulling this info from an external source, if that source changes your tests will break. Instead you can test to datatypes. So for example on line 16 you could expect it to `be_a(String)`.

lib/registrant.rb
```diff
@@ -0,0 +1,23 @@
...
+      :renewed => false,
+      }
+  end
+
+  def permit?   
+    @permit
+  end
+
+  def earn_permit
+    @permit = true
```
dcoleman21, line 21: You can make this even more robust by adding a check that `@permit = true if @age >= 15`.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+      expect(registrant_1.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+      expect(registrant_2.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+    end
+    
+    it "can earn_permit" do
+      registrant_1 = Registrant.new('Bruce', 18, true )
+      registrant_2 = Registrant.new('Penny', 15 )
+
+      registrant_2.earn_permit
+
```
dcoleman21, line 38: You can make this even more robust by possibly creating another `registrant` who's age is lower than 15 and trying to allow them to `earn_permit` to really check that your method is doing what you want.

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,86 @@
+require 'spec_helper'
+
+RSpec.describe VehicleFactory do
+  describe "Iteration 2: Reading external data sets" do
+    it "exists" do
+      factory = VehicleFactory.new
+
+      expect(factory).to be_an_instance_of(VehicleFactory)
+    end
+    it "create vehicles using wa dmv data set" do
```
dcoleman21, line 10: Same thing here for your expect return values. Just make them datatypes vs the actual info. You will learn more about how to avoid this in Mod 3 using mocking and stubbing.

dcoleman21 (commented):
Jeff, really great job! Your testing is looking fantastic! Keep finding ways to sneak in some sad path/edge case testing. It is very apparent how much you have learned and are truly understanding. I left comments for you to review. Let me know if you have any questions. Keep up the great work!

PR #157, opened 2023-04-04
Blaine Glasgow DMV

GlowMunch:


lib/facility.rb
```diff
@@ -1,14 +1,88 @@
...
+  def initialize(name:, 
+                 address:, 
+                 phone:
+                 )
     @name = name
     @address = address
     @phone = phone
     @services = []
+    @registered_vehicles = []
+    @collected_fees = (0)
```
KatBrandt, line 18: Don't need the parenthesis around the 0

lib/vehicle.rb
```diff
@@ -22,4 +30,24 @@ def antique?
...
   end
+
+  def register
+    @registration_date = Date.new
+    plate
+  end
+
+  def plate
+    age = 2023 - @year
+    if (engine == :ice) && (age < 25)
```
KatBrandt, line 41: could use the antique? method instead of the age < 25

lib/facility.rb
```diff
@@ -1,14 +1,88 @@
...
-  attr_reader :name, :address, :phone, :services
+  attr_reader :name, 
+              :address, 
+              :phone, 
+              :services,
+              :registered_vehicles,
+              :collected_fees
 
-  def initialize(name, address, phone)
+  def initialize(name:,
```
KatBrandt, line 9: You're using here, what's called keyword arguments. I really like that! Keeps it very clear what each argument is.

KatBrandt (commented):
Overall, great job working through this code. We'll want to work on refactoring, breaking apart nested if statements, and writing more focused, robust tests.

PR #159, opened 2023-04-05
Matthew Johnson

MWMJohnson:
Had trouble adding my last branch, please let me know if you have any questions.


lib/facility.rb
```diff
@@ -1,14 +1,73 @@
...
+    @services
   end
+
+  def collection_fee_assignment
+    {
+    :ev => 200,
+    :antique => 25,
+    :regular => 100
+    }
+  end
```
epintozzi, line 29: I really like this implementation of managing fees. It would make your code super easy to change in the future if the fees ever change. It also doesn't rely on if statement. A hash is a really nice choice here. Great work!

lib/facility.rb
```diff
@@ -1,14 +1,73 @@
...
+
+  def register_vehicle(vehicle)
+    if @services.include?('Vehicle Registration')
+      @registered_vehicles << vehicle
+      vehicle.update_registration_date
+      @collected_fees += collection_fee_assignment[vehicle.plate_type]
+      @registered_vehicles
+    else 
+      nil
```
epintozzi, line 38: This is not explicitly necessary. If the `if` condition is not met, by default this method will do nothing and return `nil`.

lib/vehicle.rb
```diff
@@ -22,4 +30,16 @@ def antique?
   def electric_vehicle?
     @engine == :ev
   end
+
+  def plate_type
+    if self.electric_vehicle?
+      :ev 
+    elsif self.antique?
```
epintozzi, line 37:
```suggestion
    if electric_vehicle?
      :ev 
    elsif antique?
```
While not functionally incorrect, we don't need to use `self` here. Ruby automatically will perform these methods on the object on which they were called.

lib/vehicle.rb
```diff
@@ -5,14 +5,22 @@ class Vehicle
...
     @year = vehicle_details[:year]
     @make = vehicle_details[:make]
     @model = vehicle_details[:model]
     @engine = vehicle_details[:engine]
+    @registration_date = vehicle_details[:registration_date]
+  end
+
+  def update_registration_date
+    @registration_date = Date.today
   end
```
epintozzi, line 24: I like this method. I personally don't have an issue with `attr_accessors` however this gives us more control if we ever want to add in validation before setting the date. It's also a little more readable than using an accessor which is nice.

lib/facility.rb
```diff
@@ -1,14 +1,73 @@
...
+    end
+  end
+
+  def renew_drivers_license(registrant)
+    return false if !@services.include?('Renew License')
+    if registrant.license_data[:license] == true
+      registrant.update_renew_status
+      true
+    elsif registrant.license_data[:license] == false
+      false
```
epintozzi, line 69: Same as above regarding the `elsif`

lib/facility.rb
```diff
@@ -1,14 +1,73 @@
...
+      registrant.update_license_status
+      true
+    elsif registrant.license_data[:written] == false
+      false
+    end
+  end
+
+  def renew_drivers_license(registrant)
+    return false if !@services.include?('Renew License')
+    if registrant.license_data[:license] == true
```
epintozzi, line 65:
```suggestion
    if registrant.license_data[:license]
```
Same as above: `license_data[:license]` will already return true or false, so it's not explicitly necessary to do a comparison and check that `true == true`

lib/facility_factory.rb
```diff
@@ -0,0 +1,37 @@
...
+        
+        phone: facility[:phone_number] || facility[:phone] ||= "N/A",
+        services: [],
+        registered_vehicles: [],
+        collected_fees: 0
+        }
+      )
+    end
+    
+  end
```
epintozzi, line 32: This method is doing a lot. It's determining how to format data based on source, mapping and then creating facilities. I would like to see some helpers used here to make it a little easier to test and a whole lot easier to read:
```ruby
  def create_facilities(data)
    data.map do |facility|
      Facility.new(
        {
        name: format_name(facility),
        address: format_address(facility),
        phone: format_phone(facility),
        services: [],
        registered_vehicles: [],
        collected_fees: 0
        }
      )
    end
  end

  def format_name(facility)
    facility[:title] || facility[:office_name] || facility[:name]
  end

  def format_address(facility)
    if !facility[:location_1].nil?
      formatted_address = JSON.parse(facility[:location_1][:human_address], symbolize_names: true)
      "#{formatted_address[:address]}, #{formatted_address[:city]}, #{formatted_address[:state]} #{formatted_address[:zip]}"
    elsif !facility[:street_address_line_1].nil?
      "#{facility[:street_address_line_1]}, #{facility[:city]}, #{facility[:state]} #{facility[:zip_code]}"
    elsif !facility[:address1].nil?
      "#{facility[:address1]}, #{facility[:city]}, #{facility[:state]} #{facility[:zipcode]}"
    end
  end

  def format_phone(facility)
    facility[:phone_number] || facility[:phone] ||= "N/A"
  end
```

epintozzi (commented):
Great work! I didn't comment on specific test files. Those look good to me. I think you did a really nice job testing all your helpers 🔥

PR #160, opened 2023-04-05
Iteration 1 branch

lorenzoriv:


lib/registrant.rb
```diff
@@ -0,0 +1,22 @@
+class Registrant
+  binding.pry
```
KatBrandt, line 2: Make sure to remove all pry's before pushing up code. :)

lib/registrant.rb
```diff
@@ -0,0 +1,22 @@
...
+    end
+
+    def permit?
+      @permit
+    end
+
+    def earn_permit
+      @permit = true
+    end
+    def license_data
```
KatBrandt, line 19: This method is not necessary since you defined this exact method using the attr_reader :license_data.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    expect(@registrant_1.age).to eq(18)
+    expect(@registrant_1.permit).to be true
+    expect(@registrant_2.permit).to be false
+  end
+end
+
+describe '#permit?' do
+  it 'can determine if a registrant has a permit' do 
+    expect(@registrant_1.permit?).to be true
+    expect(@registrant_2.permit?).to be false
```
KatBrandt, line 22: Great test

spec/registrant_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+  end
+end
+
+describe '#earn_permit' do
+  it 'updates permit status to true' do 
+    expect(@registrant_2.permit?).to be false
+
+    @registrant_2.earn_permit
+
+    expect(@registrant_2.permit?).to be true
```
KatBrandt, line 38: Great test, testing initial value, then changing it and retesting.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+  end
+end
+
+describe '#written_pass' do 
+  it 'updates license data after written test' do 
+    expect(@registrant_1.license_data[:written]).to be false
+
+    @registrant_1.written_pass
+
+    expect(@registrant_1.license_data[:written]).to be true
```
KatBrandt, line 48: I'm guessing this test fails since #written_pass is not a method defined in your Registrant class. Also, the interaction pattern didn't ask for this method. Be careful to only write the tests for the code that the interaction pattern asks for.

PR #161, opened 2023-04-05
Will Zale

WZale:


lib/facility.rb
```diff
@@ -1,14 +1,69 @@
...
+    else 
+      nil
+    end
+  end
+
+  def administer_written_test(registrant)
+    if !self.services.include?("Written Test") || registrant.age < 16
+      false
+    elsif self.services.include?("Written Test")
+      registrant.license_data[:written] = true
```
dcoleman21, line 45: This could be a refactor where you create a helper method in your `Registrant` class that checks if a registrant is old enough, has been tested, and then updates the `license_data[:written] = true`. Then you can call `registrant.your_helper_method`

lib/facility.rb
```diff
@@ -1,14 +1,69 @@
...
+      true
+    end
+  end
+
+  def administer_road_test(registrant)
```
dcoleman21, line 50: See my comment above. Similar format plus another possible helper method in your `Registrant` class that checks the above stated.

lib/facility.rb
```diff
@@ -1,14 +1,69 @@
...
+    if !self.services.include?("Road Test") 
+      false
+    elsif registrant.license_data[:written] == false
+      false
+    elsif self.services.include?("Road Test")
+      registrant.license_data[:license] = true
+    end
+  end
+
+  def renew_drivers_license(registrant)
```
dcoleman21, line 60: See my comment in your `administer_written_test` method. Similar format plus another possible helper method in your `Registrant` class that checks the above stated.

lib/registrant.rb
```diff
@@ -0,0 +1,26 @@
...
+                    :renewed => false
+                  }
+  end
+
+  def permit?
+    @permit
+  end
+
+  def earn_permit
+    @permit = true
```
dcoleman21, line 24: You might make this more robust by adding an `if` statement that checks that they can earn a permit `if @age >= 15`

lib/vehicle.rb
```diff
@@ -1,18 +1,23 @@
...
+  attr_accessor :registration_date,
+                :plate_type
 
   def initialize(vehicle_details)
     @vin = vehicle_details[:vin]
     @year = vehicle_details[:year]
     @make = vehicle_details[:make]
     @model = vehicle_details[:model]
     @engine = vehicle_details[:engine]
+    @plate_type = vehicle_details[:plate_type]
```
dcoleman21, line 19: I would not have this here. I would just use your `plate_type` method below, it is more dynamic.

lib/vehicle.rb
```diff
@@ -22,4 +27,11 @@ def antique?
   def electric_vehicle?
     @engine == :ev
   end
+
+  def plate_type
+    @plate_type = if antique? then :antique
```
dcoleman21, line 32: If you remove `@plate_type` from your initialize then this method then your `if/else` logic could look like:
```
if antique?
   :antique
elsif @engine == :ev 
   :ev 
else
   :regular
end
```

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,14 @@
+class VehicleFactory
+
+  def create_vehicles(vehicles)
```
dcoleman21, line 3: Nice job here 😉

spec/facility_spec.rb
```diff
@@ -2,25 +2,157 @@
...
+      @facility_1.add_service('Vehicle Registration')
+
+      expect(@facility_1.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
+    end
+  end
+
+  describe '#register vehicles' do
+    it 'can register a vehicle' do
+      @facility_1.add_service('Vehicle Registration')
+      expect(@cruz.registration_date).to eq(nil)
```
dcoleman21, line 42: This expect line is calling `registration_date` on a `Vehicle` object so you would need to have a test in your `vehicle_spec` that tests this. Good rule of thumb is to look at what object a method is being called on to guide you to where that test should live.

spec/facility_spec.rb
```diff
@@ -2,25 +2,157 @@
...
+      expect(@facility_1.registered_vehicles).to eq([])
+      expect(@facility_1.collected_fees).to eq(0)
```
dcoleman21, line 44: I would probably move anything that has to do with `collected_fees` to its own test in here.

spec/facility_spec.rb
```diff
@@ -2,25 +2,157 @@
...
+      expect(@bolt.plate_type).to eq(:ev)
+
+      expect(@facility_1.registered_vehicles).to eq([@cruz, @camaro, @bolt])
+      expect(@facility_1.collected_fees).to eq(325)
+    end
+
+    it 'cannot register a vehicle that is already registered at another facility' do
+      @facility_1.register_vehicle(@bolt)
+      expect(@bolt.registration_date).to eq(Date.today)
+      expect(@bolt.plate_type).to eq(:ev)
```
dcoleman21, line 67: Same thing here. The `plate_type` method is being called on a `Vehicle` object so it should live in a test within your `vehicle_spec`.

spec/facility_spec.rb
```diff
@@ -2,25 +2,157 @@
...
+      expect(@facility_2.register_vehicle(@bolt)).to eq(nil)
+      expect(@facility_2.registered_vehicles).to eq([])
+      expect(@facility_2.collected_fees).to eq(0)
+    end
+  end
+
+  describe '#administer tests' do
+    it 'can administer a written test' do
+      expect(@registrant_1.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+      expect(@registrant_1.permit).to eq(true)
```
dcoleman21, line 81: This should live in a test in your `registrant_spec`.

spec/facility_spec.rb
```diff
@@ -2,25 +2,157 @@
...
+      expect(@facility_1.administer_written_test(@registrant_1)).to eq(false)
+      expect(@registrant_1.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
```
dcoleman21, line 83: This should live in a test within your `registrant_spec`.

spec/facility_spec.rb
```diff
@@ -2,25 +2,157 @@
...
+
+      @facility_1.add_service('Written Test')
+
+      expect(@facility_1.administer_written_test(@registrant_1)).to eq(true)
+      expect(@registrant_1.license_data).to eq({:written=>true, :license=>false, :renewed=>false})
+      
+      @registrant_2.earn_permit
```
dcoleman21, line 90: This can live in this test because it is a command that will change the status of your `administer_written_test` method below.

spec/facility_spec.rb
```diff
@@ -2,25 +2,157 @@
...
+
+      @facility_1.add_service('New Drivers License')
+      @facility_1.add_service('Renew Drivers License')
+      @facility_1.add_service('Vehicle Registration')
+
+      expect(@facility_1.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
+    end
+  end
+
+  describe '#register vehicles' do
```
dcoleman21, line 39: See my comments below. See what object a method is being called on - that is where the test should live.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+      expect(@registrant_1).to be_an_instance_of(Registrant)
+      expect(@registrant_1.name).to eq('Bruce')
+      expect(@registrant_1.age).to eq(18)
+      expect(@registrant_1.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+    end
+  end
+
+  describe '#permits' do
+    it 'has a permit? method' do
+      expect(@registrant_1.permit?).to eq(true)
```
dcoleman21, line 20: I would have both registrants in this test so that you can check `permit?` on both.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+
+    it 'has an earn permit method' do
+      @registrant_2.earn_permit
```
dcoleman21, line 24: I would include both registrants in this test as well because then you can truly check that your `earn_permit` method is doing what you expect. It should not change `permit` unless they are the correct age.

dcoleman21 (commented):
Overall Will, nice job. This was a big project and you did a lot of things well. You are understanding more than you believe 😄 . I have left comments for you to review. Please let me know if you have any questions.

PR #163, opened 2023-04-07
Jesse Sorman

Jesse193:
This includes iteration 1, iteration 2, and iteration 3

### DIFF
--- a/lib/facility_maker.rb
+++ b/lib/facility_maker.rb
@@ -1,0 +1,50 @@
+class FacilityMaker
+  attr_reader :add_facility, :facility, :add_facility, :facility_info, :name
+  attr_accessor :name, :address, :phone, :facilities, :add_facility
+
+  def new_facility(state_facilities)
+    state_facilities.map do |state_facility|
+      gen_facility = Facility.new({name: new_facility_name(state_facility), address: new_facility_address(state_facility), 
+      city: state_facility[:city], state: state_facility[:state], zipcode: new_facility_zipcode(state_facility), 
+      phone: new_facility_phone(state_facility)})
+    end
+  end
+
+  def new_facility_name(state_facility)
+    if state_facility[:website] && state_facility[:website].include?("oregon")
+      state_facility[:title]
+    elsif state_facility[:state] && state_facility[:state] == "NY"
+      state_facility[:office_name]
+    else state_facility[:state] && state_facility[:state] == "MO"
+      state_facility[:name]
+    end
+  end
+
+  def new_facility_address(state_facility)
+    if state_facility[:website] && state_facility[:website].include?("oregon")
+      JSON.parse(state_facility[:location_1][:human_address])
+    elsif state_facility[:state] && state_facility[:state] == "NY"
+      state_facility[:street_address_line_1]
+    else state_facility[:state] && state_facility[:state] == "MO"
+      state_facility[:address1]
+    end
+  end
+
+  def new_facility_zipcode(state_facility)
+    if state_facility[:state] && state_facility[:state] == "MO"
+      state_facility[:zipcode]
+    else state_facility[:state] && state_facility[:state] == "NY"
+      state_facility[:zip_code]
+    end
+  end
+
+  def new_facility_phone(state_facility)
+    if state_facility[:website] && state_facility[:website].include?("oregon")
+      state_facility[:phone_number]
+    elsif state_facility[:state] && state_facility[:state] == "NY"
+      state_facility[:public_phone_number]
+    else state_facility[:state] && state_facility[:state] == "MO"
+      state_facility[:phone]
+    end
+  end
+end
--- a/lib/registrant.rb
+++ b/lib/registrant.rb
@@ -22,6 +22,4 @@ class Registrant
       @permit = false
     end
   end
-
-  
 end
--- a/spec/registrant_spec.rb
+++ b/spec/registrant_spec.rb
@@ -9,7 +9,6 @@ RSpec.describe Registrant do
     expect(@registrant_1.age). to eq(18)
     expect(@registrant_1.permit?).to eq(true)
     expect(@registrant_1.license_data).to eq({:written => false, :license => false, :renewed => false})
-
     expect(@registrant_2.name).to eq("Penny")
     expect(@registrant_2.age).to eq(15)
     expect(@registrant_2.permit?).to eq(false)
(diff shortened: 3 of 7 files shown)


lib/registrant.rb
```diff
@@ -0,0 +1,25 @@
...
+  
+  def license_data
+    @license_data
+  end
+
+  def earn_permit
+    if @age >= 16
+      @permit = true
+    else
+      @permit = false
```
epintozzi, line 22: Since `@permit` defaults to `false` we only need the `if` portion here where we set it to true. If the condition is not met (`@age >= 16`) then nothing changes and it will remain false as set by the default value.

lib/facility.rb
```diff
@@ -1,14 +1,89 @@
 class Facility
-  attr_reader :name, :address, :phone, :services
-
-  def initialize(name, address, phone)
-    @name = name
-    @address = address
-    @phone = phone
-    @services = []
+  attr_reader :name, :address, :phone, :services, :registration_date, :vehicle_details, :license_data
+  attr_accessor :registration_date, :name, :address, :phone
```
epintozzi, line 3: Some of these aren't attributes in this class so we don't need them.
```suggestion
  attr_reader :name, :address, :phone, :services
  attr_accessor 
```

spec/registrant_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    expect(@registrant_1.name).to eq("Bruce")
+    expect(@registrant_1.age). to eq(18)
+    expect(@registrant_1.permit?).to eq(true)
+    expect(@registrant_1.license_data).to eq({:written => false, :license => false, :renewed => false})
+    expect(@registrant_2.name).to eq("Penny")
+    expect(@registrant_2.age).to eq(15)
+    expect(@registrant_2.permit?).to eq(false)
+    expect(@registrant_2.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+    @registrant_2.earn_permit
+    expect(@registrant_2.permit?).to eq(true)
```
epintozzi, line 17:
```suggestion
    expect(@registrant_2.permit?).to eq(false)
```
I think you want this to be false? Your code says the registrant needs to be 16 or older. If this should actually be true, then the program code should be updated to `@age >= 15`

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,13 @@
+class VehicleFactory
+  attr_reader :vehicle
+  def initialize
+    @vehicles = []
+  end
+
+  def create_vehicles(vehicles)
+    vehicles.map do |vehicle|
+      car = Vehicle.new(vin: vehicle[:vin_1_10], year: vehicle[:model_year], make: vehicle[:make], model: vehicle[:model], engine: :ev)
```
epintozzi, line 9: `Vehicle` initializes with a single argument - a hash. You are passing each attribute in individually. It should probably look a little more like:
```ruby
      car = Vehicle.new(
        {
          vin: vehicle[:vin_1_10],
          year: vehicle[:model_year],
          make: vehicle[:make],
          model: vehicle[:model],
          engine: :ev
        }
      )
```

lib/facility_maker.rb
```diff
@@ -0,0 +1,50 @@
+class FacilityMaker
+  attr_reader :add_facility, :facility, :add_facility, :facility_info, :name
+  attr_accessor :name, :address, :phone, :facilities, :add_facility
```
epintozzi, line 3: This class has no attributes so we shouldn't need or have any attr readers or accessors

lib/facility_maker.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def new_facility(state_facilities)
+    state_facilities.map do |state_facility|
+      gen_facility = Facility.new({name: new_facility_name(state_facility), address: new_facility_address(state_facility), 
+      city: state_facility[:city], state: state_facility[:state], zipcode: new_facility_zipcode(state_facility), 
+      phone: new_facility_phone(state_facility)})
+    end
+  end
```
epintozzi, line 11: This would be a lot easier to read if you added in some line breaks:
```ruby
  def new_facility(state_facilities)
    state_facilities.map do |state_facility|
      Facility.new(
        {
          name: new_facility_name(state_facility),
          address: new_facility_address(state_facility),
          city: state_facility[:city],
          state: state_facility[:state],
          zipcode: new_facility_zipcode(state_facility),
          phone: new_facility_phone(state_facility)
        }
    )
    end
  end
```
Functionally, I changed nothing in your code. All I did was add in some line breaks so I could more easily read and digest what you're trying to do.

epintozzi: I like that you created helpers for formatting the data. I might call this method `create_facilities` instead of `new_facility` because it's creating many, not just one but overall you broke this out really nicely.

lib/facility_maker.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def new_facility_name(state_facility)
+    if state_facility[:website] && state_facility[:website].include?("oregon")
+      state_facility[:title]
+    elsif state_facility[:state] && state_facility[:state] == "NY"
+      state_facility[:office_name]
+    else state_facility[:state] && state_facility[:state] == "MO"
```
epintozzi, line 18:
```suggestion
    elsif state_facility[:state] && state_facility[:state] == "MO"
```
This should be an `elsif` not an `else`. Ruby doesn't evaluate anything to the right of an `else`. Else is catch all - none of the previous conditions were met so we do something by default. If we care about conditions, we should use an `elsif`. Same for lines 28, 36, and 46.

lib/facility_maker.rb
```diff
@@ -0,0 +1,50 @@
...
+  def new_facility(state_facilities)
+    state_facilities.map do |state_facility|
+      gen_facility = Facility.new({name: new_facility_name(state_facility), address: new_facility_address(state_facility), 
+      city: state_facility[:city], state: state_facility[:state], zipcode: new_facility_zipcode(state_facility), 
+      phone: new_facility_phone(state_facility)})
+    end
+  end
+
+  def new_facility_name(state_facility)
+    if state_facility[:website] && state_facility[:website].include?("oregon")
```
epintozzi, line 14: It's really smart that you are checking both `state_facility[:website]` and 
`state_facility[:website].include?("oregon")`. It helps us catch bad data and avoid our program from returning errors.

epintozzi (commented):
I'm getting 4 errors when I run your tests, including that the tests just totally stall out when trying to run `vehicle_factory_spec`. I'm not sure what's happening there. Is this in line with what you would have expected me to see? Just want to make sure I was reviewing the correct branch.

You also asked for feedback on your commits. Your commit messages look good to me. Short and sweet and communicate the work you are adding with the commit.

PR #165, opened 2023-04-07
Jorja final iteration

JorjaF:


lib/facility.rb
```diff
@@ -1,14 +1,64 @@
 class Facility
-  attr_reader :name, :address, :phone, :services
-
-  def initialize(name, address, phone)
+  attr_reader :name, :address, :phone,
```
KatBrandt, line 2: So for attr_readers, we want them either all on one line OR, one line for each - listed vertically. This format is too hard to read. I like the vertical format where each attr_reader would be on it's own line.

lib/facility.rb
```diff
@@ -1,14 +1,64 @@
...
+              :services, :registered_vehicles, :collected_fees
+              :administer_written_test
+  def initialize(name:, address:, phone:)
     @name = name
     @address = address
     @phone = phone
     @services = []
+    @registered_vehicles = []
+    @collected_fees = 0
+    @administer_written_test
```
KatBrandt, line 12: We don't need this here since this isn't actually state we want to keep track of. This is a behavior and you have a method by the same name that executes that behavior. Additionally, we wouldn't need this in the attr_readers

lib/facility.rb
```diff
@@ -1,14 +1,64 @@
...
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    return nil unless services.include?("Vehicle Registration")
```
KatBrandt, line 20: Great use of a guard clause!!!  The other way you could write this (some people don't like the unless) `return nil if !services.include?("Vehicle Registration")`. Either way is fine.

lib/facility.rb
```diff
@@ -1,14 +1,64 @@
...
+    vehicle.register
+    vehicle.registration_date == Date.today
+    @collected_fees += fee_amount(vehicle)
```
KatBrandt, line 23: Great use of a helper method! I will peek in the spec file to see if there is a test for the helper method. :)

lib/facility.rb
```diff
@@ -1,14 +1,64 @@
...
+      25
+    elsif vehicle.electric_vehicle?
+      200
+    else
+      100
+    end
+  end
+  
+  def administer_written_test(registrant)
+    return false unless services.include?("Written Test")
```
KatBrandt, line 38: I would add a space after this guard clause so it's really clear that it's one line of code and if it's not activated, then the rest of the block will run.

lib/facility.rb
```diff
@@ -1,14 +1,64 @@
...
+    end
+  end
+
+  def administer_road_test(registrant)
+    return false unless services.include?("Road Test")
+    if registrant.age >= 16 && registrant.license_data[:written]==true
+      registrant.license_data[:license] = true
+    else 
+      false
+    end
```
KatBrandt, line 52: Actually... do you need the guard clause for some of these? Technically you're checking for three different conditions to be met in order to run line 49, so you could remove the guard clause here and make line 48 read: `if services.include?("Road Test") && registrant.age >= 16 && registrant.license_data[:written]`. Then as long as all three of those evaluate to true, line 49 would run.

lib/facility_factory.rb
```diff
@@ -0,0 +1,39 @@
+class FacilityFactory
+
+  def create_or_facility(facility)
+    facility = facility.slice(:title,:location_1, :phone_number)
+    facility[:name] = facility.delete(:title)
```
KatBrandt, line 5: This is really smart! I like that you're adding in the key you need. You could also just do `facility[:name] = facility[:title]`, no need to actually delete the key, we just need the value.

lib/facility_factory.rb
```diff
@@ -0,0 +1,39 @@
+class FacilityFactory
+
+  def create_or_facility(facility)
+    facility = facility.slice(:title,:location_1, :phone_number)
```
KatBrandt, line 4: We don't need to slice this up either. We can just adjust the keys we need, then just send the whole big hash into Facility.new since our initialize method in Facility is already set up to look for only the keys it needs.

lib/facility_factory.rb
```diff
@@ -0,0 +1,39 @@
...
+    facility[:name] = facility.delete(:title)
+    facility[:address] = parse_or_address(facility.delete(:location_1))
+    facility[:phone] = facility.delete(:phone_number)
+    Facility.new(facility)
+  end
+
+  def parse_or_address(raw_address)
```
KatBrandt, line 11: Great helper method.

lib/facility_factory.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  def parse_ny_address(raw_address)
+     "#{raw_address[:street_address_line_1]}, #{raw_address[:city]}, #{raw_address[:state]} #{raw_address[:zip_code]}"
+  end
+
+  def create_mo_facility(facility)
+    facility[:address] = parse_mo_address(facility)
+    facility = facility.slice(:name, :address, :phone)
+    Facility.new(facility)
```
KatBrandt, line 33: I'm curious where these Facility objects are being stored. I would have expected that all the facility data from the api would be filtered through this method and maybe there is an attribute of `@mo_facilities = []` in the intialize of this class to store all the instances in? I'll keep looking around to see if I see this.

lib/registrant.rb
```diff
@@ -0,0 +1,18 @@
+class Registrant
+  attr_reader :name, :age, :license_data
+
+  def initialize(name, age, permit = false, license_data = false)
```
KatBrandt, line 4: We don't need the license_data in the arguments here, we can just set that within the intialize like you have

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+require './lib/dmv_data_service'
+require './lib/facility_factory'
+
+RSpec.describe FacilityFactory do
+
+  let(:oregon) do 
+    {:title=>"Albany DMV Office", :zip_code=>"97321", :website=>"http://www.oregon.gov/ODOT/DMV/pages/offices/albany.aspx", :type=>"DMV Location", :phone_number=>"541-967-2014", :agency=>"Transportation, Department of ", :location_1=>{:latitude=>"44.632897", :longitude=>"-123.077928", :human_address=>"{\"address\": \"2242 Santiam Hwy SE\", \"city\": \"Albany\", \"state\": \"OR\", \"zip\": \"97321\"}"}}
+  end
+  
+  let(:new_york) do
```
KatBrandt, line 13: I like that you wrote this to mimic the API data so you can test how your methods are working. Nice job.

spec/facility_spec.rb
```diff
@@ -22,5 +24,133 @@
...
+      
+      facility_1.register_vehicle(camaro)
+      expect(camaro.registration_date).to eq(Date.today)
+      expect(camaro.plate_type).to eq(:antique)
+      
+      facility_1.register_vehicle(bolt)
+      expect(bolt.registration_date).to eq(Date.today)
+      expect(bolt.plate_type).to eq(:ev)
+      
+
```
KatBrandt, line 60: Since you're testing collected_fees here, I would maybe do an assertion after each vehicle registration here to show that the value of collected_fees is changing.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  it "has atttibutes" do
+    registrant = Registrant.new('Bruce', 18, true )
+    expect(registrant.name).to eq("Bruce")
+    expect(registrant.age).to eq(18)
+    expect(registrant.permit?).to eq(true)
+  end
+
+  it "can be updated" do
```
KatBrandt, line 16: I'd try to be more specific with these tests. This one could be `it "can earn permit"` and then you just test that the earn_permit method changes the value of permit?

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,16 @@
...
+    vehicle = vehicle.slice(:vin_1_10, :model, :make, :model_year).merge({engine: :ev})
+    vehicle[:year] = vehicle.delete(:model_year)
+    vehicle[:vin] = vehicle.delete(:vin_1_10)
+    Vehicle.new(vehicle)
+  end
+
+  def create_vehicles(registrations)
+    registrations.map do |registration|
+      self.create_wa_vehicle(registration)
+    end
```
KatBrandt, line 14: This is good, nice use of the map. Again though, we could be storing this array in an instance variable so everytime we want these vehicles, we don't have to do all this mapping everytime. Also, we don't need the `self` keyword on line 13

lib/facility.rb
KatBrandt: I did not see a test. Remember to test all methods that are in your class, even helper methods.

PR #166, opened 2023-04-07
Jesse Thomas 

jgthomas-12:
This should be the main main.


lib/vehicle.rb
```diff
@@ -7,12 +7,17 @@ class Vehicle
               :model,
               :engine
 
+  attr_accessor :registration_date,
+                :plate_type
```
KatBrandt, line 11: Do we need plate_type here to be an attr_accessor? You have the method of `set_plate_type_dawggy` (were you tired by this point? :) ) which handles setting the plate type for you.

lib/vehicle.rb
```diff
@@ -22,4 +27,16 @@ def antique?
   def electric_vehicle?
     @engine == :ev
   end
+
+
+  def set_plate_type_dawggy
+    date = Time.new.strftime("%Y")
+    return @plate_type = :antique if (date.to_i - @year) >= 25
+    return @plate_type = :ev if @engine == :ev
+    return @plate_type = :regular
```
KatBrandt, line 36: You're guard clause happy now! lol. One guard clause, if you need more, rewrite it into an if block. It's much mor readable and allows people to see the flow of data better.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,14 @@
...
+  def create_vehicles(database)
+    database.map do |data|
+      vehicle = Vehicle.new({
+        vin: data[:vin_1_10], 
+        year: data[:model_year], 
+        make: data[:make], 
+        model: data[:model], 
+        engine: :ev
+      })
+    end
```
KatBrandt, line 12: This works! Again though, would want an attribute to store these vehicle instances in so we don't have to map over everytime we want to access this data.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,56 @@
+require 'spec_helper'
+
+RSpec.describe Registrant do 
+  before(:each) do
+    @registrant_1 = Registrant.new('Bruce', 18, true )
+    @registrant_2 = Registrant.new('Penny', 15 )
+  end
+  
+  describe "Iteration 1" do
```
KatBrandt, line 9: Moving forward, let's not use "Iteration 1" or the like for testing since that's not how we'll see it in the real world. :) Can put in here the method or feature you're testing though.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    it "exists" do
+
```
KatBrandt, line 12: This space should be removed when you only have one assertion in the test like this and no setup.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      expect(@registrant_2.permit?).to be(false)
+    end
+    
+    it "has an earn_permit method" do
+
+      expect(@registrant_2.permit?).to be(false)
+      @registrant_2.earn_permit
+
+      expect(@registrant_2.permit?).to eq(true)
```
KatBrandt, line 53: Good test to show the default value, the change, and what the new value is.

KatBrandt (commented):
I know you mentioned you were not proud of this project, but I think you did a good job! It's good to reflect on what didn't go well though and use that as a place to ensure you're learning what you want from the Battleship project. Your Github commits and branching workflow looks good!
Consider getting a slot with student support to talk through this project and review any logic or concepts that feel fuzzy still.

PR #167, opened 2023-04-10
Javen Wilson 

javenb022:
DMV project


lib/facility_factory.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def create_facility(api)
+    api.map do |facility|
+      Facility.new({
+        name: format_name(facility),
+        address: format_address(facility),
+        phone: format_phone_number(facility) 
+      })
+    end
```
KatBrandt, line 12: Great job using this map here! I do like that you used helper methods. A step further, we could have put all these new facility objects into an instance variable of `@facilities =[]` in the initialize method, that way we only have to do this mapping once. 

I read your message that you didn't test the helper methods, but we definitely would want to unit test them, to make sure that the input and output are consistent and are what we expect. Testing methods also allows us to test for sad path and edge cases to ensure the code doesn't break if an extra comma gets in or something.

lib/facility_factory.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def format_name(facility)
+    if facility[:state] && facility[:state] == "MO"
+      facility[:name].rstrip
+    elsif facility[:state] && facility[:state] == "NY"
+      facility[:office_name]
+    else facility[:website] && facility[:website].include?('oregon')
```
KatBrandt, line 20: Typically we don't put a condition on the else. If we have another condition, it should be an `elsif` block. Then we _would_ want a default value in the else incase none of the above conditions are met.

lib/facility_factory.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  end
+
+  def format_address(facility)
+    if facility[:state] && facility[:state] == "MO"
+      "#{facility[:address1]}, #{facility[:city]}, #{facility[:state]}, #{facility[:zipcode]}"
+    elsif facility[:state] && facility[:state] == "NY"
+      "#{facility[:street_address_line_1]}, #{facility[:city]}, #{facility[:state]} #{facility[:zip_code]}"
+    else facility[:website] && facility[:website].include?('oregon')
+      address = JSON.parse(facility[:location_1][:human_address], symbolize_names: true)
```
KatBrandt, line 32: Good job using JSON.parse to get the address out.

lib/facility_factory.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+  end
+  
+  def format_phone_number(facility)
+    if facility[:state] && facility[:state] == "MO"
+      facility[:phone]
+    elsif facility[:state] && facility[:state] == "NY"
+      facility[:public_phone_number]
+    else facility[:website] && facility[:website].include?('oregon')
+      facility[:phone_number]
```
KatBrandt, line 43: Really makes you wish there was more consistency huh? lol

lib/registrant.rb
```diff
@@ -0,0 +1,24 @@
+class Registrant
+  attr_reader :name, :age, :permit, :license_data
```
KatBrandt, line 2: Careful here. We only want to expose methods that we need. We don't actually need the `permit` method that you exposed here. we have the `permit?` method already.

lib/registrant.rb
```diff
@@ -0,0 +1,24 @@
...
+  def earn_permit
+    @permit = true
+  end
+
+  def permit?
+    if @permit == false
+      false
+    else
+      true
+    end
```
KatBrandt, line 22: This works, but what if all that was inside this method was `@permit`? Then if @permit was true, it would just return true. If @permit was false, it would return false. So no need for the if block in this one.

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+
+    it 'creates new facilities from OR' do
+      facility = FacilityFactory.new
+      or_dmv_office_locations = DmvDataService.new.or_dmv_office_locations
+      or_facilities = facility.create_facility(or_dmv_office_locations)
+      
+      expect(or_facilities[0].name).to be_a String
+      expect(or_facilities[1].address).to be_a String
+      expect(or_facilities[2].phone).to be_a String
```
KatBrandt, line 18: You can also check that the objects are Facility objects. `expect(or_facilities.first).to be_a Facility`. But yeah, that's really all we can check here since we don't want to check the specific data since we don't control the data.

spec/facility_spec.rb
```diff
@@ -2,25 +2,193 @@
...
   before(:each) do
-    @facility = Facility.new({name: 'Albany DMV Office', address: '2242 Santiam Hwy SE Albany OR 97321', phone: '541-967-2014' })
+    @facility_1 = Facility.new({name: 'Albany DMV Office', address: '2242 Santiam Hwy SE Albany OR 97321', phone: '541-967-2014' })
+    @facility_2 = Facility.new({name: 'Ashland DMV Office', address: '600 Tolman Creek Rd Ashland OR 97520', phone: '541-776-6092' })
+    @cruz = Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice} )
+    @bolt = Vehicle.new({vin: '987654321abcdefgh', year: 2019, make: 'Chevrolet', model: 'Bolt', engine: :ev} )
+    @camaro = Vehicle.new({vin: '1a2b3c4d5e6f', year: 1969, make: 'Chevrolet', model: 'Camaro', engine: :ice} )
+    @registrant_1 = Registrant.new('Bruce', 18, true )
+    @registrant_2 = Registrant.new('Penny', 16 )
+    @registrant_3 = Registrant.new('Tucker', 15 )
```
KatBrandt, line 12: You mentioned using the before(:each) block more in your submission survey. Just a heads up that it really is a personal preference. Some people like using the before block, and some don't, just depends. :) So I'd never mark you wrong for not using it.

spec/facility_spec.rb
```diff
@@ -2,25 +2,193 @@
...
-      @facility.add_service('Vehicle Registration')
-      expect(@facility.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
+      expect(@facility_1.services).to eq([])
+      @facility_1.add_service('New Drivers License')
+      @facility_1.add_service('Renew Drivers License')
+      @facility_1.add_service('Vehicle Registration')
+      expect(@facility_1.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
+      expect(@cruz.registration_date).to eq(nil)
+      expect(@facility_1.registered_vehicles).to eq([])
+      expect(@facility_1.collected_fees).to eq(0)
```
KatBrandt, line 33: We don't need lines 31, 32, and 33 in this test because this test is just testing that we can add services to a facility, which you did wonderfully lines 26-30.

spec/facility_spec.rb
```diff
@@ -2,25 +2,193 @@
...
-      expect(@facility.services).to eq([])
+      expect(@facility_1).to be_an_instance_of(Facility)
+      expect(@facility_1.name).to eq('Albany DMV Office')
+      expect(@facility_1.address).to eq('2242 Santiam Hwy SE Albany OR 97321')
+      expect(@facility_1.phone).to eq('541-967-2014')
+      expect(@facility_1.services).to eq([])
     end
   end
 
   describe '#add service' do
```
KatBrandt, line 24: Make sure your describe blocks encapsulate it blocks that test the method in question. So the register method should be outside of this describe block.

spec/facility_spec.rb
```diff
@@ -2,25 +2,193 @@
...
+    expect(@facility_1.collected_fees).to eq(100)
+
+    expect(@facility_1.register_vehicle(@camaro)).to eq([@cruz, @camaro])
+    expect(@camaro.registration_date).to eq(Date.today)
+    expect(@camaro.plate_type).to eq(:antique)
+    expect(@facility_1.registered_vehicles).to eq([@cruz, @camaro])
+    expect(@facility_1.collected_fees).to eq(125)
+  end
+
+  it 'registers bolt car' do
```
KatBrandt, line 69: Your tests themselves are good, I would just like to see these it blocks encompassing particular behaviors of methods, not being focused on specific instances. So this it block could be "it registers :regular vehicle", but you could also combine these three tests. (for the three cars) since they are so similar and have the it block be `it "registers vehicles"`

spec/registrant_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+      expect(registrant_1.license_data).to eq({:written => false, :license => false, :renewed => false})
+
+      expect(registrant_2.name).to eq('Penny')
+      expect(registrant_2.age).to eq(15)
+      expect(registrant_2.permit?).to eq(false)
+      expect(registrant_2.license_data).to eq({:written => false, :license => false, :renewed => false})
+
+      registrant_2.earn_permit
+
+      expect(registrant_2.permit?).to eq(true)
```
KatBrandt, line 29: Remember to test each method in it's own it block. So we'd want the earn_permit and permit? methods to have their own test block. 

Great job testing the initial value of permit and then the changed value.

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    it 'creates new vehicles from wa' do
+      factory = VehicleFactory.new
+      wa_ev_registrations = DmvDataService.new.wa_ev_registrations
+
+      new_cars = factory.create_vehicles(wa_ev_registrations)
+      expect(new_cars[0].make).to eq("SMART")
+      expect(new_cars[1].make).to eq("SMART")
+      expect(new_cars[2].make).to eq("VOLKSWAGEN")
+      expect(new_cars[3].make).to eq("VOLKSWAGEN")
+      expect(new_cars[10].make).to eq("CHEVROLET")
```
KatBrandt, line 22: This becomes a "brittle" test because it's testing data that we have no control over. If the dmv api service changes their api, all these tests will fail. Typically we just want to test the code we wrote, so like in the FacilityFactory spec, you tested the data types these methods return, which is really all we can control. :)

lib/facility_factory.rb
```diff
@@ -0,0 +1,46 @@
...
+      Facility.new({
+        name: format_name(facility),
+        address: format_address(facility),
+        phone: format_phone_number(facility) 
+      })
+    end
+  end
+
+  def format_name(facility)
+    if facility[:state] && facility[:state] == "MO"
```
KatBrandt, line 16: Hmm, I don't think you need to check both of these things actually. If the :state key didn't exist, then `facility[:state] == "MO"` would just return false since `nil == "MO"` would be false. So you just need `if facility[:state] == "MO"`

PR #194, opened 2023-05-23
Allen Russell

garussell:
Through iteration 3


lib/dmv.rb
```diff
@@ -9,8 +10,59 @@ def add_facility(facility)
...
+
+  def missouri_parse(missouri)
+    missouri.flat_map do |office|
+      name = office[:name]
+      address = office[:address1]
+      phone = office[:phone]
+
+      {name: name, address: address, phone: phone}
+    end
+  end
```
megstang, line 47: I like how you created separate methods for these states reports.

lib/dmv.rb
```diff
@@ -9,8 +10,59 @@ def add_facility(facility)
...
+
+  def parse_states_data(dmv_office)
+    if dmv_office.first.has_key?(:title)
+      locations = oregon_parse(dmv_office)
+    elsif dmv_office.first.has_key?(:office_name)
+      locations = new_york_parse(dmv_office)
+    elsif dmv_office.first.has_key?(:name)
+      locations = missouri_parse(dmv_office)
+    end
```
megstang, line 56: I believe there's a state key that comes in with the API response. That might be a better key to use to help determine how it should be formatted. It's possible that we start taking in data from Pennsylvania. If Pennsylvania wasn't formatted exactly like Missouri's response was formatted, but it did have a name key, then we get ourselves into some trouble.

lib/facility.rb
```diff
@@ -1,14 +1,62 @@
...
+      vehicle.assign_plate(:ev)
+      @collected_fees += 200
+    else
+      vehicle.assign_plate(:regular)
+      @collected_fees += 100
+    end
+    
+    @registered_vehicles << vehicle
+    vehicle.time_stamp
+  end
```
megstang, line 39: I like how you have this set up to return nil before even going into the condition.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,20 @@
...
+      model = vehicle[:model]
+      vin = vehicle[:vin_1_10]
+      year = vehicle[:model_year]
+      make = vehicle[:make]
+
+      new_vehicle = Vehicle.new({vin: vin, year: year, make: make, model: model, engine: :ev})
+      @new_vehicles << new_vehicle
+    end
+    @new_vehicles
+  end
```
megstang, line 19: If you used a `map` rather than `select`, you wouldn't need to use a placeholder to accumulate the vehicles that you create. In fact, the way you have it set up right now, `each` would be a better option than `select`, because you're iterating through a collection, doing something, and then returning the array you used to accumulate. If you use `map`, and change each vehicle data into a vehicle object, then you have your array already for you! 

```ruby
     registrants.map do |vehicle|
        Vehicle.new({
          engine: :ev,
          make: vehicle[:make],
          model: vehicle[:model],
          vin: vehicle[:vin_1_10],
          year: vehicle[:model_year]
        })
    end 
 end
```

PR #197, opened 2023-05-23
Paul Bennett

pcbennett108:
My the_dmv project files.


lib/facility_locations.rb
```diff
@@ -0,0 +1,134 @@
...
+      or_facilities(facilities)
+    elsif location == "mo"
+      mo_facilities(facilities)
+    elsif location == "ny"
+      ny_facilities(facilities)
+    end
+  end
+  
+  # The zipcode_locator determines which state API to use.
+  def zipcode_locator(zipcode)
```
KatBrandt, line 46: What a great helper method! I wonder though, if we could just combine the zipcode_locator into the method above though, could just do `if zipcode.between?(97001, 97920)` then `or_facilities(facilities)`. Does that make sense? I don't think we're using the 'or', 'mo', and 'ny' for anything else, so could jsut combine the two steps. Really like your thoguht process on breaking up this logic!

lib/facility_locations.rb
```diff
@@ -0,0 +1,134 @@
...
+    address.push(facility[:street_address_line_1])
+    if facility.include?(:street_address_line_2)
+      address.push(facility[:street_address_line_2])
+    end
+    address.push(facility[:city])
+    address.push(facility[:state])
+    address.push(facility[:zip_code])
+    address.join(" ")
+  end
+
```
KatBrandt, line 133: Yeah, this is great. Unfortunately, there is not great standardization across APIs and how everyone formats their data and their keys, so doing methods like this to account for those variations is necessary. Great job with all these helper methods!

spec/facility_locations_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+    it 'can initialize' do
+      expect(@locations).to be_an_instance_of(FacilityLocations)
+    end
+  end
+  
+  describe '#generate_location_data' do
+    it 'can fetch or_dmv_office_locations' do
+      expect(@or_dmv_office_locations).to be_an(Array)
+      expect(@or_dmv_office_locations.size).to eq(59)
+      expect(@or_dmv_office_locations[0][:title]).to eq("Albany DMV Office")
```
KatBrandt, line 21: Or even just expect this to be_a String, since we don't really care if their office name is Albany DMV Office or if they change it to Albany Count DMV Office.

spec/facility_locations_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+      expect(or_facilities[3].name).to eq("Baker City DMV Office")
+      expect(or_facilities[3].address).to eq("3370 10th St Suite A Baker City OR 97814")
+      expect(or_facilities[3].phone).to eq("541-523-4355")
+      expect(or_facilities[3].services).to eq([])
+      expect(or_facilities[3].registered_vehicles).to eq([])
+      expect(or_facilities[3].collected_fees).to eq(0)
+    end
+  end
+
+  describe '#generate_second_location_data' do
```
KatBrandt, line 52: You can be more specific with what this describe block is doing, it could be `describe "ny office locations" do` since you are very specifically testing ny, not just any random second location. 
I do really like that you've separated your tests into separate describe blocks.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,31 @@
...
+    make_inventory = {}
+    vehicles.each do |vehicle|
+      if make_inventory[vehicle[:make]] == nil
+        make_inventory[vehicle[:make]] = 1
+      else
+        make_inventory[vehicle[:make]] += 1
+      end
+    end
+    make_inventory
+  end
```
KatBrandt, line 29: Fun! :)

spec/facility_spec.rb
```diff
@@ -3,24 +3,178 @@
...
+      expect(@facility_2.services).to eq([])
+      expect(@facility_2.register_vehicle(@bolt)).to eq(nil)
+      expect(@facility_2.registered_vehicles).to eq([])
+      expect(@facility_2.collected_fees).to eq(0)
+    end
+
+  end
+
+  describe '#written Test' do
+    it 'will only provide service if added' do
```
KatBrandt, line 93: This to me suggests that this test will also add the service and then test that now we _can_ perform that service. Maybe change the name of the test to `it can't provide service if not added to services` or show that after these two assertions, that when we add the service, it does work.

spec/facility_spec.rb
```diff
@@ -3,24 +3,178 @@
...
+      expect(@facility_1.services).to eq([])
+      expect(@facility_1.administer_written_test(@registrant_1)).to eq(false)
+    end
+    
+    it 'can check registrant status' do
+      expect(@registrant_1.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+      expect(@registrant_1.permit?).to eq(true)
```
KatBrandt, line 100: I'm not sure what this it block is testing. These tests are already in the registrant_spec so they're not necessary here.

PR #201, opened 2023-05-23
Chris C the_dmv

topher-nullset:
This is the actual main branch pull request, sorry for the first one.


lib/registrant.rb
```diff
@@ -0,0 +1,27 @@
+class Registrant
+
+  attr_reader :name, :age, :permit
```
KatBrandt, line 3: We don't need/want the `:permit` in the attr_reader since we don't want to or need to expose that method. Or `permit?` method acts as a reader for the @permit attribute.

lib/registrant.rb
```diff
@@ -0,0 +1,27 @@
...
+      license: false,
+      renewed: false
+    }
+  end
+
+  def permit?
+    @permit
+  end
+
+  def license_data
```
KatBrandt, line 19: Put this in the attr_reader

spec/registrant_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    expect(registrant_2.permit?).to be false
+  end
+
+  it 'can give permit to registrant' do
+    registrant_2 = Registrant.new('Penny', 15)
+
+    expect(registrant_2.permit?).to be false
+    expect(registrant_2.license_data).to eq({ written: false, license: false, renewed: false })
+    registrant_2.earn_permit
+    expect(registrant_2.permit?).to be true
```
KatBrandt, line 22: Good job testing the earn_permit method.

lib/vehicle.rb
```diff
@@ -6,13 +6,18 @@ class Vehicle
               :make,
               :model,
               :engine
+  
+  attr_accessor :registration_date,
```
KatBrandt, line 10: With an attr_accessor, remember you are creating two methods, a reader (a getter) and a setter, so we want to make sure we test both of those in the vehicle_spec. Can we read the registration_date AND can we reset it?

KatBrandt (commented):
To answer your question on the survey, you did not go down a random fork. You are on the right track and your code is looking good. Let me know if you have any questions regarding my feedback.

PR #202, opened 2023-05-23
Connor R

ConnorRichmond:
Project Submission


lib/vehicle.rb
```diff
@@ -1,25 +1,48 @@
...
+  def register
+    @registration_date = Date.today
+    if self.electric_vehicle?
+      @plate_type = :ev
+    elsif self.antique?
+      @plate_type = :antique
+    else
+      @plate_type = :regular
+    end
+  end
```
megstang, line 45: Your use of `self` here is totally fine, but also isn't necessary. When you leave off the `self`, it's still going to call to an instance method.

lib/facility.rb
```diff
@@ -1,14 +1,64 @@
...
+    if vehicle.antique?
+      @collected_fees += 25
+    elsif vehicle.electric_vehicle?
+      @collected_fees += 200
+    else
+      @collected_fees += 100
+    end
+    @registered_vehicles << vehicle
+   end
+  end
```
megstang, line 38: Nice use of your conditionals here to check if the service is available first, and then collect fees accordingly.

lib/registrant.rb
```diff
@@ -0,0 +1,53 @@
...
+  def initialize(name, age, permit = false)
+    @name = name
+    @age = age
+    @permit = permit
+    @license_data = {
+      :written => false,
+      :license => false,
+      :renewed => false
+    }
+  end
```
megstang, line 20: Nice use of a default parameter here.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,27 @@
...
+          newformat[:make] = registration[:make]
+          newformat[:model] = registration[:model]
+          newformat[:engine] = :ev
+          formatted << newformat
+      end
+      formatted
+  end
+
+  #end class
+end
```
megstang, line 27: This looks really good, and again, I like that you're getting the practice with creating your own helper methods. We could maybe slim this down a little bit. If we did do one method, we could do something like: 

```ruby
     registrants.map do |vehicle|
        Vehicle.new({
          engine: :ev,
          make: vehicle[:make],
          model: vehicle[:model],
          vin: vehicle[:vin_1_10],
          year: vehicle[:model_year]
        })
 end
```

spec/facility_locations_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+  #did check for other locations in the files to show that it isnt pulling from already set vars
+  it 'creates data form new york api' do
+    ny_dmv_office_locations = DmvDataService.new.ny_dmv_office_locations
+    ny_dmv = locations.create_facility(ny_dmv_office_locations)
+    expect(ny_dmv).to be_a(Array)
+    expect(ny_dmv[1]).to be_a(Facility)
+    expect(ny_dmv[1].name).to eq('ROCHESTER DOWNTOWN')
+    expect(ny_dmv[1].address).to eq('200 E. MAIN STREET STE. 101 ROCHESTER NY 14604')
+    expect(ny_dmv[1].phone).to eq('5857531604')
+  end
```
megstang, line 53: We'll talk more about class methods in a later class, but with the way you have this `create_facility` set up, it might be a good time to use a class method. I choose to use a class method when I want functionality in a class, but don't care much about state. Take a look into class methods if you'd like!

lib/facility_locations.rb
```diff
@@ -0,0 +1,39 @@
...
+        (JSON.parse location[:location_1][:human_address]).values.join(" ")
+    elsif location[:state] == "NY"
+        [location[:street_address_line_1], location[:street_address_line_2], location[:city], location[:state], location[:zip_code]].compact.join(" ")
+    elsif location[:state] == "MO"
+        [location[:address1], location[:city], location[:state], location[:zipcode]].join(" ")
+    end
+  end
+
+  #end class
+end
```
megstang, line 39: This is all so great! You're already creating some helper methods to help format your return value. We could make other helper methods to help us format values, kind of like what you did on line 19. For example, we could make a method to `format_ny_address`. We'd probably need to pass in `location`, but we could have the logic on line 32 in that method, and then on line 32, we'd just scall `format_ny_address(location)`.

PR #204, opened 2023-05-23
Clay O'Leary - This is my final submission 

Captainlearyo:
with love


lib/dmv.rb
```diff
@@ -8,9 +9,52 @@ def add_facility(facility)
...
+    arr =[]
+    service.each do |i|  
+    @facilities.each do |facility|
+      if facility.services.include?(i)
+        arr << facility
+      end
+     end
     end
+    arr
   end
```
megstang, line 22: This will work! I think you've got a lot of extra code that you also don't need though. We might be able to get this working with a `find_all`. I'm thinking something like: 

```ruby
    @facilities.find_all do |facility|
    end
  end
```

lib/dmv.rb
```diff
@@ -8,9 +9,52 @@ def add_facility(facility)
...
+            facility_deatails[:phone] = i[:phone_number]
+
+            new_facility_arr << facility_deatails
+        end
+    end
+    new_facilities_hash = new_facility_arr.map{ |i| Facility.new(i) }
+    new_facilities_hash
+
+   
+  end
```
megstang, line 58: The logic in here looks good. Ruby convention is that methods should not be longer than 7 lines. Once we start to get longer methods, we want to start considering how we can create other helper methods to pull out some logic from another method. This also helps align with single responsibility principle (SRP). Currently, this method is responsible for formatting NY, MO, and OR facilities. One option is we could create a method to format each state. That way, we pull out some responsibility into other methods.

lib/facility.rb
```diff
@@ -1,14 +1,59 @@
...
+      @collected_fees += 25
+      vehicle.plate_type = :antique
+    elsif vehicle.electric_vehicle?
+      @collected_fees += 200
+      vehicle.plate_type = :ev
+    else
+    @collected_fees += 100
+    vehicle.plate_type = :regular
+    end
+  end
```
megstang, line 30: When a vehicle is registered, we also want to set a registration date. At the moment, in your vehicle class, your registration date is set upon initialization, but never updated once the vehicle is officially registered.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,22 @@
...
+        vehicles_deatails[:year] = i[:model_year]
+        vehicles_deatails[:make] = i[:make]
+        vehicles_deatails[:model] = i[:model]
+        vehicles_deatails[:engine] = i[:electric_vehicle_type]
+        vehicles_deatails_arr << vehicles_deatails
+      end
+      new_vehicles_arr = vehicles_deatails_arr.map{ |i| Vehicle.new(i) }
+
+      new_vehicles_arr
+    end
```
megstang, line 20: We could probably slim this up quite a bit. Maybe we do something like: 

```ruby
     registrants.map do |vehicle|
        Vehicle.new({
          engine: :ev,
          make: vehicle[:make],
          model: vehicle[:model],
          vin: vehicle[:vin_1_10],
          year: vehicle[:model_year]
        })
    end 
 end
```

PR #205, opened 2023-05-23
Derek Chavez

DChavez18:
PR for project submission


KatBrandt (commented):
Overall you're doing a great job thinking about testing and writing your methods. Ruby convention looks good overall. I did leave feedback on areas you can improve. In general for testing, think about what we want to test. Are we testing the return value and all the side effects that this method is causing?

PR #206, opened 2023-05-23
Dylan Timmons

DylanScotty:


lib/facility.rb
```diff
@@ -1,14 +1,54 @@
...
+    @collected_fees = 0
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    return nil if !services.include?('Vehicle Registration')
```
juliet-e, line 23: Nice use of the guard clause! When we use these, it's good practice to put a space afterward to make it really clear that there's an edge case there

lib/facility.rb
```diff
@@ -1,14 +1,54 @@
...
+    @registered_vehicles << vehicle
+    if vehicle.plate_type == :antique
+      @collected_fees += 25
+    elsif vehicle.plate_type == :ev
+      @collected_fees += 200
+    else
+      @collected_fees += 100
+    end
+    @collected_fees
```
juliet-e, line 32: There's a minor detail here of updating the vehicle registration date when you register it. It can be a one line thing:
`vehicle.registration_date = Date.today`

lib/dmv.rb
```diff
@@ -9,8 +9,9 @@ def add_facility(facility)
   end
 
   def facilities_offering_service(service)
-    @facilities.find do |facility|
+    @facilities.find_all do |facility|
```
juliet-e, line 12: Nice bug fixes!

lib/facility.rb
```diff
@@ -1,14 +1,54 @@
...
+
+  def administer_road_test(registrant)
+    return false unless services.include?('Road Test') && registrant.license_data[:written]
+    registrant.license_data[:license] = true
+    true
+  end
+
+  def renew_drivers_license(registrant)
+    return false unless services.include?('Renew License')
+    return false unless registrant.license_data[:written] && registrant.license_data[:license]
```
juliet-e, line 49: Breaking all these cases up into different lines was a good call for readability!

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,22 @@
+require './lib/vehicle'
+
+class VehicleFactory
+    def create_vehicles(registrations)
+      registrations.map do |registration|
```
juliet-e, line 5: Nice use of .map

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,55 @@
+require 'spec_helper'
+require './lib/vehicle'
+require './lib/vehicle_factory'
+require './lib/dmv_data_service'
+require './lib/facility_factory'
+
+RSpec.describe FacilityFactory do
+
+    let(:oregon) do
```
juliet-e, line 9: This is actually a GREAT practice of using test data in this test rather than calling the DmvDataFactory. Good call! This way you know exactly what data is being passed into your tests.

juliet-e: In mod3, when you are handling lots of data like this, we will show you some other ways to store data in a test but for now this is totally fine. I wouldn't say this is too long of a test file or there is too much going on

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+                :human_address=>"{\"address\": \"2242 Santiam Hwy SE\", \"city\": \"Albany\", \"state\": \"OR\", \"zip\": \"97321\"}"}
+            }
+    end
+
+    it "exists" do
+        facility_factory = FacilityFactory.new
+        expect(facility_factory).to be_an_instance_of(FacilityFactory)
+    end
+
+    it "can make a oregon office" do
```
juliet-e, line 28: For this step, we would've liked to see you create a method that can parse ALL the Oregon output (i.e. an array of hashes that you turn into an array of facilities)

spec/facility_spec.rb
```diff
@@ -23,4 +26,150 @@
...
+    it 'registers a vehicle and sets the registration date' do
+      facility_1 = Facility.new({ name: 'Albany DMV Office', address: '2242 Santiam Hwy SE Albany OR 97321', phone: '541-967-2014' })
+      cruz = Vehicle.new({ vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice })
+
+      facility_1.add_service('Vehicle Registration')
+      facility_1.register_vehicle(cruz)
+
+      cruz.get_registered
+
+      expect(cruz.registration_date).to eq(Date.today)
```
juliet-e, line 61: I see now that you covered this functionality but that you have to call it on the instance of the Vehicle. That makes sense. It would be totally fine to have this registration date update in your facility method but I think I understand where you were coming from (creating methods to manipulate its own attributes maybe). It is sometimes okay to have classes manipulate other object's data when it is a *side effect* of what their method is doing. For example, updating the registration date of a vehicle is a *side effect* of registering a car at a facility.Happy to chat more about this if you'd like

spec/facility_spec.rb
```diff
@@ -23,4 +26,150 @@
       expect(@facility.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
     end
   end
+
+  describe 'Vehicle Registration' do
```
juliet-e, line 30: Nice robust test cases!

spec/facility_spec.rb
```diff
@@ -23,4 +26,150 @@
...
+      expect(registrant_2.permit?).to eq(false)
+      expect(facility_1.administer_written_test(registrant_2)).to eq(false)
+
+      registrant_2.earn_permit
+
+      expect(facility_1.administer_written_test(registrant_2)).to eq(true)
+      expect(registrant_2.license_data).to eq({:written=>true, :license=>false, :renewed=>false})
+    end
+
+    it 'administer a written test to registrant_3' do
```
juliet-e, line 131: it'd be great to name these test cases a bit more specifically based on what they're doing. For example `it 'should not administer a written test to someone who is under 16`

PR #207, opened 2023-05-23
Daniel Gallagher - DMV

Daniel-Gallagher92:


lib/facility.rb
```diff
@@ -1,14 +1,62 @@
...
+      fee = 100
+      vehicle.plate_type = :regular
+    end
+    @registered_vehicles.push(vehicle)
+    @collected_fees += fee
+    vehicle
+  end
+
+  def administer_written_test(registrant) 
+    if registrant.permit? == true && registrant.age >= 16
```
dcoleman21, line 39: I think you might also need to take into consideration here whether or not `@services` includes `Written Test`.

lib/facility.rb
```diff
@@ -1,14 +1,62 @@
...
+      registrant.license_data[:written] = true
+    else
+      false
+    end
+  end
+
+  def administer_road_test(registrant)
```
dcoleman21, line 46: See my comment above, `@services` would need to include `Road Test`.

lib/facility.rb
```diff
@@ -1,14 +1,62 @@
...
+    if registrant.license_data[:written] == true 
+      true
+      registrant.license_data[:license] = true 
+    else
+      false
+    end
+  end
+
+  def renew_drivers_license(registrant) 
+    if registrant.license_data[:license] == true
```
dcoleman21, line 56: See my comment above, `@services` would need to include `Renew License`.

lib/registrant.rb
```diff
@@ -0,0 +1,18 @@
...
+    @permit = permit
+    @license_data = {:written=>false, :license=>false, :renewed=>false}
+  end
+
+  def permit?
+    @permit
+  end
+
+  def earn_permit 
+    @permit = true
```
dcoleman21, line 15: You could also add what is called a guard clause here that would check
```ruby
if @age >= 15
```
That we they could only `earn_permit` if they were of the correct age to do so.

spec/vehicle_spec.rb
```diff
@@ -14,7 +14,7 @@
       expect(@cruz.make).to eq('Chevrolet')
       expect(@cruz.model).to eq('Cruz')
       expect(@cruz.engine).to eq(:ice)
-      expect(@cruz.registration_date).to eq(nil)
+      expect(@cruz.registration_date).to eq(Date.today)
```
dcoleman21, line 17: One test that you are missing in this is for the `registration_date` method that you have.

dcoleman21 (commented):
Overall nice job on this project. See my comments and let me know if you have any questions.

PR #208, opened 2023-05-23
Matthew Lim

MatthewTLim:
Made it through iteration 3 but didn't start iteration 4. Parsers are cool.


lib/facility.rb
```diff
@@ -1,14 +1,64 @@
 class Facility
-  attr_reader :name, :address, :phone, :services
+  attr_reader :name, :address, :phone, :services, :registered_vehicles, :collected_fees
 
-  def initialize(name, address, phone)
-    @name = name
-    @address = address
-    @phone = phone
+  def initialize(info = {})
```
dcoleman21, line 4: You don't need to set info with a hash default value. The attributes come in as a hash when`.new` is called on a facility.
```ruby
    @facility = Facility.new({name: 'Albany DMV Office', address: '2242 Santiam Hwy SE Albany OR 97321', phone: '541-967-2014' })
```

lib/facility.rb
```diff
@@ -1,14 +1,64 @@
...
+    @collected_fees = 0
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    if self.services.include?("Vehicle Registration") && vehicle.antique? == true
```
dcoleman21, line 18: I wouldn't call `self.services` use the instance variable from your `initialize`.  Also you will only need to check that `@services.include?("Vehicle Registration")` once and then you can proceed to check the `plate_type` for each scenario. An example:
```ruby
def register_vehicle(vehicle)
  return nil if !services.include?("Vehicle Registration")
  if vehicle.antique? 
    @collected_fees += 25
    vehicle.plate_type = :antique
    @collected_fees += 200
    vehicle.plate_type = :ev
  else 
    @collected_fees += 100
    vehicle.plate_type = :regular 
  end
    @collected_fees
end
```
Something like that.

lib/facility.rb
```diff
@@ -1,14 +1,64 @@
...
+      @registered_vehicles
+    else
+      nil
+    end
+  end
+
+  def administer_written_test(registrant)
+    if registrant.permit? == true && registrant.age >= 16 && self.services.include?("Written Test")
+      registrant.license_data[:written] = true
+      true
```
dcoleman21, line 41: I think you can get ride of line 41. Also you don't need to call `self.services` use the instance variable `@services.include?("Written Test")`

lib/facility.rb
```diff
@@ -1,14 +1,64 @@
...
+    else
+      false
+    end
+  end
+
```
dcoleman21, line 46: See my comment in your `administer_written_test` method for the next two methods.

lib/facility_factory.rb
```diff
@@ -0,0 +1,47 @@
+class Facility_factory
+
+  def create_facility(facilities_data)
+    facilities = []
+    facilities_data.map do |facility_data|
+      if facility_data.has_key?(:office_name)
+        data = ny_parser(facility_data)
+        # require 'pry'; binding.pry
+      elsif facility_data.has_key?(:address1)
```
dcoleman21, line 9: Nice job creating helpers. Make sure to get rid of your commented out `prys`

lib/registrant.rb
```diff
@@ -0,0 +1,22 @@
...
+      renewed: false
+    }
+  end
+
+  def permit?
+    @permit
+  end
+
+  def earn_permit
+    @permit = true
```
dcoleman21, line 20: Nice job here. You could add a guard clause like
```ruby
@permit = true if @age >= 15
```
To ensure that only eligible registrants can get permits.

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    # expect(@factory.create_facility(@ny_dmv_office_locations)).to eq([])
+    end
+
+    it 'can create MO facilities using the dmv data' do
+      expect(@factory.create_facility(@mo_dmv_office_locations)).to be_an(Array)
+      expect(@factory.create_facility(@mo_dmv_office_locations)[0]).to be_a(Facility)
+      # expect(@factory.create_facility(@mo_dmv_office_locations)).to eq([])
+    end
+  end
+
```
dcoleman21, line 33: Nice job testing your helpers!

spec/facility_spec.rb
```diff
@@ -23,4 +23,180 @@
...
+      @bolt = Vehicle.new({vin: '987654321abcdefgh', year: 2019, make: 'Chevrolet', model: 'Bolt', engine: :ev} )
+      @camaro = Vehicle.new({vin: '1a2b3c4d5e6f', year: 1969, make: 'Chevrolet', model: 'Camaro', engine: :ice} )
+      @facility_1.add_service('Vehicle Registration')
+    end
+
+    it "can register a vehicle at the appropriate facility" do
+      expect(@cruz.registration_date).to eq (nil)
+      expect(@facility_1.registered_vehicles).to eq([])
+      expect(@facility_1.collected_fees).to eq(0)
+      expect(@facility_1.register_vehicle(@cruz)).to eq([@cruz])
```
dcoleman21, line 41: I would probably take this out of an expect statement and make it a command query like
```ruby
@facility_1.register_vehicle(@cruz)
```
Then your expect on line 42 is testing that your `register_vehicle(vehicle)` method is doing what you want.

spec/facility_spec.rb
```diff
@@ -23,4 +23,180 @@
...
+
+    it "charges the correct registration fee for the vehicle type" do
+      @facility_1.register_vehicle(@cruz)
+      @facility_1.register_vehicle(@camaro)
+      @facility_1.register_vehicle(@bolt)
+
+      expect(@facility_1.collected_fees).to eq(325)
+    end
+
+    it "adds the date the vehicle was registered to vehicle info" do
```
dcoleman21, line 53: Did this test not work?

spec/facility_spec.rb
```diff
@@ -23,4 +23,180 @@
...
+
+      expect(@facility_1.renew_drivers_license(@registrant_3)).to eq(false)
+      expect(@registrant_3.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+    end
+
+    it "can renew a drivers license if the registrant meets the requiremnts and the facility offers the appropriate services" do
+      @facility_1.add_service('Renew License')
+      expect(@facility_1.renew_drivers_license(@registrant_2)).to eq(true)
+      expect(@registrant_2.license_data).to eq({:written=>true, :license=>true, :renewed=>true})
+    end
```
dcoleman21, line 200: Overall I think that these tests look good.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      registrant_1 = Registrant.new('Bruce', 18, true )
+
+      expect(registrant_1.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+    end
+  end
+
+  describe "earn_permit" do
+    it "can earn a permit" do
+      registrant_2 = Registrant.new('Penny', 15 )
+
```
dcoleman21, line 43: I might add an additional expect statement here
```ruby
```
Then your expect statement on line 45 is accurately testing that `earn_permit` is doing what you expect.

dcoleman21 (commented):
Matt, overall I think you did well. There are a few places were you can DRY up your code a little bit. Your tests are looking tons better. I left comments for you to review. Also, moving forward, we really need to see you implement a stronger git workflow. You only had 15 commits and 0 PRs. Let me know if you have any questions. Keep up the good work.

PR #209, opened 2023-05-23
andi lovetto

andilovetto:


lib/dmv.rb
```diff
@@ -9,7 +9,7 @@ def add_facility(facility)
   end
 
   def facilities_offering_service(service)
-    @facilities.find do |facility|
+    @facilities.find_all do |facility|
```
dcoleman21, line 12: Nice debugging here.

lib/facility.rb
```diff
@@ -1,14 +1,76 @@
...
+    @name = facility_hash[:name]
+    @address = facility_hash[:address]
+    @phone = facility_hash[:phone]
     @services = []
+    @registered_vehicles = []
+    @collected_fees = 0
   end
 
-  def add_services(service)
+  def add_service(service)
```
dcoleman21, line 14: Those pesky `ssss`. Good job.

lib/facility.rb
```diff
@@ -1,14 +1,76 @@
...
+      end
+      vehicle.registration_date = Date.today.year
+      @registered_vehicles.push(vehicle)
+    else
+      nil
+    end
+  end
+
+  def administer_written_test(registrant)
+    if @services.include?("Written Test")
```
dcoleman21, line 38: You could probably combine your if statement on line 38 and 39 like
```ruby
if @services.include?("Written Test") && registrant.permit && registrant.age >= 16
  registrant.license_date[:written] = true
else
  false
end 
```

lib/facility.rb
```diff
@@ -1,14 +1,76 @@
...
+        true
+        registrant.license_data[:renewed] = true
+      else
+        false
+      end
+    else
+      false
+    end
+  end
 end
```
dcoleman21, line 76: There is some repetitive code in the above 3 methods
```ruby
registrant.permit && registrant.age >= 16
```
Perhaps you could make a helper method that can be used in each of those methods instead? :)
Nice job.

lib/facility_factory.rb
```diff
@@ -0,0 +1,51 @@
+class FacilityFactory
```
dcoleman21, line 1: I think this is a really great start. One thing to consider is how can we make one `create_facilities(state_data)` method that can be dynamic to read in any state that we pass it in our test?
I like the helpers that you made to format numbers and addresses, can these also be dynamic??

lib/registrant.rb
```diff
@@ -0,0 +1,18 @@
...
+    @permit = permit
+    @license_data = {:written=>false, :license=>false, :renewed=>false}
+  end
+
+  def permit?
+    @permit
+  end
+
+  def earn_permit
+    @permit = true
```
dcoleman21, line 16: This is great. We could also add in a guard clause like
```ruby
@permit = true if @age >= 15
```
Which would really ensure that a registrant could not get a permit unless eligible.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,14 @@
+class VehicleFactory 
+
+  def create_vehicles(vehicles)
```
dcoleman21, line 3: This looks good. Another way you could simplify this would be to move line 11 up to line 5 like so and 
```ruby
Vehicle.new({
  engine: :ev,
  make: vehicle[:make],
  model: vehicle[:model],
  vin: vehicle[:vin_1_10],
  year: vehicle[:model_year
})
```

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    }
+    expected = "3164 Telegraph Road, St Louis, Mo, 63125"
+    expect(factory.formatted_mo_address(address_hash)).to eq(expected)
+  end
+
+  it 'reformats MO phone numbers' do
+    factory = FacilityFactory.new
+    expect(factory.formatted_mo_phone("(314) 887-1050")).to eq("314-887-1050")
+  end
+end
```
dcoleman21, line 68: I love that you tested all of your helpers. Like I mentioned in your `facility factory` class. Could you just have one method called `create_facilities` and then pass it the data that you want?
```ruby
it '#create_facilities for Oregon' do 
   factory = FacilityFactory.new
   oregon_facilities = DmvDataService.new.or_dmv_office_locations
   expect(factory.create_facilities(oregon_faciliites).to be_an(Array)
end 
```
This would then follow the same format for `MO` and `NY`.

spec/facility_spec.rb
```diff
@@ -1,6 +1,15 @@
...
 RSpec.describe Facility do
+  let(:facility_1) { Facility.new({name: 'Albany DMV Office', address: '2242 Santiam Hwy SE Albany OR 97321', phone: '541-967-2014' }) }
+  let(:facility_2) { Facility.new({name: 'Ashland DMV Office', address: '600 Tolman Creek Rd Ashland OR 97520', phone: '541-776-6092' }) }
+  let(:cruz) { Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice} ) }
+  let(:bolt) { Vehicle.new({vin: '987654321abcdefgh', year: 2019, make: 'Chevrolet', model: 'Bolt', engine: :ev} ) }
+  let(:camaro) { Vehicle.new({vin: '1a2b3c4d5e6f', year: 1969, make: 'Chevrolet', model: 'Camaro', engine: :ice} ) }
+  let(:registrant_1) { Registrant.new('Bruce', 18, true) }
+  let(:registrant_2) { Registrant.new('Penny', 16) }
+  let(:registrant_3) { Registrant.new('Tucker', 15) }
+
```
dcoleman21, line 12: I would probably use either `let` statements or a `before(:each)`

spec/facility_spec.rb
```diff
@@ -23,4 +32,76 @@
       expect(@facility.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
     end
   end
+
+  #facility_1 & facility_2
+  it 'can register vehicles' do
+    expect(facility_1.add_service('Vehicle Registration')).to eq(["Vehicle Registration"])
```
dcoleman21, line 38: This could probably just be a command query
```ruby
facility_1.add_service('Vehicle Registration')
```
then have an expect statement like
```ruby
expect(facility_1.services).to eq(['Vehicle Registration'])
```

spec/facility_spec.rb
```diff
@@ -23,4 +32,76 @@
...
+    expect(cruz.registration_date).to eq(nil)
+    expect(facility_1.registered_vehicles).to eq([])
+    expect(facility_1.collected_fees).to eq(0)
+    facility_1.register_vehicle(cruz)
+    expect(cruz.registration_date).to eq(Date.today.year)
+    expect(cruz.plate_type).to eq(:regular)
+    expect(facility_1.registered_vehicles).to eq([cruz])
+    expect(facility_1.collected_fees).to eq(100)
+    facility_1.register_vehicle(camaro)
```
dcoleman21, line 47: Is there a way that you could break this big test into smaller tests? Like a test for `collected fees`, `adding services`, and `registering a vehicle`?? Great job being robust though.

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'spec_helper'
+
+RSpec.describe VehicleFactory do
+  it 'creates vehicle objects' do
```
dcoleman21, line 4: You could follow this format for your `facility_factory_spec`.

dcoleman21 (commented):
Overall, really nice job. I think that your testing is strong as far as making sure that you are testing everything, especially helpers. Maybe see if you could work on breaking them up to be more specific to the particular methods. If you want more explanation we can chat. I left some comments on ways that you could DRY up your code a little more. Great job. If you have any questions on the comments that I left, let me know.

PR #212, opened 2023-05-23
Dan Sauer

danieljsauer:
Got through iteration 3 and into 4. Made a method that went through the vehicle API data and removed all duplicate vehicles based on vin number in order to get the total number of cars.

I have developed an unfair bias against the state of Oregon because of this project.  I lost hours of my life to human_address. 

Also, my git got pretty messy... but I did learn a lot about it.


lib/dmv.rb
```diff
@@ -1,16 +1,73 @@
...
+      website = website_location[1]
+      formatted_facility[:website] = website
+      
+    else
+      formatted_facility[key] = value
+    end
+  end
+
+  formatted_facility
+end
```
megstang, line 73: This was definitely a time consuming and frustrating method to write, I'm sure. THere's a lot going on.. it might be nice to think about making some more helper methods to pull out some of this logic into other methods. That way, you'd better align with SRP (single responsibility principle)

lib/facility.rb
```diff
@@ -1,14 +1,88 @@
...
   end
 
-  def add_services(service)
+  def add_service(service)
+    valid_entries = ["Road Test", "Written Test", "Renew License", "Vehicle Registration"]
+    unless service.is_a?(String) && valid_entries.include?(service)
+      return "Invalid entry. Services offered are: #{valid_entries.join(", ")}"
+    end
     @services << service
   end
```
megstang, line 34: Great additions here, especially making sure that it's a valid entry.

lib/facility.rb
```diff
@@ -1,14 +1,88 @@
...
+      car.registration_date = Date.today.strftime("%Y-%m-%d")
+      collect_cash(25)
+      @registered_vehicles << car 
+    else
+      car.plate_type = :regular
+      car.registration_date = Date.today.strftime("%Y-%m-%d")
+      collect_cash(100) 
+      @registered_vehicles << car 
+    end
+  end
```
megstang, line 55: I like that you are returning something if a certain condition is met first, because then your code doesnt need to go through the entire conditional tree for an unnecessary reason.

Rather than setting the plate_type and the registration_date in this method, you could consider putting a method in the Vehicle class that will do those things for you. That would again, align more with SRP, and it would abstract that logic so that the Vehicle class is the only one that knows about it, and this class doesn't need to worry about it.

lib/registrant.rb
```diff
@@ -0,0 +1,19 @@
...
+  end 
+
+  def permit?
+    @permit
+  end 
+
+  def earn_permit
+    @permit = true 
+  end 
+end
```
megstang, line 19: 👍

lib/facility.rb
```diff
@@ -1,14 +1,88 @@
...
+    if person.license_data[:drivers_license] == true 
+      person.license_data[:renewed] = true 
+    else 
+      "#{person.name} does not qualify for license renewal."
+    end 
+  end 
+
+  def collect_cash(cash) 
+    @collected_fees += cash
+  end
```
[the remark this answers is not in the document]
megstang: And also, make sure you're testing ANY method that you write, even if it's a helper method.

megstang (commented):
@danieljsauer Nice work! I left some comments for you to take a look over. Please let me know if you have any questions.

PR #216, opened 2023-05-23
Ian Lyell DMV Pull request

ILyell:


lib/dmv.rb
```diff
@@ -1,16 +1,60 @@
...
+        address: format_address(location),
+        phone: format_phone(location)
+      }
+      @facilities.push(Facility.new(facility_hold))
+    end
   end
 
   def facilities_offering_service(service)
-    @facilities.find do |facility|
+    @facilities.find_all do |facility|
```
dcoleman21, line 20: Nice job debugging this to use `find_all`.

lib/dmv.rb
```diff
@@ -1,16 +1,60 @@
 class Dmv
+  attr_reader :facilities
 
   def initialize
     @facilities = []
   end
 
-  def add_facility(facility)
-    @facilities << facility
+  def add_facilities(dmv_locations)
```
dcoleman21, line 8: This method looks good.

lib/dmv.rb
```diff
@@ -1,16 +1,60 @@
...
+
+  def format_phone(facility_hash)
+    if (facility_hash[:phone] && facility_hash[:address]) || (facility_hash[:phone_number] && facility_hash[:title])
+      facility_hash[:phone] || facility_hash[:phone_number]
+    elsif facility_hash[:public_phone_number]
+      "#{facility_hash[:public_phone_number][0..2]}-#{facility_hash[:public_phone_number][3..5]}-#{facility_hash[:public_phone_number][6..9]}"
+    elsif facility_hash[:phone]
+      facility_hash[:phone].tr("()", "").tr(" ","-")
+    end
+  end
```
dcoleman21, line 58: Nice job creating all of these helper method to get your data formatted. It would probably be better to create a whole new class called `facility_factory.rb` and have all of the methods that you created live there. This helps adhere to SRP (single responsibility principle) as well as Abstraction and Encapsulation. The DMV class should only be responsible for keeping track of facilities that offer specific services.

lib/facility.rb
```diff
@@ -1,14 +1,75 @@
...
+    @name = facility_hash[:name] 
+    @address =  facility_hash[:address] 
+    @phone = facility_hash[:phone]
     @services = []
+    @registered_vehicles = []
+    @collected_fees = 0
   end
 
-  def add_services(service)
+  def add_service(service)
```
dcoleman21, line 14: Good catch on the extra `s`.

lib/facility.rb
```diff
@@ -1,14 +1,75 @@
...
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    if @services.include?("Vehicle Registration") && !@registration_date
+      vehicle.add_registration
+      @registered_vehicles << vehicle
+      if vehicle.antique?
+        @collected_fees += 25
```
dcoleman21, line 23: Nice use of the helper methods you created in this method.

lib/facility.rb
```diff
@@ -1,14 +1,75 @@
...
+      else
+        @collected_fees += 100
+        vehicle.add_plate(:regular)
+      end
+    else
+      false
+    end
+  end
+
+  def administer_written_test(registrant)
```
dcoleman21, line 37: You could condense your `if` statements in this method. Something like
```ruby
if @service.include?("Written Test") && registrant.age >= 16 && registrant.permit?
  registrant.change_license_data(:written, true)
else
  false
end
```

lib/facility.rb
```diff
@@ -1,14 +1,75 @@
...
+      else
+        false
+      end
+    else
+      false
+    end
+  end
+  
+  def administer_road_test(registrant)
+    if @services.include?("Road Test")
```
dcoleman21, line 51: See my comment above for an idea to condense your `if` statements.

lib/facility.rb
```diff
@@ -1,14 +1,75 @@
...
+      else
+        false
+      end
+    else
+      false
+    end
+  end
+  
+  def renew_drivers_license(registrant)
+    if @services.include?('Renew License')
```
dcoleman21, line 64: See my comment in `administer_written_test` method to condense your `if` statements.

lib/facility.rb
```diff
@@ -1,14 +1,75 @@
...
+      if registrant.license_data[:license]
+        registrant.change_license_data(:renewed, true)
+        true
+      else
+        false
+      end
+    else
+      false
+    end
+  end
```
dcoleman21, line 74: You could also consider creating a helper method to help DRY up any repeated code in these methods.

lib/registrant.rb
```diff
@@ -0,0 +1,27 @@
...
+            renewed: false
+        }
+    end
+
+    def permit?
+        @permit
+    end
+    
+    def earn_permit
+        @permit = true
```
dcoleman21, line 21: This is great. You could also add a what is called a guard clause to ensure that a permit cannot be earned if a registrant is not eligible.
```ruby
@permit = true if @age >= 15
```

lib/vehicle.rb
```diff
@@ -5,21 +5,33 @@ class Vehicle
...
     @engine == :ev
   end
+
+  def add_registration
+    @registration_date = Date.today.year
+  end
+
+  def add_plate(plate_type)
+    @plate_type = plate_type
+  end
```
dcoleman21, line 36: Nice helper methods.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,14 @@
...
+    attr_reader :vehicles
+
+    def initialize
+        @vehicles = []
+    end
+
+    def create_vehicles(vehicle_list)
+
+        vehicle_list.each do |vehicle|
+            @vehicles << Vehicle.new(vehicle)
```
dcoleman21, line 11: Is this creating a Vehicle object with its attributes in a hash?

spec/dmv_spec.rb
```diff
@@ -2,42 +2,158 @@
...
+      expect(@dmv.format_phone(@ny_locations[1])).to eq("585-753-1604")
+      expect(@dmv.format_phone(@mo_locations[1])).to eq("573-358-3584")
+      
+    end
+  end
+  
+  describe '#format_hours' do
+    it 'takes hours from a facility hash' do
+      
     end
```
dcoleman21, line 157: Nice job testing all of the helper methods that you created. As I mentioned in the `dmv.rb` class I would create a `facility_factory_spec.rb` and move all of these tests in there. You would also create a `facility_factory.rb` and move all the methods you created for creating new facilities in there.

spec/facility_spec.rb
```diff
@@ -2,25 +2,199 @@
...
+      @facility_1.administer_written_test(registrant_1)
+      @facility_1.administer_road_test(registrant_1)
+      @facility_1.renew_drivers_license(registrant_1)
+      @facility_1.administer_written_test(registrant_2)
+      
+      expect(@facility_1.renew_drivers_license(registrant_1)).to eq(true)
+      expect(@facility_1.renew_drivers_license(registrant_2)).to eq(false)
+      expect(registrant_1.license_data).to eq({:written=>true, :license=>true, :renewed=>true})
+      expect(registrant_2.license_data).to eq({:written=>true, :license=>false, :renewed=>false})
     end
```
dcoleman21, line 198: I think that these tests look good. I noticed that you created helper methods of `add_plate` and `add_registration` are those tested in the `vehicle` class??

spec/registrant_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+            registrant_1 = Registrant.new('Bruce', 18, true)
+            registrant_2 = Registrant.new('Penny', 15)
+
+            expect(registrant_1.license_data).to eq({
+                written: false,
+                license: false,
+                renewed: false
+            })
+
+            registrant_1.change_license_data(:written, true)
```
dcoleman21, line 67: Looks like you are testing that your `change_license_data` method does indeed update the `license_data` method

spec/vehicle_spec.rb
```diff
@@ -33,4 +71,44 @@
...
+  end
+
+  describe '#add_plate' do
+    it 'can add a plate type' do
+
+      expect(@cruz.plate_type).to eq(nil)
+      expect(@bolt.plate_type).to eq(nil)
+      expect(@camaro.plate_type).to eq(nil)
+      
+      @cruz.add_plate(:regular)
```
dcoleman21, line 103: Nice job testing your `add_plate` method here. This would be considered a unit test.

dcoleman21 (commented):
Overall, nice job with this project. One this to keep in mind with testing is that you want to unit test and integration test. So for example, you created helper methods in your `vehicle` class for `add_registration` and `add_plate` and you indeed tested them in your `vehicle_spec` (unit test) but you also call on those method in your `facility` classes `register_vehicle` method. So they would also need to be tested in that spec as well (integration test). 
I left comments for you to review. Please let me know if you have any questions.

PR #217, opened 2023-05-23
Anna Wiley

awiley33:


lib/facility.rb
```diff
@@ -1,14 +1,61 @@
...
+        if vehicle.engine == :ev
+          @collected_fees += 200
+        elsif vehicle.plate_type == :antique
+          @collected_fees += 25
+        else
+          @collected_fees += 100
+        end
+      else
+    end
+  end
```
megstang, line 39: Nice! I really like that you created a `register` method in your `Vehicle` class to manage setting the registration date and also assigning a plate type. I think that you're demonstrating the OOP pillar of Abstraction by doing this -- keeping logic in classes that only that class needs to know about.

megstang: In your survey, you mentioned wanting some examples of how to DRY up your code. On lines 31, 33, and 35, you have very similar logic happening. You could, create a method, `collect_fee(amount)`, and have that logic in there. And then on the lines I listed, you could call to the method to perform the logic.

lib/facility.rb
```diff
@@ -1,14 +1,61 @@
...
+      registrant.take_road_test
+    else false
+    end
+  end
+
+  def renew_drivers_license(registrant)
+    if @services.include?('Renew License') && registrant.license_data[:license] == true 
+      registrant.have_license_renewed
+    else false
+    end
```
megstang, line 59: again... love that you are using helper methods in the respective classes to change values associated with that class. Great example of Abstraction again.

lib/facility.rb
```diff
@@ -1,14 +1,61 @@
...
+
+  def administer_written_test(registrant)
+    if @services.include?('Written Test') && registrant.permit? == true && registrant.age > 15
+      registrant.take_written_test
+    else false
+    end
+  end
+
+  def administer_road_test(registrant)
+    if @services.include?('Road Test') && registrant.license_data[:written] == true
```
megstang, line 49: on this line, and also on line 56 of this class you are useing `==true`, but that's not entirely necessary since `registrant.license_data[:written]` will evaluate to true if it is indeed true.

lib/registrant.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def take_road_test
+    license_data[:license] = true
+  end
+
+  def have_license_renewed
+    license_data[:renewed] = true
+  end
+end
```
megstang, line 34: 👍

lib/vehicle.rb
```diff
@@ -22,4 +29,17 @@ def antique?
...
+    @registration_date = Date.today
+    
+    if self.antique? == true
+      @plate_type = :antique
+    elsif @engine == :ev
+      @plate_type = :ev
+    else
+      @plate_type = :regular
+    end
+  end
```
megstang, line 44: 👍

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,12 @@
...
+  def create_vehicles(dmv_vehicles_data)
+    dmv_vehicles_data.map do |vehicle_data|
+      vehicle_data = vehicle_data.slice(:vin_1_10, :model_year, :make, :model)
+      vehicle_data[:vin] = vehicle_data.delete(:vin_1_10)
+      vehicle_data[:year] = vehicle_data.delete(:model_year)
+      vehicle_data[:engine] = :ev
+      Vehicle.new(vehicle_data)
+    end
+  end
+end
```
megstang, line 12: Super clean! I like it

megstang (commented):
@awiley33 Nice work! I left some comments for you to look over. Please let me know if you have ANY questions!

PR #218, opened 2023-05-23
Zanna Fitch

z-fitch:


lib/DMVfacility.rb
```diff
@@ -0,0 +1,30 @@
+class DMVfacility
```
KatBrandt, line 1: Remember to use PascalCase, where we capitalize every new word but keep them together, so `DMVFacility`

KatBrandt: And then the file name would be `dmv_facility.rb`

lib/DMVfacility.rb
```diff
@@ -0,0 +1,30 @@
...
+    
+    def create_facility(datas)
+        datas.map do |data|
```
KatBrandt, line 4: Good use of map

lib/DMVfacility.rb
```diff
@@ -0,0 +1,30 @@
...
+            Facility.new({
+                name: name(data),
+                address: address(data), 
+                phone: phone(data)})
+        end
+    end
+
+    def name(database)
```
KatBrandt, line 12: Good use of these helper methods to help you sort through the api data. I'm not sure `database` is the best name for a variable here, maybe `data` or `data` would be better.

spec/DMVfacility_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+            NY = facilities.create_facility(new_york_facilities)
+            expect(NY[0].name).to eq("JAMAICA KIOSK")
+            expect(NY[0].address).to eq("168-46 91ST AVE., 2ND FLR JAMAICA NY 11432")
+            expect(NY[1].phone).to eq("5857531604")
+
+            MO = facilities.create_facility(missouri_facilities)
+            expect(MO[0].name).to eq("OAKVILLE")
+            expect(MO[0].address).to eq("3164 TELEGRAPH ROAD ST LOUIS MO 63125")
+            expect(MO[0].phone).to eq("(314) 887-1050")
+        end
```
KatBrandt, line 25: Since you created those helper methods in your DMVFacility class, you'll want to be sure to test those helper methods in isolation - that's what we call a unit test. Given a specific in put, what is the output you expect? That's what we want to test. 
The tests you have here look good though.

lib/facility.rb
```diff
@@ -1,14 +1,71 @@
...
+    @collected_fees = 0
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(car)
+    if @services.include?("Vehicle Registration")
```
KatBrandt, line 18: Great job taking care of all these tasks in this method.

lib/facility.rb
```diff
@@ -1,14 +1,71 @@
...
+      car.registration_date = Date.today
+      @registered_vehicles.push(car)
+    else
+      nil
+    end
+  end
+
+  def administer_written_test(registrant)
+    if @services.include?("Written Test")
+        if registrant.age >= 16 && registrant.permit? == true
```
KatBrandt, line 38: We can reduce nesting of if/else blocks by chaining our boolean logic operators. So line 37 could be: `if @services.include?("Written Test") && registrant.age >= 16 && registrant.permit?` and then you can have your logic on line 39.

lib/facility.rb
```diff
@@ -1,14 +1,71 @@
...
+            registrant.license_data[:written] = true
+        else 
+          false
+        end
+    else
+        false
+    end
+  end
+
+  def administer_road_test(registrant)
```
KatBrandt, line 48: I don't see any tests for these three methods in the facility_spec file. Remember, every single method, we need to test and we want to test all possible return values for every method. These three methods have a lot of flow control so we'd want to test all the different combinations of conditions for these three methods.

lib/registrant.rb
```diff
@@ -0,0 +1,21 @@
+class Registrant
+    attr_reader :name, :age, :permit, :license_data
```
KatBrandt, line 2: We don't need/want `:permit` in the attr_reader since we have our method `permit?` which acts as our reader.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+
+            expect(@registrant_2.license_data).to eq({written: false, license: false, renewed: false})
+        end
+    end
+
+    describe '#earn_permit' do
+        it 'can add set peprmit to true' do
+            @registrant_2.earn_permit
+
+            expect(@registrant_2.permit?).to eq(true)
```
KatBrandt, line 34: Good job testing the side effects of earn_permit.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+        
+            facility_1 = Facility.new({name: 'Albany DMV Office', address: '2242 Santiam Hwy SE Albany OR 97321', phone: '541-967-2014' })
+            facility_2 = Facility.new({name: 'Ashland DMV Office', address: '600 Tolman Creek Rd Ashland OR 97520', phone: '541-776-6092' })
+        
+            ## WRITTEN TEST
+
+            expect(registrant_1.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+            expect(registrant_1.permit?).to eq(true)
+
+            expect(facility_1.administer_written_test(registrant_1)).to eq(false)
```
KatBrandt, line 52: oooo, I see now, you put the tests for these methods in the registrant_spec file. We will want these method tests to live in the facility spec file since that's the class these methods are defined in. Yes their side effects change the Registrant object, but we can check for those side effects in the facility spec file when we test these methods.

lib/vehicle.rb
```diff
@@ -6,13 +6,16 @@ class Vehicle
...
 
   def initialize(vehicle_details)
-    @vin = vehicle_details[:vin]
+    @vin = vehicle_details[:vin] 
     @year = vehicle_details[:year]
     @make = vehicle_details[:make]
     @model = vehicle_details[:model]
     @engine = vehicle_details[:engine]
+    @registration_date = registration_date
+    @plate_type = plate_type
```
KatBrandt, line 18: Technically you're setting @registration_date and @plate_type to nil here since you don't have a variable named registration_date or plate_type.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,17 @@
+class VehicleFactory
+    def initialize
+        
+    end
+
+    def create_vehicles(registrations)
+        registrations.map do |registration|
```
KatBrandt, line 7: Good use of map

PR #219, opened 2023-05-23
Mike Wood - 2305

MWoodshop:


lib/facility.rb
```diff
@@ -1,14 +1,188 @@
+# Helper module to title case strings.
+module Helper
```
KatBrandt, line 2: Yeah, I like that you used a module for these helper methods. We haven't learned this yet, but a module would get it's own file and then we'd use `include Helper` in the classes we want to "mixin" this code. We would test the Module by testing the methods in the spec file of the class we're using them in. So if you use these methods in the Facility class, you'd write tests for them in the facility_spec.rb.

lib/facility.rb
```diff
@@ -1,14 +1,188 @@
...
+    @address = data[:address]
+    @phone = data[:phone]
+    @website = data[:website]
     @services = []
+    @registered_vehicles = []
+    @collected_fees = 0
+  end
+
+  # Determines which state the facility is in and calls the appropriate method
+  def self.create_facility(state, facility_data)
```
KatBrandt, line 48: Yeah, I could see this as being a class method.

lib/facility.rb
```diff
@@ -1,14 +1,188 @@
...
+    else
+      # Returns an error if the state is not supported
+      raise "Unsupported state: #{state}"
+    end
+  end
+
+  # Creates a facility if the first argument in Facility.create_facility is 'OR'
+  def self.create_facility_from_oregon(facility_data)
+    # Map over the array of facility data and create a new Facility object for each facility
+    facility_data.map do |data|
```
KatBrandt, line 67: Yeah, I like that you're using a map here. The next thing you could do is once these are created, store them somewhere, maybe a class variable so that you only have to do this mapping once.

lib/facility.rb
```diff
@@ -1,14 +1,188 @@
...
+      name = Helper.title_case(data[:office_name])
+      street = Helper.title_case(data[:street_address_line_1])
+      city = Helper.title_case(data[:city])
+      state = data[:state]
+      zip_code = data[:zip_code]
+      phone = data[:public_phone_number]
+      # New York does not provide a website for each facility.
+      # Spoke with Kat who said this was fine. Could also be left as nil.
+      website = 'undefined'
```
KatBrandt, line 94: Yeah, and maybe one day they add a website, so could still do `website = data[:website]` because then if they don't have one, it would just be nil.

lib/facility.rb
```diff
@@ -1,14 +1,188 @@
...
+  def add_service(service)
     @services << service
   end
+
+  # Registers a vehicle if the facility offers the service
+  # and the vehicle is not already registered
+  def register_vehicle(vehicle)
+    raise 'Vehicle already registered.' if @registered_vehicles.include?(vehicle)
+
+    return nil unless @services.include?('Vehicle Registration')
```
KatBrandt, line 127: Yeah, nice use of guard clauses here.

lib/facility.rb
```diff
@@ -1,14 +1,188 @@
...
+    @registered_vehicles << vehicle
+    calculate_collected_fees(vehicle)
+    @registered_vehicles
+  end
+
+  # Calculates the fees collected for a vehicle registration
+  def calculate_collected_fees(vehicle)
+    plate_type_fees = {
+      ev: 200,
+      antique: 25
```
KatBrandt, line 140: Nice!!! I love a good hash lookup table!!!

lib/facility.rb
```diff
@@ -1,14 +1,188 @@
...
+    }
+
+    @collected_fees += plate_type_fees[vehicle.plate_type] || 100
```
KatBrandt, line 143: Nice

lib/facility.rb
```diff
@@ -1,14 +1,188 @@
...
+  # Renews a drivers license if the facility offers the service,
+  # if the registrant has taken the written test and road test,
+  # and if the registrant has not already renewed their license.
+  def renew_drivers_license(registrant)
+    raise 'Already renewed drivers license.' if registrant.renewed_drivers_license?
+
+    return nil unless registrant.written_test? && registrant.road_test? && services.include?('Renew License')
+
+    registrant.renew_drivers_license
+  end
```
KatBrandt, line 187: Great job overall with breaking up methods into smaller helper methods. It really helps with readability!

spec/facility_spec.rb
```diff
@@ -23,4 +37,489 @@
...
+    end
+
+    it 'registrants license data is correct' do
+      expect(@registrant_1.license_data).to eq({ written: false, license: false, renewed: false })
+      expect(@registrant_2.license_data).to eq({ written: false, license: false, renewed: false })
+      expect(@registrant_3.license_data).to eq({ written: false, license: false, renewed: false })
+    end
+
+    it 'registrants permit data is correct (tests permit? method)' do
+      expect(@registrant_1.permit?).to eq(true)
```
KatBrandt, line 186: This test and the one above it are not necessary since you've already tested these methods in the Registrant spec class, so we don't need to retest them here. We can assume they are working

spec/facility_spec.rb
```diff
@@ -23,4 +37,489 @@
...
+    it 'creates a single facility from New York (compare mapped data to API data)' do
+      facility_data = DmvDataService.new.ny_dmv_office_locations
+      new_york_facilities = Facility.create_facility('NY', facility_data)
+
+      expected_data = facility_data[0]
+      expect(new_york_facilities[0].name).to eq(Helper.title_case(expected_data[:office_name]))
+      expect(new_york_facilities[0].address).to eq("#{Helper.title_case(expected_data[:street_address_line_1])}, #{Helper.title_case(expected_data[:city])}, #{Helper.title_case(expected_data[:state])}, #{expected_data[:zip_code]}")
+      expect(new_york_facilities[0].phone).to eq(expected_data[:public_phone_number])
+      expect(new_york_facilities[0].website).to eq('undefined')
+    end
```
KatBrandt, line 444: Don't forget that white line space between an end and def. There is a lot in these tests so want to make it easier to read.

spec/facility_spec.rb
```diff
@@ -23,4 +37,489 @@
...
+
+    it 'return error on renew_drivers_license if registrant already renewed drivers license' do
+      @facility_1.add_service('Renew License')
+      @facility_1.renew_drivers_license(@registrant_1)
+      expect(@registrant_1.license_data).to eq({ written: true, license: true, renewed: true })
+      expect { @facility_1.renew_drivers_license(@registrant_1) }.to raise_error(StandardError, 'Already renewed drivers license.')
+    end
+  end
+
+  # Iteration 3
```
KatBrandt, line 349: Glad to see you have tested your class methods as well!

lib/registrant.rb
```diff
@@ -0,0 +1,53 @@
+class Registrant
+  attr_reader :name, :age, :permit, :license_data
```
KatBrandt, line 2: We don't want/need the `:permit` here in the attr_reader.

lib/registrant.rb
```diff
@@ -0,0 +1,53 @@
...
+    @permit = true
+  end
+
+  # Returns true if registrant has a permit, false otherwise
+  def permit?
+    @permit
+  end
+
+  # Changes written test status to true
+  def administer_written_test
```
KatBrandt, line 25: Nice, I like that you have these setter methods in this class.

KatBrandt: I don't see tests for all these methods though, make sure we test every helper method and test all possible return values of those methods.

lib/vehicle.rb
```diff
@@ -1,25 +1,50 @@
...
   end
 
+  # Determines if a vehicle is an electric vehicle
   def electric_vehicle?
     @engine == :ev
   end
+
+  # Determines the plate type of a vehicle
+  def determine_plate_type
+    @plate_type = if antique?
```
KatBrandt, line 42: It's a bit more readable if you place the assignment operator inside each block of the if/else statement. is line 43 would be `@plate_type = :antique` and then again on line 45 and 47.

spec/vehicle_spec.rb
```diff
@@ -1,36 +1,67 @@
 require 'spec_helper'
 
-RSpec.describe Vehicle do
+RSpec.describe Facility do
```
KatBrandt, line 3: Hmm... I'm confused because this file is the vehicle_spec.rb file, where are the vehicle tests?

KatBrandt (commented):
Overall great job! Very thorough and love that you're using things we haven't taught yet.

lib/vehicle.rb
MWoodshop: Makes sense, thanks.

PR #222, opened 2023-05-23
Parker Boeing - DMV Project

ParkerBoeing:
Thank you all for reviewing!


lib/facility.rb
```diff
@@ -1,14 +1,73 @@
...
+    @registered_vehicles = []
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    return p "Facility does not currently offer this service." unless @services.include?("Vehicle Registration")
```
KatBrandt, line 27: This is what we call a guard clause. Love it. I recommend putting a space after a guard clause so it's very clear, then the rest of the code below it should be at the same indentation as the guard clause as that's a one-liner and is not creating a block.

lib/facility.rb
```diff
@@ -1,14 +1,73 @@
...
+        @collected_fees += 100
+      end
+  end
+
+  def administer_written_test(registrant)
+    if @services.include?("Written Test") && registrant.age >= 16 && registrant.permit?
+      registrant.license_data[:written] = true
+    elsif @services.include?("Written Test") && registrant.age < 16
+      p "Sorry, the registrant is not old enough to take the written test."
+    elsif @services.include?("Written Test") && registrant.permit? == false
```
KatBrandt, line 47: Instead of doing `registrant.permit? == false`, we can do `!registrant.permit?` to shorten up the code.

lib/facility.rb
```diff
@@ -1,14 +1,73 @@
...
+      p "Registrant has not yet earned permit."
+    else
+      p "Facility does not currently offer this service."
+    end
+  end
+
+  def administer_road_test(registrant)
+    if @services.include?("Road Test") && registrant.license_data[:written] == true
```
KatBrandt, line 55: same thing here, whenever we're using a comparison operator and comparing something to a boolean, usually we don't need the comparison, just the first half of it, so this first line could be `if @services .include?("Road Test") && registrant.license_data[:written]`

lib/facility.rb
```diff
@@ -1,14 +1,73 @@
...
+  def renew_drivers_license(registrant)
+    if @services.include?("Renew License") && registrant.license_data[:license] == true
+      registrant.license_data[:renewed] = true
+    elsif @services.include?("Renew License") && registrant.license_data[:license] == false
+      p "Registrant must first obtain a license for it to be renewed."
+    else
+      p "Facility does not currently offer this service."
+    end
+  end
 end
```
KatBrandt, line 73: Great job with flow control and boolean logic.

spec/facility_spec.rb
```diff
@@ -23,4 +32,142 @@
...
+      expect(@facility_1.registered_vehicles).to eq [@cruz, @bolt, @camaro]
+      expect(@facility_1.collected_fees).to eq 325
+      expect(@facility_2.services).to eq []
+      expect(@facility_2.register_vehicle(@bolt)).to eq "Facility does not currently offer this service."
+      expect(@facility_2.registered_vehicles).to eq []
+      expect(@facility_2.collected_fees).to eq 0
+    end
+  end
+
+  describe '#administer_written_test' do
```
KatBrandt, line 91: Great use of describe blocks to organize your tests for specific methods.

lib/facility_factory.rb
```diff
@@ -0,0 +1,69 @@
+class FacilityFactory
+  attr_reader :operating_facilities
+
+  def initialize
+    @operating_facilities = []
```
KatBrandt, line 5: I like that you created an instance variable here to store the data in! That way we only have to run the method below once.

lib/facility_factory.rb
```diff
@@ -0,0 +1,69 @@
...
+  end
+
+  def define_facilities(facilities)
+    facilities.each do |facility_data|
+      if facility_data[:location_1].is_a?(Hash)
+        address_data = JSON.parse(facility_data[:location_1][:human_address], symbolize_names: true)
```
KatBrandt, line 11: This is where helper methods can be useful to shorten up this method to make it more clear what's going on. Could have each block within each flow (if/elsif) be it's own helper method.

lib/facility_factory.rb
```diff
@@ -0,0 +1,69 @@
...
+        })
+        @operating_facilities << facility
+      else
+        p "Invalid input, please use list of facilities from relevant states as argument."
+      end
+    end
+    @operating_facilities
+  end
+
+  def capitalize_words(string)
```
KatBrandt, line 42: Interesting formatting helper method. Nice job.

lib/facility_factory.rb
```diff
@@ -0,0 +1,69 @@
...
+    end
+    words.join(" ")
+  end
+
+  def capitalize_operating_hours(hours)
+    if hours
+    words = hours.split.map {|word| word.capitalize}
+    words.join(" ")
+    else
+      nil
```
KatBrandt, line 59: could use a guard clause for this. So the first line of this method would be `return nil if !hours` then all the other code could be outside the if stmt. Nice job with this helper method though!

lib/facility_factory.rb
```diff
@@ -0,0 +1,69 @@
...
+    end
+  end
+
+  def format_phone_number(phone)
+    return nil if phone.nil?
```
KatBrandt, line 64: yeah! Guard clause. Nice

lib/facility_factory.rb
```diff
@@ -0,0 +1,69 @@
...
+      if excluded_words.include?(word)
+        word
+      else
+        word.capitalize
+      end
+    end
+    words.join(" ")
+  end
+
+  def capitalize_operating_hours(hours)
```
KatBrandt, line 54: I don't see tests in the facility_factory_spec.rb file for all these helper methods. Remember, anytime you add a method to a class, we want a test. Yes, these are being tested indirectly when you call your `define_facilites` method, and you DO want to test that, which is called an integration test - making sure everything works together. But then we also want a Unit test for every single method tested in isolation to ensure when we put in a specific input, we get an expected output.

lib/registrant.rb
```diff
@@ -0,0 +1,23 @@
+class Registrant
+  attr_reader :name, :age, :permit, :license_data
```
KatBrandt, line 2: Don't need a `:permit` in the attr_reader since that is a method that's never used. We wrote the `permit?` method to act as the attr_reader.

lib/registrant.rb
```diff
@@ -0,0 +1,23 @@
...
+      renewed: false
+    }
+  end
+
+  def permit?
+    @permit
+  end
+
+  def earn_permit
+    return p "Registrant is not yet old enough to earn permit." if age < 16
```
KatBrandt, line 20: Nice guard clause

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,95 @@
...
+          year: vehicle_data[:model_year],
+          make: vehicle_data[:make],
+          model: vehicle_data[:model],
+          engine: :ev,
+          county: vehicle_data[:county]
+          })
+          @created_vehicles << vehicle
+      end
+      @created_vehicles
+    elsif vehicles[0][:record_type]
```
KatBrandt, line 22: Could potentially use some boolean logic to merge these two blocks together. The if and the elsif. Could do something like `if vehicles[0][:electric_vehicile_type] || vehicles[0][:record_type]` and then use that same logic when setting the keys `vin: vehicle_data[:vin_1_10] || vehicle_data[:vin]`

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,95 @@
...
+
+    max_count = 0
+    most_popular_make_model = "No vehicles."
+    make_model_counts.each do |make_model, count|
+      if count > max_count
+        max_count = count
+        most_popular_make_model = make_model
+      end
+    end
+    most_popular_make_model
```
KatBrandt, line 55: This method and logic works! Nice job with this. Could also play with the `max_by` enumerable to see if that can help reduce some code.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  def models_in_a_year(model, year)
+    count = 0
+    @created_vehicles.each do |vehicle|
```
KatBrandt, line 60: There's also a `count` enumerable that could probably help shorten this code too, but same logic! Great job.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,95 @@
...
+    max_count = 0
+    most_popular_county = "No vehicles listed."
+    county_counts.each do |county, count|
+      if count > max_count
+        max_count = count
+        most_popular_county = county
+      end
+    end
+    most_popular_county
+  end
```
KatBrandt, line 94: These methods look really great! Nice work. In the future, see if you can identify more helper methods to break out of these methods. A good way to first identify these is to look at the blocks of code you have, here one block is 79-83, the next block from 85-93, and see if you can break those into helper methods. Testing each method of course. :)

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    it 'can count the number of vehicles that are registered in a given year' do
+      wa_ev_registrations = DmvDataService.new.wa_ev_registrations
+      @factory.create_vehicles(wa_ev_registrations)
+      expect(@factory.vehicles_in_a_year("2015")).to eq 93
+      expect(@factory.vehicles_in_a_year("2016")).to eq 95
+    end
+  end
+
+  describe '#most_popular_county method' do
```
KatBrandt, line 56: Great job testing all these methods

PR #223, opened 2023-05-23
Gabe Torres

Gabe-Torres:


lib/dmv.rb
```diff
@@ -1,15 +1,17 @@
...
 
+  attr_reader :facilities
   def initialize
     @facilities = []
   end
 
+
   def add_facility(facility)
-    @facilities << facility
+    @facilities.push(facility)
```
juliet-e, line 10: You can use either the shovel ( << ) or .push here! They accomplish the same goal. The .push method can take multiple arguments but in this case (with 1 argument) there's no difference between what they can do

lib/dmv.rb
```diff
@@ -1,15 +1,17 @@
...
   end
 
   def facilities_offering_service(service)
-    @facilities.find do |facility|
+    @facilities.find_all do |facility|
```
juliet-e, line 14: nice use of find_all!

lib/facility.rb
```diff
@@ -1,14 +1,113 @@
...
+    @registered_vehicles = []
+    @collected_fees = 0
+  end
+
+  def register_vehicle(data)
+    if @services.include?('Vehicle Registration')
+    data.registration_date = Date.today
+    @registered_vehicles.push(data)
+      if data.antique? 
+        data.plate_type = :antique
```
juliet-e, line 25: Nice catch assigning plate_type!

lib/facility.rb
```diff
@@ -1,14 +1,113 @@
...
+    end
+  end
+
+  def parse_address(office_details)
+    if office_details[:address]
+        office_details[:address]
+    elsif office_details[:location_1]
+          address = Hash.new
+          location_1 = JSON.parse office_details[:location_1][:human_address]
+          address.store(:new_address, location_1)
```
juliet-e, line 87: nice use of `.store`!

lib/facility.rb
```diff
@@ -1,14 +1,113 @@
...
+    if office_details[:name]
+        office_details[:name]
+    elsif office_details[:title] 
+          office_details[:title] 
+    elsif office_details[:office_name]
+          office_details[:office_name]
+    end
+  end
+
+  def parse_address(office_details)
```
juliet-e, line 81: This method is a bit long. Consider extracting helper methods like `parse_location1` and `parse_location2` to improve readability

lib/facility.rb
```diff
@@ -1,14 +1,113 @@
...
+
+  def parse_phone(office_details)
+    if office_details[:phone]
+      office_details[:phone]
+    elsif office_details[:phone_number] 
+          office_details[:phone_number]
+    elsif office_details[:public_phone_number]
+          office_details[:public_phone_number]
+    end
+  end
```
juliet-e, line 112: We will talk a lot more about this theme throughout your journey at Turing, but when deciding where functionality should live, it's good to think about WHO or WHAT(class) should be responsible for what logic. As long as you can defend your design choices, lots of different things work. Parsing the facility data so that you can pass it right into the Facility objects that you make is a creative decision that was cool to see!

lib/office.rb
```diff
@@ -0,0 +1,8 @@
+class Office
+
+  def create_offices(data)
+    data.map do |office_data|
+      Facility.new(office_data)
```
juliet-e, line 5: Because you do all the parsing in the Facility class, this is a really clean method! 

Another option could be to have those parsers here. It might make more sense to call this class `Office_Factory` or something to indicate that its purpose to create facilities (it's not really a representation of an Office)

spec/facility_spec.rb
```diff
@@ -13,14 +22,171 @@
...
+    it 'can update registration dates on vehicles' do
+      @facility.add_service('Vehicle Registration')
+      @facility.register_vehicle(@cruz)
+      @facility.register_vehicle(@camaro)
+      @facility.register_vehicle(@bolt)
+      expect(@cruz.registration_date).to eq(Date.today)
+      expect(@camaro.registration_date).to eq(Date.today)
+      expect(@bolt.registration_date).to eq(Date.today)
+    end
+    it 'can assign plate type' do
```
juliet-e, line 48: Because these test cases require some test data setup, it's okay to test that they add the cars to registered_vehicles, update registration dates and assign plate_types in one `it` block, with the test case being `it can register vehicles`. Sometimes it takes a lot of work to set up each test case so if you're duplicating a lot, you can add more to a before each block or consolidate some assertions together into one it block (as long as they are related to the same functionality/theme)

spec/facility_spec.rb
```diff
@@ -13,14 +22,171 @@
...
+      @facility.administer_written_test(@registrant_2)
+      @facility.administer_road_test(@registrant_2)
+      @facility.renew_drivers_license(@registrant_2)
+      @facility.administer_written_test(@registrant_3)
+      expect(@facility.renew_drivers_license(@registrant_1)).to eq(true)
+      expect(@facility.renew_drivers_license(@registrant_2)).to eq(true)
+      expect(@facility.administer_written_test(@registrant_3)).to eq(false)
+      expect(@registrant_1.license_data).to eq({:written=>true, :license=>true, :renewed=>true})
+      expect(@registrant_2.license_data).to eq({:written=>true, :license=>true, :renewed=>true})
+      expect(@registrant_3.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
```
juliet-e, line 189: Excellent edge case testing!

spec/facility_spec.rb
```diff
@@ -13,14 +22,171 @@
...
     end
   end
-
+  describe '#registered vehicles' do
+    it 'can keep data on register vehicles' do 
+      expect(@facility.registered_vehicles).to eq([])
+      expect(@facility_2.registered_vehicles).to eq([])
+    end 
+  end
+  describe '#register vehicles' do
```
juliet-e, line 31: For readability it's best to add a blank space between test cases and methods

spec/tem.rb
```diff
@@ -0,0 +1,70 @@
+RSpec.describe Registrant do
+  before(:each) do
+      @registrant_1 = Registrant.new({name: 'Bruce', age: 18, permit: true} )
+      @registrant_2 = Registrant.new({name: 'Penny', age: 15} )
+    end
+    describe '#initialize' do
```
juliet-e, line 6: Again just remember spaces between it blocks, describes, etc. For readability

PR #224, opened 2023-05-23
Wil Fady

fadwil:
Learned a lot!


lib/facility.rb
```diff
@@ -1,14 +1,82 @@
...
+      end
+      @registered_vehicles << vehicle
+    else
+      nil
+    end
+  end
+
+  def administer_written_test(registrant)
+    if @services.include?('Written Test')
+      true
```
KatBrandt, line 45: Not sure why this 'true' is here. 
To reduce nesting, you can string together your if conditions, for example, your line 44 could read: `if @services.include?("Written Test") && registrant.permit? && registrant.age >= 16`

KatBrandt: Same feedback for the two following methods

lib/facility.rb
```diff
@@ -1,14 +1,82 @@
...
+      end
+    else
+      false
+    end
+  end
+
+  def renew_drivers_license(registrant)
+    if @services.include?('Renew License')
+      true
+      if registrant.license_data[:license] == true
```
KatBrandt, line 72: We technically don't need the comparison operator `==` since the statement `registrant.license_data[:license]` will itself return true or false.

spec/facility_spec.rb
```diff
@@ -2,25 +2,136 @@
...
+      expect(@facility_1.administer_road_test(@registrant_1)).to eq(true)
+      expect(@facility_1.renew_drivers_license(@registrant_1)).to eq(true)
+      expect(@registrant_1.license_data).to eq({:written=>true, :license=>true, :renewed=>true})
+      expect(@facility_1.renew_drivers_license(@registrant_3)).to eq(false)
+      expect(@registrant_3.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+      @registrant_2.earn_permit
+      expect(@facility_1.administer_written_test(@registrant_2)).to eq(true)
+      expect(@facility_1.administer_road_test(@registrant_2)).to eq(true)
+      expect(@facility_1.renew_drivers_license(@registrant_2)).to eq(true)
+      expect(@registrant_2.license_data).to eq({:written=>true, :license=>true, :renewed=>true})
```
KatBrandt, line 134: I do feel like these tests are robust enough. This class has a lot of different combinations and we have to test all of them to ensure we're testing every possible return value from a method. 
You could add more white spaces between 'chunks' of data within a test though to make it easier to read. Maybe group them by setup, then a white line, then the assertions, then a white line, then more setup, etc.

lib/facility_factory.rb
```diff
@@ -0,0 +1,52 @@
+class FacilityFactory
+  attr_reader :locations
+
+  def initialize
+    @locations = []
+  end
```
KatBrandt, line 6: I like that you created an instance variable to store all these locations in, makes it so you don't have to rerun these methods everytime you want to get the locations.

lib/facility_factory.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def facility_creator(dmv)
```
KatBrandt, line 8: I'm a bit confused on what this `dmv` argument is we're passing in here. If it's already the data from the service, then we probably don't want to do the comparison operator on lines 9, 11, and 13 because that will cause a lot of API calls and the data could get unwieldy. Maybe facility _creator just calls all there of your create methods: `create_oregon_facilities` and passes them their corresponding data: `DMV.new.or_dmv_office_locations` and then they just all run and add their locations to @locations.

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,77 @@
+require 'spec_helper'
+
+RSpec.describe FacilityFactory do
```
KatBrandt, line 3: This testing is fine for now. Testing 3rd party API data can be tricky so these assertions are good.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  describe '#initialize' do
+    it 'can initialize' do
+      expect(@registrant_1).to be_an_instance_of(Registrant)
+      expect(@registrant_1.name).to eq("Bruce")
+      expect(@registrant_1.age).to eq(18)
+      expect(@registrant_1.permit?).to eq(true)
+      expect(@registrant_1.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+      expect(@registrant_2.name).to eq("Penny")
+      expect(@registrant_2.age).to eq(15)
```
KatBrandt, line 17: Things like .age and .name, we really only need to test for one instance.

KatBrandt (commented):
Overall you've done a great job!

PR #225, opened 2023-05-23
Kaina Cockett

kcockett:


lib/facility.rb
```diff
@@ -1,14 +1,98 @@
...
+              :address, 
+              :phone, 
+              :services, 
+              :registered_vehicles
 
-  def initialize(name, address, phone)
-    @name = name
-    @address = address
-    @phone = phone
+  #def initialize(name, address, phone) #<original entry>
```
KatBrandt, line 10: Nice update

lib/facility.rb
```diff
@@ -1,14 +1,98 @@
...
+        vehicle.registration_fee = 25
+      elsif vehicle.engine == :ev
+        vehicle.plate_type = :ev
+        vehicle.registration_fee = 200
+      else
+        vehicle.plate_type = :regular
+        vehicle.registration_fee = 100
+      end
+    else
+      #p "Service not available for this facility"
```
KatBrandt, line 40: Or just return nil

lib/facility.rb
```diff
@@ -1,14 +1,98 @@
...
+    @collected_fees = 0
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    if @services.include?("Vehicle Registration")
```
KatBrandt, line 25: You can reduce nesting here by using a "guard clause". Then the first line of this method would be `return nil if !@services.include?("Vehicle Registration") and if it's not included, it will just return nil and not run anything else. Then the rest of your code would be un-nested.

lib/facility.rb
```diff
@@ -1,14 +1,98 @@
...
+      #p "Service not available for this facility"
+    end
+  end
+
+  def collected_fees #iterate through registered vehicles
+    fee_total = 0
+    @registered_vehicles.map do |vehicle|
+      fee_total += vehicle.registration_fee
+    end
+    fee_total
```
KatBrandt, line 49: This does theoretically work. I could forsee different facilities charging different fees though because of whre they are located, so this fee should be associated with the Facility rather than the Vehicle.

lib/facility.rb
```diff
@@ -1,14 +1,98 @@
...
+  end
+
+  def administer_written_test(registrant)
+    if registrant.permit?
+      if @services.include?("Written Test")
```
KatBrandt, line 54: You can reduce nesting by using our boolean logic operators. The first line of this could be `if registrant.permit? && @services.include?("Written Test")` and then you could un-nest the inner if stmt.

spec/facility_spec.rb
```diff
@@ -23,4 +31,79 @@
...
+      expect(@cruz.plate_type).to eq(:regular)
+      expect(@facility_1.collected_fees).to eq(100)
+    end
+
+    it 'register multiple vehicles' do
+      @facility_1.add_service('Vehicle Registration')
+      @facility_1.register_vehicle(@cruz)
+      @facility_1.register_vehicle(@camaro)
+      @facility_1.register_vehicle(@bolt)
+
```
KatBrandt, line 52: To make this test more robust we could also test that @facillity_1.registered_vehicles now holds all these vehicles in that array.

spec/facility_spec.rb
```diff
@@ -23,4 +31,79 @@
...
+    it '#renew_drivers_license requires service' do
+      expect(@facility_1.renew_drivers_license(@registrant_1)).to eq(false)
+      
+      @facility_1.add_service('Renew License')
+      expect(@facility_1.renew_drivers_license(@registrant_1)).to eq(false)
+      
+      @registrant_1.license_data[:license] = true
+      expect(@facility_1.renew_drivers_license(@registrant_1)).to eq(true)
+    end
+  end
```
KatBrandt, line 108: overall good testing in this class. There are so many possible combinations of logistics going on here and it's important to test them all.

lib/facility_factory.rb
```diff
@@ -0,0 +1,37 @@
+class FacilityFactory
+    def initialize
+        #
+    end
```
KatBrandt, line 4: Don't need to put an initialize method if it's empty. :)

lib/facility_factory.rb
```diff
@@ -0,0 +1,37 @@
...
+            else
+                p "Unknown name match: #{facility}"
+            end
+            output << Facility.new({
+                name: final_name, 
+                address: final_address, 
+                phone: final_phone, 
+                })
+        end
+        output
```
KatBrandt, line 35: This all works but this is quite the monster of a method! Every block within the if/else statement is a good candidate for being broken out into a helper method.

lib/registrant.rb
```diff
@@ -0,0 +1,24 @@
...
+        @permit = permit
+        @license_data = {:written=>false, :license=>false, :renewed=>false}
+    end
+
+    def permit?
+        @permit
+    end
+
+    def earn_permit 
+        if @age >= 16
```
KatBrandt, line 17: Technically there wasn't an age restriction on early a permit, only on earning a licesnse. If there was an age restriction, this would be perfect! :)

spec/registrant_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'spec_helper'
+
+RSpec.describe Registrant do
+    before(:each) do
+        @registrant_1 = Registrant.new('Bruce', 18, true )
+        @registrant_2 = Registrant.new('Penny', 16 )
+        @registrant_3 = Registrant.new('Tucker', 15 )
+        @facility_1 = Facility.new({name: 'Albany DMV Office', address: '2242 Santiam Hwy SE Albany OR 97321', phone: '541-967-2014' })
+        @facility_2 = Facility.new({name: 'Ashland DMV Office', address: '600 Tolman Creek Rd Ashland OR 97520', phone: '541-776-6092' })
```
KatBrandt, line 9: We shouldn't need any facility objects to test our Registrant class. If we do, maybe our Registrant is too dependant on another class and we should decouple those.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    describe '#earn_permit' do
+        it "updates 'permit' to true" do
+            expect(@registrant_2.permit?).to eq(false)
+            @registrant_2.earn_permit
+            expect(@registrant_2.permit?).to eq(true)
+        end
+
+        it 'cannot #earn_permit if age < 16' do
+            @facility_1.add_service('Written Test')
+            expect(@facility_1.administer_written_test(@registrant_3)).to eq(false)
```
KatBrandt, line 35: We don't want to test facility methods from the registrant_spec file. Only want to test the Registrant methods in this spec file.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,21 @@
...
+        vehicle_list.each do |vehicle|
+            output << Vehicle.new({
+                vin: vehicle[:vin_1_10], 
+                year: vehicle[:model_year], 
+                make: vehicle[:make], 
+                model: vehicle[:model], 
+                engine: :ev
+                })
+        end
+        output
```
KatBrandt, line 19: Could also use .map for this and get rid of the accumulator `output`, but this works too!

PR #226, opened 2023-05-23
Terry Lamb - DMV

TLamb32:


lib/dmv.rb
```diff
@@ -9,8 +12,9 @@ def add_facility(facility)
   end
 
   def facilities_offering_service(service)
-    @facilities.find do |facility|
+    @facilities.find_all do |facility|
```
dcoleman21, line 15: Nice job updating this to give us all vs just one.

lib/facility.rb
```diff
@@ -1,14 +1,17 @@
...
+  attr_reader :name, 
+              :address, 
+              :phone, 
+              :services
 
-  def initialize(name, address, phone)
-    @name = name
-    @address = address
-    @phone = phone
+  def initialize(facility_data)
```
dcoleman21, line 7: Great job here.

lib/registrant.rb
```diff
@@ -0,0 +1,20 @@
...
+        @permit = permit
+        @license_data = {:license=>false, :renewed=>false, :written=>false}
+    end
+
+    def permit?
+        @permit
+    end
+
+    def earn_permit
+        @permit = true
```
dcoleman21, line 18: Nice job here. You could also add what we call a guard clause like
```ruby
@permit = true if @age >= 15
```
This just makes sure that only eligible registrants can earn a permit.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    describe '#license_data' do
+        it 'has data' do
+            expect(@registrant_1.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+            expect(@registrant_2.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+        end
+    end
+
+    describe '#earn_permit' do
+        it 'can earn a permit' do
```
dcoleman21, line 44: I might add an expect statement above line 45 like
```ruby
```
That way you can really see that the `command query` is doing what you expect.

dcoleman21 (commented):
Overall, what you have looks good as far as testing, spacing, and indentation goes. I would change you default spaces to 2. You can do this by clicking
```
code -> preferences -> settings, then under Commonly Used, go to Editor: Tab Size and change that to 2
```
Keep working to get a little further each time. I left some comments for you to review. Let me know if you have any questions.

PR #227, opened 2023-05-23
Ethan Black

ethanrossblack:


lib/facility.rb
```diff
@@ -1,14 +1,76 @@
...
+    @registered_vehicles = []
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    return nil if !@services.include?("Vehicle Registration")
```
KatBrandt, line 23: Great use of a guard clause!

lib/facility.rb
```diff
@@ -1,14 +1,76 @@
...
+    return false if !@services.include?("Renew License")
+    return false if !registrant.is_a?(Registrant)
+
+    if registrant.license_data[:license] == true
+      registrant.renew_license
+      true
+    else
+      false
+    end
+  end
```
KatBrandt, line 75: Great methods! Very clean, good use of boolean logic and flow control.

spec/facility_spec.rb
```diff
@@ -1,26 +1,336 @@
...
+      expect(@facility_1.registered_vehicles).to eq([])
+    end
+    
+    it "can assign a registration_date to a Vehicle" do
+      @facility_1.add_service("Vehicle Registration")
+      
+      expect(@cruz.registration_date).to be nil
+      
+      @facility_1.register_vehicle(@cruz)
+      expect(@cruz.registration_date).to be_a(Date)
```
KatBrandt, line 63: yup, this is great. And we'll learn about mocks and stubs in week 5 that will help with this type of testing.

spec/facility_spec.rb
```diff
@@ -1,26 +1,336 @@
...
+      @facility_1.add_service("Vehicle Registration")
+
+      expect(@facility_1.collected_fees).to eq(0)
+
+      @facility_1.register_vehicle(@cruz)
+
+      expect(@facility_1.collected_fees).to eq(100)
+    end
+    
+    it "can collect registration fees from all kinds of Vehicles" do
```
KatBrandt, line 105: This test feels redundant since you just tested all of this above.

spec/facility_spec.rb
```diff
@@ -1,26 +1,336 @@
...
+    end
+    
+    it "can collect fees from all kinds of Vehicles" do
+      expect(@facility_1.collected_fees).to eq(0)
+      
+      @facility_1.collect_fee(@camaro)
+      @facility_1.collect_fee(@bolt)
+      @facility_1.collect_fee(@cruz)
+      expect(@facility_1.collected_fees).to eq(325)
+    end
```
KatBrandt, line 151: I think this describe block got added twice.

spec/facility_spec.rb
```diff
@@ -1,26 +1,336 @@
...
+    
+    it "can only administer written tests to Registrants with permits and who are at least 16 years old" do
+      @facility_1.add_service("Written Test")
+      
+      expect(@registrant_1.permit?).to be true
+      expect(@registrant_1.age >= 16).to be true
+      expect(@registrant_2.permit?).to be false
+      expect(@registrant_2.age >= 16).to be true
+      expect(@registrant_3.permit?).to be false
+      expect(@registrant_3.age >= 16).to be false
```
KatBrandt, line 191: Good testing all the different variations

spec/facility_spec.rb
```diff
@@ -1,26 +1,336 @@
...
+      expect(@registrant_2.license_data[:renewed]).to be false
+    end
+
+    it "can only renew a license for Registrant objects" do
+      @facility_1.add_service("Renew License")
+
+      expect(@facility_1.renew_drivers_license(@cruz)).to be false
+      expect(@facility_1.renew_drivers_license(@registrant_1)).to be true
     end
   end
```
KatBrandt, line 335: Your testing is very thorough and I can see that you think about all the possible outcomes and test for those. Great job!

lib/facility_factory.rb
```diff
@@ -0,0 +1,58 @@
+class FacilityFactory
+
+  def create_facilities(facility_data)
+    facility_data.map do |facility|
+      Facility.new({
+        name:    format_name(facility),
+        address: format_address(facility),
+        phone:   format_phone(facility)
```
KatBrandt, line 8: Great job making these helper methods to format this data. Unfortunately, there is not good standardization across APIs so doing this kind of checking and formatting is often necessary.

lib/facility_factory.rb
```diff
@@ -0,0 +1,58 @@
+class FacilityFactory
+
+  def create_facilities(facility_data)
+    facility_data.map do |facility|
```
KatBrandt, line 4: Great use of #map

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+      facilities = @facility_factory.create_facilities(@or_dmvs)
+      first_facility = facilities.first
+      
+      expect(facilities).to be_a(Array)
+      expect(facilities.count).to eq(59)
+      expect(first_facility).to be_a(Facility)
+    end
+  end
+
+  describe "#format_name" do
```
KatBrandt, line 28: Great job testing your helper methods!!! For these ones though, I'd actually want to test actual data. You could input your own data to ensure everything is being formatted correctly, so you could set or_facility to something specific, a hash that could come back from the api, and then test that your `format_name` method actually formats it the way you would expect. Does that make sense? If not, let me know, happy to explain further.

KatBrandt: Same with the helper methods below. I really really appreciate that you tested them. And in the future, if you know what the input to the method will be, you can create that fake data on your own (like how we setup a fake Facility object to test all the Facility methods) and ensure that your methods gives the correct output based on the fake input.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+      expect(@registrant_2.name).to eq("Penny")
+    end
+
+    it "has an @age" do
+      expect(@registrant_1.age).to eq(18)
+      expect(@registrant_2.age).to eq(15)
+    end
+    
+    it "has @license_data as a Hash" do
+      expect(@registrant_1.license_data).to be_a(Hash)
```
KatBrandt, line 27: I don't think this test is necessary because you are the one setting the Hash in the code, so it will always be a Hash.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,96 @@
+require "spec_helper"
+require "./lib/registrant"
+
+describe "Registrant" do
+  before(:each) do
+    @registrant_1 = Registrant.new("Bruce", 18, true)
+    @registrant_2 = Registrant.new("Penny", 15)
+  end
+
+  describe "#initialize" do
```
KatBrandt, line 10: Great use of describe blocks to separate tests

lib/vehicle.rb
```diff
@@ -22,4 +26,21 @@ def antique?
   def electric_vehicle?
     @engine == :ev
   end
+
+  def register
+    @registration_date = Date.today
+    assign_plate
```
KatBrandt, line 32: Nice, I like that you put this method in this class.

spec/vehicle_spec.rb
```diff
@@ -14,23 +14,66 @@
...
       expect(@camaro.electric_vehicle?).to eq(false)
     end
   end
+
+  describe "#register" do
+    it "can assign a registration_date" do
+      expect(@cruz.registration_date).to eq(nil)
+      @cruz.register
+      expect(@cruz.registration_date).to be_a(Date)
+      # is there a chance this fails if the test is run at the exact moment before midnight to the next day?
```
KatBrandt, line 43: Yup, we'll learn mocks and stubs to help us with this kind of testing.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,16 @@
+class VehicleFactory
+
+  # Only works with Washington EV car data
```
KatBrandt, line 3: This is actually a very helpful comment to leave in your code!

PR #229, opened 2023-05-23
Garrett Varenhorst

GVarenhorst:
DMV


lib/dmv.rb
```diff
@@ -1,16 +1,20 @@
...
 
   def add_facility(facility)
     @facilities << facility
   end
 
   def facilities_offering_service(service)
-    @facilities.find do |facility|
-      facility.services.include?(service)
+    select_service = []
+    @facilities.each do |facility|
```
juliet-e, line 15: This works, but consider using something that could simplify it even more for you, like .find_all

lib/dmv.rb
```diff
@@ -1,16 +1,20 @@
 class Dmv
-
+  attr_reader :facilities,
+              :service
   def initialize
     @facilities = []
+    @service = service
```
juliet-e, line 6: Is this a necessary instance variable?

lib/dmv_facility.rb
```diff
@@ -0,0 +1,28 @@
+class DmvFacility
+  def create_facilities(databases)
+    datum = databases.map do |data|
+      Facility.new({
+        name: access_name(data),
```
juliet-e, line 5: I love that you used helper methods to extract and parse this data! Great solution

lib/dmv_facility.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+  end
+
+  # require 'pry'; binding.pry
+  def access_name(datum)
+    datum[:title]
+  end
+
+  def access_address(datum)
+    datum.include?("oregon")
```
juliet-e, line 18: By having helper methods for each attribute, you set yourself up nicely for extending things to different states

lib/dmv_facility.rb
```diff
@@ -0,0 +1,28 @@
...
+      address_data = eval(datums[:location_1][:human_address])
+      address = address_data.fetch_values(:address, :city, :state, :zip).join(" ")
+    end
+
+  def access_phone(data)
+    data.include?("oregon")
```
juliet-e, line 24: It isn't very conventional to have this conditional `data.include?("oregon")` just hanging out by itself. Do you want it in an if statement or a guard clause?

lib/facility.rb
```diff
@@ -1,14 +1,63 @@
...
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    @services.include?('Vehicle Registration')
+      collect_fees(vehicle)
```
juliet-e, line 24: Great work extracting this data out into separate helper methods! REally cleans up this method

lib/facility.rb
```diff
@@ -1,14 +1,63 @@
...
+      @collected_fees += 25
+    elsif vehicle.electric_vehicle?
+      @collected_fees += 200
+    else
+      @collected_fees += 100
+    end
+  end
+
+  def administer_written_test(registrant)
+    if registrant.permit? == true && @services.include?('Written Test') && registrant.age >= 16
```
juliet-e, line 41: In an conditional statement, you actually don't need to say `== true` for a boolean variable. If you just did:
`if registrant.permit?` alone, that will evaluate to true or false and serve the same purpose. It can clean up these lines. 

Also, it's generally not super readable to have a conditional with 3 different conditions X && Y && Z. Consider breaking those up into nested if statements. Sometimes it can be tempting to make everyone one line even if it's not very readable

lib/facility.rb
```diff
@@ -1,14 +1,63 @@
...
+      registrant.license_data[:written] = true
+    else
+      false
+    end
+  end
+
+  def administer_road_test(registrant)
+    if @services.include?('Road Test') && registrant.license_data[:written] = true
```
juliet-e, line 49: be careful of having `= true` --> you meant `== true` which we already discussed isn't necessary. But `= true` is the assignment operator so it's trying to assign the value to true, which we don't want

lib/registrant.rb
```diff
@@ -0,0 +1,25 @@
...
+    @permit = permit
+    @license_data = {
+      :written=>false,
+      :license=>false,
+      :renewed=>false
+    }
+  end
+
+  def permit?
+    @permit == true
```
juliet-e, line 19: Similar to above, you can just have this line read `@permit` and it will evaluate to true or false, no need to include `== true`

lib/vehicle.rb
```diff
@@ -22,4 +26,14 @@ def antique?
   def electric_vehicle?
     @engine == :ev
   end
+
+  def set_plate_type
```
juliet-e, line 30: Since you have `set_plate_type` you don't need to make `@plate_type` an attr_accesor. You are changing it within the class, not outside the class, so you can do lines 32, 34 and 36 without needing the attr_accessor

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,13 @@
+class VehicleFactory
+  def create_vehicles(registrations)
+    auto = registrations.map do |registration|
```
juliet-e, line 3: Nice work extracting and parsing this data! You actually don't even need to set it to the variable `auto`. It can just live as it is with the .map and that will return the collection of Vehicles

lib/dmv_facility.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  # require 'pry'; binding.pry
+  def access_name(datum)
+    datum[:title]
+  end
+
+  def access_address(datum)
+    datum.include?("oregon")
+      address_data = eval(datums[:location_1][:human_address])
```
juliet-e, line 19: watch out for this typo - datum not datums

juliet-e: with this typo fixed, this works! you were very close!

spec/dmv_facility_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'spec_helper'
+
+RSpec.describe DmvFacility do
+  
+  describe "#exists" do
+    it "exists" do
+      
+      add_facility1 = DmvDataService.new.or_dmv_office_locations
+      # require 'pry'; binding.pry
+      expect(add_facility1).to be_a(Array)
```
juliet-e, line 10: Keep in mind that you don't need to make assertions on what the data looks like BEFORE it goes into your DmvFacility.createFacilities method. We just want to assert on what comes out.

spec/dmv_facility_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+      expect(add_facility1[0][:title]).to eq("Albany DMV Office")
+        expect(add_facility1[1][:title]).to eq("Ashland DMV Office")
+        # expect(add_facility1.create_facilities.access_name).to eq("Albany DMV Office")
```
juliet-e, line 13: You need to initialize your DmvFacility object somewhere so you can call `.create_facilities`. Here you are trying to call it on the data that we pass into `create_facilities` but you need to call it on the DmvFacility object. add_facility1 is the collection of data we're passing in as an argument to that method

PR #230, opened 2023-05-24
Project Submission

ChrisAsercion:


lib/dmv.rb
```diff
@@ -9,7 +9,7 @@ def add_facility(facility)
   end
 
   def facilities_offering_service(service)
-    @facilities.find do |facility|
+    @facilities.find_all do |facility|
```
KatBrandt, line 12: Great use of find_all here.

lib/facility.rb
```diff
@@ -1,14 +1,74 @@
...
+        
+      elsif vehicle.antique? == true
+        vehicle.plate_version(:antique)
+        @collected_fees += 25
+      else #regular
+        vehicle.plate_version(:regular)
+        @collected_fees += 100
+      end
+    else
+      nil
```
KatBrandt, line 40: Yup, good job with your flow control here. In general, with an if statement, if you're doing a comparison to a boolean value, you usually don't need to. For example, line 28, if you just had `if vehicle.electric_vehicle?` that would work fine because that statement already gives you back a true/false value.

lib/facility.rb
```diff
@@ -1,14 +1,74 @@
...
+    end
   end
 
-  def add_services(service)
-    @services << service
+  def administer_written_test(student)
+    if @services.include?('Written Test') && student.age >= 16 && student.permit == true
```
KatBrandt, line 45: Nice job with this boolean logic.

spec/facility_spec.rb
```diff
@@ -22,5 +27,82 @@
...
       expect(@facility.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
     end
+
+    it '#has to have the service in order to do the function' do
+      registrant_1 = Registrant.new('Bruce', 18, true )
+      expect(registrant_1.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+      expect(registrant_1.permit?).to eq(true)
+      expect(@facility_1.administer_written_test(registrant_1)).to eq(false)
+      @facility_1.add_service('Written Test')
+      @facility_1.administer_written_test(registrant_1)
```
KatBrandt, line 37: Where is the assertion that this is now true? You have line 35 that shows me it's false, then you add the service, so I'd want to see an assertion that shows me this is now working and returning true or has the desired side effects.

spec/facility_spec.rb
```diff
@@ -22,5 +27,82 @@
...
+  end
+  describe '#administer_written_test' do
+    it 'can give written tests' do
+      registrant_1 = Registrant.new('Bruce', 18, true )
+      @facility_1.add_service('Written Test')
+      expect(@facility_1.services).to eq(["Written Test"])
+      expect(@facility_1.administer_written_test(registrant_1)).to eq(true)
+      expect(registrant_1.license_data).to eq({:written=>true, :license=>false, :renewed=>false})
+    end
+    it 'can check attributes required for test taking' do
```
KatBrandt, line 66: Let's add one line of white space between an `end` and `def`

spec/facility_spec.rb
```diff
@@ -22,5 +27,82 @@
...
+      expect(@facility_1.services).to eq(["Written Test", "Road Test"])
+      expect(@facility_1.administer_road_test(registrant_1)).to eq(true)
+      expect(registrant_1.license_data).to eq({:written=>true, :license=>true, :renewed=>false})
+    end
+    it 'can check attributes required to take road test' do
+      registrant_3 = Registrant.new('Tucker', 15 )
+      @facility_1.add_service('Written Test')
+      @facility_1.add_service('Road Test')
+      expect(registrant_3.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+    end
```
KatBrandt, line 94: I'm not sure what this test is accomplishing. We haven't done anything so I would definitely expect this license_data to still be all false.

lib/facility_handler.rb
```diff
@@ -0,0 +1,83 @@
+
+class FacilityHandler
+  def state_sorter(dmv_locations)
```
KatBrandt, line 3: yeah, nice, I like this helper method

lib/facility_handler.rb
```diff
@@ -0,0 +1,83 @@
...
+    elsif dmv_locations[1][:state].include?("NY")
+      create_new_york_facilities(dmv_locations)
+    else dmv_locations[1][:website].include?('oregon')
+      create_oregon_facilities(dmv_locations)
+    end
+  end
+  #this method will sort through each JSON file to confirm which state the facility builder should do
+
+  def create_oregon_facilities(dmv_locations)
+      oregon_facility = dmv_locations.map do |data|
```
KatBrandt, line 15: Good use of map

lib/facility_handler.rb
```diff
@@ -0,0 +1,83 @@
...
+        Facility.new(name:  data[:title],
+        phone: data[:phone_number], 
+        address: address_formatter_oregon(data))
+      end
+  end
+  def create_missouri_facilities(dmv_locations)
```
KatBrandt, line 21: Line of white space between end of method and def of a new one.

lib/facility_handler.rb
```diff
@@ -0,0 +1,83 @@
...
+  def address_formatter_oregon(data)
+    address = JSON.parse(data[:location_1][:human_address])
+    address.values.join(", ")
+  end
+
+  def address_formatter_missouri(data)
+    address = "#{data[:address1]}, #{data[:city]}, #{data[:state]}, #{data[:zipcode]}"
+  end
+
+  def address_formatter_new_york(data)
```
KatBrandt, line 46: Yeah, I really like all these helper methods you used.

spec/facility_handler_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    missouri_facilities.each do |facility|
+      expect(facility.address.include?('MO')).to eq(true) 
+    end
+    oregon = DmvDataService.new.or_dmv_office_locations
+    oregon_facilities = facility_array.create_oregon_facilities(oregon)
+    oregon_facilities.each do |facility|
+      expect(facility.address.include?('OR')).to eq(true)
+    end
+  end
+end
```
KatBrandt, line 49: Let's make sure to test every single method we have in a class. So every method and helper method in your FacilityHandler class should be tested in here. And every possible return value from a method should be tested. Your helper methods are tested indirectly via the Integration tests you have for create_oregon_facilities, but we want to have unit tests to ensure that with a given input, the helper method gives an expected output.

lib/registrant.rb
```diff
@@ -0,0 +1,27 @@
+class Registrant
+  attr_reader :name, :age, :permit
```
KatBrandt, line 2: We don't need/want the `:permit` in the attr_reader since we don't want to expose that method. We have the `permit?` acting as a reader.

lib/registrant.rb
```diff
@@ -0,0 +1,27 @@
...
+    @age = age
+    @permit = permit
+    @license_data = 
+      {:written=>false, 
+      :license=>false, 
+      :renewed=>false}
+  end
+
+  def permit?
+    if permit == true
```
KatBrandt, line 16: What if the only thing in this method was `@permit`. Would it return what you want?

spec/registrant_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+      :license=>false, 
+      :renewed=>false})
+    end
+    it "can add registrants with different attributes" do
+      registrant_1 = Registrant.new('Bruce', 18, true)
+      registrant_2 = Registrant.new('Penny', 15 )
+      expect(registrant_1.name).to eq("Bruce")
+      expect(registrant_2.name).to eq("Penny")
+      expect(registrant_1.permit?).to eq(true)
+      expect(registrant_2.permit?).to eq(false)
```
KatBrandt, line 21: We also need to test our earn_permit method, to ensure that it changes permit? from false to true.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,20 @@
...
+  def create_vehicles(car_registrations)
+    vehicle_key_editor(car_registrations)
+    car_registrations.map do |car|
+      Vehicle.new(car)
+    end
+  end
+  def vehicle_key_editor(car_registrations)
+    car_registrations.each do |parts|
+      parts[:vin] = parts[:vin_1_10]
+      parts[:model] = parts[:model]
```
KatBrandt, line 13: No need to change the key if the key doesn't need to be changed. :)

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,20 @@
+
+
+class VehicleFactory 
+  def create_vehicles(car_registrations)
+    vehicle_key_editor(car_registrations)
+    car_registrations.map do |car|
+      Vehicle.new(car)
+    end
+  end
+  def vehicle_key_editor(car_registrations)
```
KatBrandt, line 10: I do like this helper method

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    it 'takes in car registrations and makes them Vehicle objects' do
+      factory = VehicleFactory.new
+      @wa_ev_registrations = DmvDataService.new.wa_ev_registrations
+      vehicle_array = factory.create_vehicles(@wa_ev_registrations)
+      expect(vehicle_array.first).to be_an_instance_of(Vehicle)
+      expect(vehicle_array).to be_a(Array)
+      expect(vehicle_array.first.make).to eq("TESLA")
+      expect(vehicle_array.first.vin).to eq("7SAYGDEF6N")
+    end
+  end
```
KatBrandt, line 23: And we'll also want to test the vehicle_key_editor method you wrote as well.

PR #234, opened 2023-05-25
Blake Sergesketter 

bserge13:


lib/dmv.rb
```diff
@@ -9,8 +14,12 @@ def add_facility(facility)
...
-      facility.services.include?(service)
-    end
+    selected_facility = []
+    @facilities.each do |facility|
+     if facility.services.include?(service)
+        selected_facility.push(facility)
+      end
+    end   
+    selected_facility
   end
```
megstang, line 24: Nice work here, good debugging! You could also use the `find_all` or `select` method here. What you have is great, but those are enums that could slim up this code a bit. It would look something like: 

```ruby
    @facilities.find_all do |facility|
    end   
  end
```

lib/vehicle.rb
```diff
@@ -22,4 +26,15 @@ def antique?
...
+  def register
+    @registration_date = Date.today
+     if self.antique? 
+        @plate_type = :antique
+     elsif self.electric_vehicle?
+        @plate_type = :ev 
+     else 
+      @plate_type = :regular 
+     end
+  end
```
megstang, line 39: This method looks really great! I like that you created it in this class.

lib/facility.rb
```diff
@@ -1,14 +1,61 @@
...
+        @collected_fees += 100
+      end
+
+      @registered_vehicles << vehicle 
+    end 
+  end
+
+  def administer_written_test(registrant)
+    if @services.include?("Written Test") && registrant.permit? && registrant.age >= 16
+      true
```
megstang, line 37: we probably don't need this line

lib/facility.rb
```diff
@@ -1,14 +1,61 @@
...
+      registrant.license_data[:written] = true
+    else 
+      false 
+    end
+  end
+
+  def administer_road_test(registrant)
+    if @services.include?("Road Test") && registrant.permit? && registrant.age >= 16
+      true
```
megstang, line 46: same thing here, we probably don't need line 46.

lib/registrant.rb
```diff
@@ -0,0 +1,20 @@
...
+    end
+
+    def earn_permit
+        @permit = true
+    end
+
+    def permit?
+        @permit
+    end
+end
```
megstang, line 20: 👍

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,21 @@
...
+    end
+
+    def create_facility(facility_locations)
+        facility_locations.each do |facility|
+            facility_details = ({name: facility[:name], address: facility[:address], phone: facility[:phone]})
+            facility_locations = Facility.new(facility_locations)
+            return locations 
+        end 
+    end
+end
```
megstang, line 21: I think you're off to a good start here. I'll attach a possible option for how this class could be set up: 

```ruby 
class VehicleFactory

  def create_vehicles(vehicle_dataset)
    vehicle_dataset.map do |vehicle_details|
      parsed_data = {
      vin: vehicle_details[:vin_1_10],
      year: vehicle_details[:model_year],
      make: vehicle_details[:make],
      model: vehicle_details[:model],
      engine: :ev
    }
      Vehicle.new(parsed_data)
    end
  end
end
```


Because there is no need to instantiate this object with arguments (i.e. we don't need to pass anything in when we call VehicleFactory.new), we don't need an initialize method. So, the one method that you see in this example is an instance method that can be called on any instance of the VehicleFactory class. 

```ruby
vehicle_factory = VehicleFactory.new
VehicleFactory.create_vehicles(data)
```

spec/vehicle_spec.rb
```diff
@@ -33,4 +33,21 @@
...
+      @cruz.register 
+      @bolt.register 
+      @camaro.register 
+      expect(@cruz.registration_date).to eq(Date.today)
+      expect(@bolt.registration_date).to eq(Date.today)
+      expect(@camaro.registration_date).to eq(Date.today)
+      expect(@cruz.plate_type).to eq(:regular)
+      expect(@bolt.plate_type).to eq(:ev)
+      expect(@camaro.plate_type).to eq(:antique)
+    end
```
megstang, line 51: nice! Good job remembering to test your helper methods that you write.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+            expect(@facility_1.renew_drivers_license(@registrant_1)).to eq(true)
+            expect(@registrant_1.license_data).to eq({:written=>true, :license=>true, :renewed=>true})
+
+            expect(@facility_1.renew_drivers_license(@registrant_3)).to eq(false)
+            expect(@registrant_3.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+            expect(@facility_1.renew_drivers_license(@registrant_2)).to eq(true)
+            expect(@registrant_2.license_data).to eq({:written=>true, :license=>true, :renewed=>true})
+        end
+    end
+end
```
megstang, line 103: just a general comment on testing -- your tests are looking really good. I like that you have a lot of different instance that you test your methods on, that is what makes robust tests!

megstang (commented):
@bserge13 I left some comments for you to take a look over. Please let me know if you ahve any questions!

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,21 @@
+class VehicleFactory 
+
+    def initialize 
+    end
+
+    def create_vehicles(registration_locations)
+        registration_locations.each do |vehicle|
+            vehicle_details = ({vin: vehicle[:vin_1_10] , year: vehicle[:model_year], make: vehicle[:make], model: vehicle[:model], engine: [:ev]})
+            new_vehicles = Vehicle.new(vehicle_details)
+            return new_vehicles
```
megstang, line 10: This explicit return will only return this one vehicle that you're iterating through. We want to create a collection of vehicles and return the entire collection of vehicles. A placeholder might be helpful to shovel in each vehicle that you create, and then you'll need to return that placeholder after you're done iterating. 

```ruby 
vehicles = []
registration_locations.each do |vehicle|
            vehicle_details = ({vin: vehicle[:vin_1_10] , year: vehicle[:model_year], make: vehicle[:make], model: vehicle[:model], engine: [:ev]})
            vehicles << Vehicle.new(vehicle_details)
     end  
   vehicles
end 
```

PR #235, opened 2023-05-25
Davis Weimer

DavisWeimer:


spec/facility_spec.rb
```diff
@@ -24,3 +24,237 @@
...
+    @facility_2 = Facility.new({name: 'Ashland DMV Office', address: '600 Tolman Creek Rd Ashland OR 97520', phone: '541-776-6092' })
+    @cruz = Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice} )
+    @bolt = Vehicle.new({vin: '987654321abcdefgh', year: 2019, make: 'Chevrolet', model: 'Bolt', engine: :ev} )
+    @camaro = Vehicle.new({vin: '1a2b3c4d5e6f', year: 1969, make: 'Chevrolet', model: 'Camaro', engine: :ice} )
+  end
+  describe "Iteration 2" do
+    it 'can add DMV service' do
+      expect(@facility_1.services).to eq([])
+      @facility_1.add_service('Vehicle Registration')
+      expect(@facility_1.services).to eq(["Vehicle Registration"])
```
KatBrandt, line 43: Good job testing the initial value, and then the change that the method caused.

spec/facility_spec.rb
```diff
@@ -24,3 +24,237 @@
...
+    @facility_1 = Facility.new({name: 'Albany DMV Office', address: '2242 Santiam Hwy SE Albany OR 97321', phone: '541-967-2014' })
+    @facility_2 = Facility.new({name: 'Ashland DMV Office', address: '600 Tolman Creek Rd Ashland OR 97520', phone: '541-776-6092' })
+  end
+  describe "Iteration 2" do
+    it 'Registrants and Facilities exist' do
+      expect(@registrant_1).to be_an_instance_of(Registrant) 
+      expect(@registrant_2).to be_an_instance_of(Registrant) 
+      expect(@registrant_3).to be_an_instance_of(Registrant)
+      expect(@facility_1).to be_an_instance_of(Facility)
+      expect(@facility_2).to be_an_instance_of(Facility)
```
KatBrandt, line 121: Not necessary to test the Registrants in this class since you already do that in the Registrant class. Also, really only need to test one instance of Facility for this test.

spec/facility_spec.rb
```diff
@@ -24,3 +24,237 @@
...
+    # ---------- ROAD TEST ----------
+    
+    it 'cannot administer Road Test without service
+    or to registrant without permit and written test' do
+    @facility_1.add_service('Written Test')
+    expect(@facility_1.services).to eq(['Written Test'])
+    @facility_1.add_service('Road Test')
+    @facility_1.administer_road_test(@registrant_3)
+    expect(@facility_1.administer_road_test(@registrant_3)).to eq false
+
```
KatBrandt, line 179: Good test here for testing that the method does nothing before the service has been added.

lib/vehicle.rb
```diff
@@ -22,4 +26,29 @@ def antique?
   def electric_vehicle?
     @engine == :ev
   end
+
+  def date_registered
```
KatBrandt, line 30: usually method names are verbs since they perform an action, so naming this `add_registered_date` would be better or something like that.

lib/vehicle.rb
```diff
@@ -22,4 +26,29 @@ def antique?
...
+  end
+
+  def plate_finder
+    if antique?
+      :antique
+    elsif electric_vehicle?
+      :ev
+    elsif @engine == :ice
+      :regular
+    else nil
```
KatBrandt, line 41: nil should be below, else gets its own line, it's special like that. :)

KatBrandt (commented):
Overall this is looking great!!!

PR #311, opened 2023-09-02
Feature/vehicle registration

Sykogst:
Update classes to provide functionality for vehicle registration.


Sykogst (commented):
Maybe find a way to shorten conditional statement block(s) for register_vehicle method in Facility class

PR #314, opened 2023-09-02
Taylor "TP" Pubins

trpubz:


lib/dmv.rb
```diff
@@ -9,8 +11,35 @@ def add_facility(facility)
...
+          key
+        end
+      end
+
+      d[:address] = [d[:address1], d[:address2], d[:city], d[:state], d[:zip]].join(' ')
+
+      @facilities << Facility.new(d)
+     end
+  end
 end
```
abdulredd, line 45: Great job in this class, your `parse_facilities` method is superb. Great use of the `case` statement. I'm not a huge fan of single-letter variable names, however 😉.

I will add that you can eliminate the need to shovel newly created `Facility` objects into an array but switching out `.each` for `.map`.

lib/data_query_controller.rb
```diff
@@ -0,0 +1,122 @@
...
+  **Format For Queries: <Opt#> <STATE> <ARGS>**
+
+  1. Summary of WA or NY registrations:
+  2. Registered vehicle count for a model year:
+  3. Registered vehicle count by engine type (gas, diesel, flex, ev):
+
+  q: (Quit)
+  "
+
+end
```
abdulredd, line 118: Quite a bit going on in here, but I spent my time looking through and don't have any pressing suggestions and think this class is great. Good job!

I believe you wanted additional feedback on your query syntax and readability. I think it does take a little to work through but I don't think there is anything incorrect here and method chaining is common practice.

lib/facility.rb
```diff
@@ -1,14 +1,67 @@
...
+
+  def renew_drivers_license(registrant)
+    if @services.include?("Renew License") && registrant.license?
+      registrant.renew_license
+      return true
+    else
+      return false
+    end
+  end
 end
```
abdulredd, line 67: 👍

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,27 @@
...
+        puts "Unrecognized engine type: #{d[:fuel_type]}"
+      end
+
+
+      vehicles << Vehicle.new(d)
+    end
+
+    return vehicles
+  end
+end
```
abdulredd, line 27: Not really much to say about this class as well, really great job. I do have a little nitpicky suggestion, and that is to be cautious when using bang methods (`!`) because of how they modify the original hash. Doesn't look like you needed the original hash here for anything important so no surprising side effects. :)

PR #317, opened 2023-09-04
Feature/add registrant class

hdmcneil:
Added registrant class with basic parameters set for license date, name, age, and permit. Added setter methods for permit? and earn_permit.

On Vehicle class, setter methods were added for registration_date and plate_type.

On Facility class, method for adding services to a facility, a method to check if a service is available at a facility, and a registration method were added.

Testing was added for all of this functionality.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -23,6 +23,7 @@ GEM
 
 PLATFORMS
   arm64-darwin-20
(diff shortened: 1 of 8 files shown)


hdmcneil:
I would like to tmerge this to the main branch

PR #327, opened 2023-09-05
Xander Hendry

XanderHendry:


lib/facility.rb
```diff
@@ -1,14 +1,91 @@
...
+    @name = facility_details[:name]
+    @address = facility_details[:address]
+    @phone = facility_details[:phone]
     @services = []
   end
-
-  def add_services(service)
+  
+  def add_service(service)
+    self.collected_fees_helper
```
jamisonordway, line 18: (this applies to many lines of your code)
In Ruby, the use of `self` within a method is implied. This could be
```
def add_service(service)
  collected_fees_helper
  ...
end
```

lib/facility.rb
```diff
@@ -1,14 +1,91 @@
...
+  def collected_fees
+    self.collected_fees_helper
+    @collected_fees
+  end
+
+  def register_vehicle_helper
+    if @registered_vehicles.nil?
+      @registered_vehicles = []
+    end
+  end
```
jamisonordway, line 39: I would use caution declaring/assigning values to ivars outside of the `initialize` method. It can cause issues with clarity and code readability down the road as projects scale. 
In my opinion it would be fine to instantiate every `Facility` with `@registered_vehicles = []` so that you wouldn't need this method.

lib/facility.rb
```diff
@@ -1,14 +1,91 @@
...
+  def administer_written_test(registrant)
+    if @services.include?('Written Test')
+      if registrant.permit? && registrant.age > 15
+        registrant.license_data[:written] = true
+      else
+        false
+      end
+    else
+      false
+    end
```
jamisonordway, line 73: You could 'flatten' this conditional with a guard clause, something like
```
return false if !@services.include?('Written Test')
```
That way, the condition about what services are included doesn't need to wrap the other conditional.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,34 @@
...
+      vehicle_details = {vin: car[:vin_1_10], year: car[:model_year].to_i, make: car[:make], model: car[:model], engine: :ev}
+    end
+  end
+
+  def create_vehicles(data)
+    self.create_vehicles_helper(data)
+    cars = @production_list.map do |car|
+      Vehicle.new(car)
+    end
+  end
```
jamisonordway, line 32: I really like this use of a helper method, making `create_vehicles` much more straightforward with a single focused responsibility of creating an array of new `Vehicle` instances.

spec/facility_spec.rb
```diff
@@ -1,26 +1,230 @@
...
+    before(:each) do
+      @registrant_1 = Registrant.new('Bruce', 18, true )
+      @registrant_2 = Registrant.new('Penny', 16 )
+      @registrant_3 = Registrant.new('Tucker', 15 )
+    end
+    context 'when adminstering written tests to qualified registrants' do
+      before(:each) { @facility_1.add_service('Written Test')}
+      it 'Will change a registrants license data to show written as true' do
+        expect(@facility_1.administer_written_test(@registrant_1)).to eq(true)
+        expect(@registrant_1.license_data[:written]).to eq(true)
```
jamisonordway, line 133: For tests like this one, consider adding an assertion before the execution method call (in this case `administer_written_test`. That way, the test clearly demonstrates that the state starts out with one value and is changed after the relevant call.
```
expect(@registrant_1.license_data[:written]).to eq(false)
expect(@facility_1.administer_written_test(@registrant_1)).to eq(true)
expect(@registrant_1.license_data[:written]).to eq(true)
```

spec/registrant_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    it 'has a permit that defaults to false' do
+      expect(registrant_2.permit?).to eq(false)
+    end
+
+    it 'will change the value of permit to true' do
+      registrant_2.earn_permit
+      expect(registrant_2.permit?).to eq(true)
+    end
+
+    it 'will not change the value of permit to false' do
```
jamisonordway, line 34: Great edge case to consider and test against.

jamisonordway (commented):
Awesome work overall! I'm really impressed with how organized and thorough your tests are, and I left you some feedback with ideas about ways to make your code more readable/succinct/conventional. Please let me know if you have questions about any of these comments.

PR #328, opened 2023-09-05
Scott DeVoss

scottdevoss:


lib/facility.rb
```diff
@@ -1,14 +1,65 @@
...
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    return nil if !@services.include?("Vehicle Registration")
+    if @services.include?("Vehicle Registration")
```
jamisonordway, line 21: If `!@services.include?("Vehicle Registration")` is `true`, this method will return `nil`, never reaching line 21. Therefore, line 21 will never evaluate to false, so you could remove this outer 'if' block.

lib/facility.rb
```diff
@@ -1,14 +1,65 @@
...
+      @registered_vehicles << vehicle
+      vehicle.registration_date = Date.today
+      if vehicle.plate_type == :antique
+        @collected_fees += 25
+        elsif vehicle.plate_type == :ev
+        @collected_fees += 200
+        elsif vehicle.plate_type == :regular
+        @collected_fees += 100
+      end
```
jamisonordway, line 30: 1. Keep an eye on your indentation
2. Conventionally, every `if` block with an `elsif` should also have an `else` as a catch-all. In this case, I'd suggest
```
if vehicle.plate_type == :antique
  @collected_fees += 25
elsif vehicle.plate_type == :ev
  @collected_fees += 200
else # all other vehicles are regular. Else doesn't take a condition.
  @collected_fees += 100
end
```

lib/facility.rb
```diff
@@ -1,14 +1,65 @@
...
+    end
+    @registered_vehicles
+  end
+
+  def collected_fees
+    @collected_fees
+  end
+
+  def administer_written_test(registrant)
+    if @services.include?("Written Test") && registrant.permit? == true && registrant.age >= 16
```
jamisonordway, line 40: I think abstracting some of these long lines to a helper method could be more readable.
```
return false if !@services.include?("Written Test")
if written_eligible?(registrant)
```
```
def written_eligible?(registrant)
  registrant.permit? && registrant.age >= 16
end
```
Note that we can omit `== true` from this because `registrant.permit?` will already return a boolean value.

lib/facility_factory.rb
```diff
@@ -0,0 +1,28 @@
...
+      end 
+    elsif data == DmvDataService.new.mo_dmv_office_locations
+      data.map do |facility_data|
+      address = "#{facility_data[:address1]}, #{facility_data[:city]}, #{facility_data[:state]}, #{facility_data[:zipcode]}"
+      facility = Facility.new({name: facility_data[:name],
+        address: address,
+        phone: facility_data[:phone]})
+      end 
+    end 
+  end
```
jamisonordway, line 25: Great question about how to refactor this-- it works well 🎉 but got pretty long and difficult to read.
1. If the return value of `DmvDataService.new.co_dmv_office_locations` for example, is passed to this method as an argument, we shouldn't call that method again if we can avoid it. Especially since this method calls the 'outside world' to get data from an external API, which can be an expensive process (meaning long, but also, you will eventually work with APIs that cost money to ping). 
I would look for some value in the nested dataset which could identify which state it's coming from, and use that in the conditional instead.
2. As far as refactoring, I might start with a helper method for each state which could be called in this method's `maps`:
```
def create_facilities(data)
  if data[:some_value][:some_unique_co_value] # pardon my pseudocode 
    data.map { |facility_data| new_co_facility(facility_data) }
  elsif data[:some_unique_ny_value]
    data.map { |facility_data| new_ny_facility(facility_data) }
  else
    data.map { |facility_data| new_mo_facility(facility_data) }
  end
end
```

lib/registrant.rb
```diff
@@ -0,0 +1,22 @@
...
+    @license_data = license_data
+  end
+
+  def permit?
+    if @permit == true
+      true
+    else 
+      false
+    end
+  end
```
jamisonordway, line 17: Nothing wrong with this conditional logic, it works. That said, since `@permit` already returns a boolean, we could simply return `@permit`.

lib/vehicle.rb
```diff
@@ -22,4 +27,18 @@ def antique?
...
+
+  def plate_type
+    if @year <= Date.today.year.to_i - 25
+      @plate_type = :antique
+    elsif @engine == :ev
+      @plate_type = :ev
+    else 
+      @plate_type = :regular
+    end
+  end
```
jamisonordway, line 43: Is there value in having a `plate_type` method in addition to the `plate_type` attribute above, which has an attr_accessor?
Every time `vehicle_instance.plate_type` is called, it will run through this logic _and_ update the attribute. Maybe another method could be responsible for updating the `@plate_type` attribute, say, at the time of vehicle registration. That way when we call `vehicle_instance.plate_type`, we're actually calling the attribute directly.

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+  end
+
+  describe '#create facilities' do
+    it 'can create a facility from provided data' do
+      factory = FacilityFactory.new
+      co_dmv_office_locations = DmvDataService.new.co_dmv_office_locations
+      result = factory.create_facilities(co_dmv_office_locations)
+      expect(result.class).to eq(Array)
+      expect(result.first.class).to eq(Facility)
```
jamisonordway, line 17: I think it would be valuable to also `expect(result.last.class).to eq(Facility)`. Alternatively:
```
result.each do |facility|
  expect(facility).to be_instance_of(Facility)
  # could test the data types of each Facility instance's attributes here, too
end
```

spec/facility_spec.rb
```diff
@@ -23,4 +23,115 @@
...
+      expect(facility_1.administer_written_test(registrant_1)).to be true
+      expect(registrant_1.license_data).to eq({:written=>true, :license=>false, :renewed=>false})
+      expect(facility_1.administer_written_test(registrant_2)).to be false
+      registrant_2.earn_permit
+      
+      expect(facility_1.administer_written_test(registrant_2)).to be true
+      expect(facility_1.administer_written_test(registrant_3)).to be false
+      registrant_3.earn_permit
+      expect(facility_1.administer_written_test(registrant_3)).to be false
+    end
```
jamisonordway, line 83: It's great that you're testing all the scenarios having to do with facility and registrant eligibility 👍🏻 
Time/bandwidth permitting, you could consider splitting tests like these into multiple `it` or even `describe` blocks for better readability.

jamisonordway (commented):
Great work on this project! Please let me know if you have any follow up questions about my feedback.

PR #329, opened 2023-09-05
Kevin Zolman

zkevkev:


lib/registrant.rb
```diff
@@ -0,0 +1,21 @@
...
+              :age, 
+              :license_data, 
+              :permit
+
+  def initialize(name, age, permit = false)
+    @name = name
+    @age = age
+    @license_data = {:written=>false, :license=>false, :renewed=>false}
+    @permit = permit
+  end
```
megstang, line 12: This looks great. I like that you are using a default value for `permit` upon initialization.

lib/vehicle.rb
```diff
@@ -6,13 +6,18 @@ class Vehicle
...
 
   def initialize(vehicle_details)
     @vin = vehicle_details[:vin]
     @year = vehicle_details[:year]
     @make = vehicle_details[:make]
     @model = vehicle_details[:model]
     @engine = vehicle_details[:engine]
+    @registration_date = nil
+    @plate_type = nil
   end
```
megstang, line 21: 👍

lib/facility.rb
```diff
@@ -1,14 +1,82 @@
...
+    @collected_fees += 25
+    registration_helper(vehicle)
+  end
+
+  def register_reg(vehicle)
+    return if !@services.include?('Vehicle Registration')
+    vehicle.plate_type = :regular
+    @collected_fees += 100
+    registration_helper(vehicle)
+  end
```
megstang, line 47: I LOVE that you are creating helper methods to assist in registering specific vehicle types. This is really great.

lib/facility.rb
```diff
@@ -1,14 +1,82 @@
...
+    end
+  end
+
+  def renew_drivers_license(registrant)
+    if registrant.license_data[:license] == true && @services.include?('Renew License')
+      registrant.license_data[:renewed] = true
+    else
+      false
+    end
+  end
```
megstang, line 81: Your logic in these methods looks really good. Nice work!

lib/facility_factory.rb
```diff
@@ -0,0 +1,36 @@
+class FacilityFactory
+  
+  def initialize
+  end
+
```
megstang, line 5: Since you do not need to initialize the `FacilityFactory` objects with any state (attributes), you don't need this `initialize` method. We really only need to have that when we want to instantiate an object with state from the very start.

lib/facility_factory.rb
```diff
@@ -0,0 +1,36 @@
...
+  def normalize_phone(phone_string)
+    if phone_string != nil && !phone_string.start_with?('(')
+      phone_string.insert(0, '(').insert(4, ')').insert(5, ' ').insert(9, '-')
+    else return phone_string
+    end
+  end
```
megstang, line 11: Nice job here! I like that you've got another method to help format your phone number.

megstang (commented):
@zkevkev Nice work on this project! I'll follow up in slack within the next day or so with a bit more feedback. In the meantime, please take a look at my comments and let me know if you have any questions.

PR #330, opened 2023-09-05
Joseph Lee

josephlee702:


lib/facility.rb
```diff
@@ -1,14 +1,73 @@
...
+    end
+  end
+
+  #helper method
+  def check_for_service(service)
+    @services.include?(service)
+  end
+
+  def register_vehicle(car)
+    if check_for_service('Vehicle Registration') == true
```
jamisonordway, line 34: This is a really cool use of a helper method. A couple of thoughts:
1. You could omit the `== true` since `check_for_service('Vehicle Registration')` will already return a boolean, which would work on its own for the conditional logic. 
2. Per convention, you could define the helper method with a `?` at the end of the name to indicate it will return a boolean value.
```
def has_service?(service)
  ...
end
```
3. By using a guard clause, you could avoid the need to nest conditional statements in this method:
```
def register_vehicle(car)
  return false unless has_service?('Vehicle Registration')
  @registered_vehicles << car
  ...
```

lib/facility.rb
```diff
@@ -1,14 +1,73 @@
...
+      @registered_vehicles << car
+      car.registration_date = Date.today
+      car.plate_type = set_plate_type(car)
+      if car.antique?
+        @collected_fees += 25
+      elsif car.electric_vehicle?
+        @collected_fees += 200
+      else 
+        @collected_fees += 100
+      end
```
jamisonordway, line 44: Could this also live in a helper method?

lib/facility.rb
```diff
@@ -1,14 +1,73 @@
...
+  def administer_road_test(registrant)
+    if check_for_service('Road Test') && registrant.license_data[:written]
+      registrant.license_data[:license] = true
+    else
+      false
+    end
+  end
+
+  def renew_drivers_license(registrant)
+    if check_for_service('Renew License') && registrant.license_data[:written] && registrant.license_data[:license]
```
jamisonordway, line 67: This line got a bit long. I might consider making it a helper method.
```
def renew_eligible?(registrant)
  written = registrant.license_data[:written]
  license = registrant.license_data[:license]
  check_for_service('Renew License') && written && license
end
```

lib/facility_factory.rb
```diff
@@ -0,0 +1,31 @@
...
+          address: "#{facility[:street_address_line_1]} #{facility[:street_address_line_2]} #{facility[:city]} #{facility[:state]} #{facility[:zip_code]}",
+          phone: facility[:public_phone_number]
+        )
+      else facility[:name] != nil
+        Facility.new(
+          name: facility[:name],
+          address: "#{facility[:address1]} #{facility[:city]} #{facility[:state]} #{facility[:zipcode]}",
+          phone: facility[:phone]
+        )
+      end
```
jamisonordway, line 28: You could consider creating a helper method for each of these conditions, especially since the data parsing logic is a little bit different across all of them.
```
if facility[:dmv_office] != nil
  new_co_facility(facility)
elsif facility[:office_name] != nil
  new_ny_facility(facility)
else # note - unlike elsif, else does not take a condition
  new_mo_facility(facility)
end
```

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,16 @@
...
+  end
+
+  def create_vehicles(registration)
+    vehicles = registration.map do |vehicle|
+      vin = vehicle[:vin_1_10]
+      year = vehicle[:model_year]
+      make = vehicle[:make]
+      model = vehicle[:model]
+
+      vehicle = Vehicle.new(vin: vin, year: year, make: make, model: model, engine: :ev)
```
jamisonordway, line 13: Another common syntax/format for this type of pattern:
```
vehicle = Vehicle.new(
  vin: vehicle[:vin_1_10],
  year: vehicle[:model_year],
  make: vehicle[:make],
  model: vehicle[:model]
)
```

spec/facility_factory.spec.rb
```diff
@@ -0,0 +1,39 @@
...
+RSpec.describe Facility_Factory do
+  describe '#CO_DMV facility data' do
+    it 'creates "CO facilities' do
+      facility = Facility_Factory.new
+      colorado_facilities = DmvDataService.new.co_dmv_office_locations
+
+      expect(facility.create_office_locations(colorado_facilities)[0]).to be_an_instance_of (Facility)
+      facility.create_office_locations(colorado_facilities).each do |data|
+        expect(data).to be_instance_of (Facility)
+      end
```
jamisonordway, line 12: With the block you opened on line 10, you no longer need line 9.

spec/facility_factory.spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  describe '#NY_DMV facility data' do
+    it 'creates "NY facilities' do
+      facility = Facility_Factory.new
+      NY_facilities = DmvDataService.new.ny_dmv_office_locations
+
+     expect(facility.create_office_locations(NY_facilities)[0]).to be_an_instance_of (Facility)
+     facility.create_office_locations(NY_facilities).each do |data|
+      expect(data).to be_instance_of (Facility)
```
jamisonordway, line 23: In tests like these, it may also be a good idea to verify the presence and/or data types of each instance's attributes.

spec/facility_spec.rb
```diff
@@ -2,25 +2,266 @@
...
+      @facility_1.add_service('Vehicle Registration')
+      @facility_1.register_vehicle(@cruz)
+      expect(@cruz.registration_date).to eq(Date.today)
+      expect(@cruz.plate_type).to eq(:regular)
+      #require the facility to have the service you're performing
+    end
+  end
+
+  #decided not to make a test for my helper method `set_plate_type` 
+  #because those tests are relevant and included in my other tests
```
jamisonordway, line 62: Fine for this project, though in the future we will ask you to test all helper methods. What you have covering the `set_plate_type` behavior are **integration** tests; testing `set_plate_type` in isolation would be a **unit** test.

spec/facility_spec.rb
```diff
@@ -2,25 +2,266 @@
...
+      @facility_2.register_vehicle(@bolt)
+      expect(@facility_2.register_vehicle(@bolt)).to eq(nil)
+      expect(@facility_2.registered_vehicles).to eq([])
+      expect(@facility_2.collected_fees).to eq(0)
+      # expect(facility_2.registered_vehicles).to eq([])
+      # expect(facility_2.collected_fees).to eq(0)
+    end
+  end
+#written test
+  describe '#written test' do
```
jamisonordway, line 132: Great use of `describe` to organize groups of test examples. Nice work covering the various scenarios for written/license/renew, as well.

jamisonordway (commented):
Really good job on this project! Please let me know if you have follow up questions about any of my comments.

PR #332, opened 2023-09-05
Michael Kuhlmeier

mbkuhl:


lib/facility_constructor.rb
```diff
@@ -0,0 +1,104 @@
...
+      api_hash[:phone] = api_hash[:public_phone_number]
+    else
+      api_hash[:phone] = "(800) 698-2931"
+    end
+    api_hash
+  end
+
+  def ny_hours(api_hash)
+    hours_string = ""
+    if api_hash[:monday_beginning_hours] != nil
```
jamisonordway, line 55: I noticed that every if statement in this method uses _almost_ the same condition, with the days of the week changing. I think you could refactor this with string interpolation and iteration.
```
def ny_hours(api_hash)
  hours_string = ""
  weekdays = ["Monday", "Tuesday", "Wednesday", "Thursday", "Friday"]
  weekdays.each do |weekday|
    beginning_hours = api_hash["#{weekday.downcase}_beginning_hours"]
    ending_hours = api_hash["#{weekday.downcase}_ending_hours"]
    if beginning_hours != nil
      hours_string += "#{weekday} - #{beginning_hours}-#{ending_hours}"
    end
  end
  hours_string
end
```

spec/analytics_dept_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+    it 'can find the most popular model' do
+      @analytics_dept.get_data
+      expect(@analytics_dept.find_popular(:model)).to be_an_instance_of(String)
+    end
+
+    it 'will not run if there is no data' do
+      expect(@analytics_dept.find_popular(:make)).to be false
+      expect(@analytics_dept.find_popular(:model)).to be false
+    end
```
jamisonordway, line 41: This is a good test. As a thought experiment, how could you use mocks and stubs to make the test more realistic? 
For example, what if this test included `@analytics_dept.get_data`, and it was somehow set up to return nil?

spec/facility_spec.rb
```diff
@@ -23,4 +30,218 @@
       expect(@facility.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
     end
   end
+
+  describe '#register_vehicle' do
```
jamisonordway, line 34: Lots of scenarios to cover in these tests, and this is super thorough 👍🏻

jamisonordway (commented):
Awesome work. Most of my comments are refactoring suggestions; for formatting, my main advice would be to keep an eye on your indentation and line length (80 characters max is the typical convention). Please let me know if you have any follow up questions about this feedback.

PR #333, opened 2023-09-05
Eric Belongea

EricBelongea:


spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+      expect(@facility_factory.locations_ny.first.phone).to eq("(716) 858-7450")
+    end
+
+    it 'impliments phone number helper' do
+      @facility_factory.create_locations(@new_york_facilities)
+      
+      expect(@facility_factory.locations_ny.first.phone).to eq("(716) 858-7450")  
+    end
+  end
+end
```
abdulredd, line 79: **Facility Factory, and if it's okay not to use the before(:each) method in testing?**

Yes absolutely, `before(:each)` is just a mechanism we can use so we don't have to duplicate method setup. You do not have to use it. Quite frankly, I believe the use of it here makes a lot of sense to me.

PR #334, opened 2023-09-05
Kam Kennedy

kameronk92:


lib/facility_factory.rb
```diff
@@ -0,0 +1,65 @@
...
+        name:facility[:name],
+        address:"#{facility[:address1]}, #{facility[:city]}, #{facility[:state]}, #{facility[:zipcode]}",
+        phone:facility[:phone],
+        services:[],
+        collected_fees:0,
+        registered_vehicles:[]}
+      facilities << Facility.new(facility_data)
+    end
+    facilities
+  end
```
jamisonordway, line 64: In your submission form you asked for feedback on these methods (bringing in data from other states) and suggestions on how to make them leaner. A few ideas/foods for thought:
- is there an enumerable that would eliminate the need for lines 52 and 63?
- I don't feel this is a top priority refactor, but have you considered abstracting the contents of the block to an additional helper method? I'd imagine that way this method might look something like
```
def create_mo_facilities
  facility_datas = DmvDataService.new.mo_dmv_office_locations
  facility_datas.map do |facility_data|
    mo_facility(facility_data)
  end
end
```
That way the helper method could have a single responsibility: making a facility instance from a subset of the MO data.
```
def mo_facility(data)
  Facility.new({
    name: data[:name],
    ...
  })
end
```

lib/registrant.rb
```diff
@@ -0,0 +1,25 @@
+class Registrant 
+  attr_reader :name,
+              :age,
+              :permit
+  attr_accessor :license_data
+
+  def initialize(name, age, permit = false, license_data ={:written=>false, :license=>false, :renewed=>false})
```
jamisonordway, line 7: Great use of default arguments! Since this is getting to be on the longer side, I think it would also be fine to instantiate every registrant with false values for the `license_data` hash:
```
  ...
  @license_data = {
    written: false,
    license: false,
    renewed: false
  }
end
```

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  describe '#create_facilities(state_abbv)' do
+    it 'creates DMVs specific to the specified state' do
+      factory = FacilityFactory.new
+      facilities = factory.create_facilities("CO")
+      facilities.each do |facility|
+      expect(facility).to be_an_instance_of(Facility)
+      expect(facility.address).to include("CO")
+      expect(factory.create_facilities("HI")).to eq("sorry, we don't yet serve HI")
```
jamisonordway, line 18: I think this line should be outside of the `do`/`end` block, the assertion doesn't need to run for every `facility` from line 14.

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+      end
```
jamisonordway, line 19: (This applies to several `do`/`end` blocks in your submission)
Be sure to keep an eye on your indentation:
```
facilities.each do |facility|
  expect(facility).to be_an_instance_of(Facility)
  ...
end
```

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  describe '#create_mo_facilities' do
+    it 'creates Missouri DMVs with relavent data' do
+      factory = FacilityFactory.new
+      facilities = factory.create_mo_facilities
+      facilities.each do |facility|
+      expect(facility).to be_an_instance_of(Facility)
+      expect(facility.address).to_not eq(nil)
+      end
```
jamisonordway, line 52: (This applies to multiple tests/examples)
Great job using iteration in your tests to verify that every element is the correct type of object. I wonder if there are state-specific values you could assert in these, as well.

spec/facility_spec.rb
```diff
@@ -2,25 +2,150 @@
...
+
+  describe '#register_vehicle' do
+    it 'registers vehicles if capable' do
+      expect(@facility_1.register_vehicle(@bolt)).to eq(nil)
+      @facility_1.add_service("Vehicle Registration")
+      @facility_1.register_vehicle(@cruz)
+      expect(@facility_1.collected_fees).to eq(100)
+      expect(@facility_1.instance_variable_get(:@registered_vehicles).length).to eq(1)
+      expect(@cruz.plate_type).to eq(:regular)
+      expect(@cruz.registration_date).to eq(Date.today)
```
jamisonordway, line 43: I really like that you tested the return value of `register_vehicle` before adding the appropriate service and testing it again- this clearly demonstrates that calling the `add_service` method makes the intended change.

spec/facility_spec.rb
```diff
@@ -2,25 +2,150 @@
...
+      expect(@facility_1.collected_fees).to eq(325)
+    end
+
+    it 'stores registered vehicles' do
+      @facility_1.add_service('Vehicle Registration')
+      @facility_1.register_vehicle(@cruz)
+      @facility_1.register_vehicle(@camaro)
+      @facility_1.register_vehicle(@bolt)
+      expect(@facility_1.instance_variable_get(:@registered_vehicles).length).to eq(3)
+      expect(@facility_1.instance_variable_get(:@registered_vehicles)).to all be_a(Vehicle)
```
jamisonordway, line 73: `instance_variable_get` is great for testing ivars without exposing them to outside classes with something like an `attr_reader`. Since you do have an `attr_reader` in the `Facility` class for these, I think you could just call those attributes as methods in the test
```
expect(@facility_1.registered_vehicles).to all be_a(Vehicle)
```

spec/registrant_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  describe '#unearn_permit' do
+    it 'changes permit status from true to false' do
+      @registrant_2.earn_permit
+      expect(@registrant_2.permit).to eq(true)
+      @registrant_2.unearn_permit
+      expect(@registrant_2.unearn_permit).to eq(false)
+    end
+  end
```
jamisonordway, line 44: This method wasn't required for the project, but since you added it, I'm really glad to see it's tested 👍🏻

jamisonordway (commented):
Great work on this project 🎉  Most of my comments have to do with Ruby conventions or ways you could continue to refactor your code, or make your test assertions more specific. Please let me know if you have any follow up questions!

PR #338, opened 2023-09-05
Blaine Kennedy

bkchilidawg:


lib/dmv_factory.rb
```diff
@@ -0,0 +1,25 @@
...
+    @office_list = api_data.map do |office|
+      @office = {
+       name: "#{office[:dmv_id]} #{office[:dmv_office]}", 
+       address: "#{office[:address_li]} #{office[:address__1]} #{office[:city]} 
+                 #{office[:state]} #{office[:zip]}", 
+       phone: office[:phone], 
+       services: office[:services_p]
+      }
+    end
+  end
```
jamisonordway, line 17: This works, and the rest of my comment is just something to think about:
Generally my advice is to avoid declaring/assigning instance variables outside of the `initialize` method when possible. Since this method is returning the value of `api_data.map...`, I wonder if you could access this collection by calling on the method directly instead.

lib/dmv_factory.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def create_offices(api_data)
+    self.create_offices_helper(api_data)
+    offices = @office_list.map do |office|
+     facility = Facility.new(office)
+    end
+  end
```
jamisonordway, line 24: 1. The use of `self` within methods is implied in Ruby. 
2. If you were to take my advice in the previous comment about doing away with the `@office_list` ivar, this might look something like
```
def create_offices(api_data)
  office_list = create_offices_helper(api_data)
  office_list.map do |office|
    Facility.new(office)
  end
end
```
Note that I removed the unused variables `offices` and `facility` as well.

lib/facility.rb
```diff
@@ -1,14 +1,62 @@
...
+      vehicle.registration_date = Date.today
+      @collected_fees += 200
+    else
+      vehicle.plate_type = :regular
+      vehicle.registration_date = Date.today
+      @collected_fees += 100
+    end
+    @registered_vehicles << vehicle
+  end
+  def administer_written_test(registrant)
```
jamisonordway, line 37: Be sure to leave a blank line between method ends/definitions.

lib/vehicle.rb
```diff
@@ -5,18 +5,26 @@ class Vehicle
...
     @make = vehicle_details[:make]
     @model = vehicle_details[:model]
     @engine = vehicle_details[:engine]
+    @registration_date = nil
+    @plate_type = nil
   end
 
   def antique?
-    Date.today.year - @year > 25
+    Date.today.year - @year >= 25
```
jamisonordway, line 24: You could make your code more DRY (don't repeat yourself) by reusing the age method:
```
age >= 25
```

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,14 @@
+class VehicleFactory
+  def initialize
+  end
+  
+  def create_vehicles(api_data)
+    @factory_vehicles = []
+    api_data.select do |data|
```
jamisonordway, line 7: `select` is an enumerable best used for only keeping elements when a certain condition about those elements evaluates to true. Since lines 8-10 will always return truthy values (meaning: they don't contain any conditional logic or filter anything out), I might reach for `.map` here instead.

spec/dmv_factory_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      @office_list = @ny_office.create_offices(@ny_dmv_office_locations)
+      expect(@office_list.length).to eq(@ny_dmv_office_locations.length)
+      expect(@office_list.first).to be_a(Facility)
+    end
+    it 'creates Facility Location objects from API data from multiple cities' do
+      @office_list = @mo_office.create_offices(@mo_dmv_office_locations)
+      expect(@office_list.length).to eq(@mo_dmv_office_locations.length)
+      expect(@office_list.first).to be_a(Facility)
+    end
+  end
```
jamisonordway, line 33: You might also want to assert that `@office_list.last` is a `Facility` instance in each of these cases. This would do more to ensure that every element is the right type of object.

spec/dmv_spec.rb
```diff
@@ -3,9 +3,9 @@
 RSpec.describe Dmv do
   before(:each) do
     @dmv = Dmv.new
-    @facility_1 = Facility.new({name: 'DMV Tremont Branch', address: '2855 Tremont Place Suite 118 Denver CO 80205', phone: '(720) 865-4600'})
-    @facility_2 = Facility.new({name: 'DMV Northeast Branch', address: '4685 Peoria Street Suite 101 Denver CO 80239', phone: '(720) 865-4600'})
-    @facility_3 = Facility.new({name: 'DMV Northwest Branch', address: '3698 W. 44th Avenue Denver CO 80211', phone: '(720) 865-4600'})
+    @facility_1 = Facility.new({:name => 'DMV Tremont Branch', :address => '2855 Tremont Place Suite 118 Denver CO 80205', :phone => '(720) 865-4600'})
+    @facility_2 = Facility.new({:name => 'DMV Northeast Branch', :address => '4685 Peoria Street Suite 101 Denver CO 80239', :phone => '(720) 865-4600'})
+    @facility_3 = Facility.new({:name => 'DMV Northwest Branch', :address => '3698 W. 44th Avenue Denver CO 80211', :phone => '(720) 865-4600'})
```
jamisonordway, line 8: The original key/value syntax (`{name: 'DMV Tremont Branch', address: '2285...'}`) is more conventional in newer versions of Ruby, but both will work.

spec/facility_spec.rb
```diff
@@ -1,26 +1,132 @@
...
+      registered_vehicles = @facility_1.register_vehicle(@camaro)
+      registered_vehicles = @facility_1.register_vehicle(@bolt)
+      expect(@cruz.registration_date).to eq(Date.today) 
+      expect(@cruz.plate_type).to eq(:regular)
+      expect(@camaro.registration_date).to eq(Date.today)
+      expect(@camaro.plate_type).to eq(:antique)
+      expect(@bolt.registration_date).to eq(Date.today)
+      expect(@bolt.plate_type).to eq(:ev)
+      expect(@facility_1.registered_vehicles).to include(@cruz, @camaro, @bolt)
+      expect(@facility_1.collected_fees).to eq 325
```
jamisonordway, line 57: Great to see an _integration_ test, asserting that these values are updated across multiple vehicle registrations.

spec/facility_spec.rb
```diff
@@ -1,26 +1,132 @@
...
+    end
+  end
+
+  describe '#administer_written_test' do
```
jamisonordway, line 61: This should also test that a written test isn't administered if the registrant is under 16 years of age or does not have a permit.

spec/facility_spec.rb
```diff
@@ -1,26 +1,132 @@
...
+
+    it 'fails written test if no service is available' do
+      facility_1 = Facility.new({name: 'DMV Tremont Branch', address: '2855 Tremont Place Suite 118 Denver CO 80205', phone: '(720) 865-4600'})
+      registrant_1 = Registrant.new('Bruce', 18, true)
+      expect(facility_1.administer_written_test(registrant_1)).to be false
+      expect(registrant_1.license_data[:written]).to be false
+    end
+  end
+
+  describe '#administer_road_test' do
```
jamisonordway, line 79: This should also test that a road test is not administered if the registrant has not completed their written test.

spec/facility_spec.rb
```diff
@@ -1,26 +1,132 @@
...
+
+    it 'fails road test if no service is available' do
+      facility_1 = Facility.new({name: 'DMV Tremont Branch', address: '2855 Tremont Place Suite 118 Denver CO 80205', phone: '(720) 865-4600'})
+      registrant_1 = Registrant.new('Bruce', 18, true)
+      expect(facility_1.administer_road_test(registrant_1)).to be false
+      expect(registrant_1.license_data[:license]).to be false
+    end
+  end
+
+  describe '#renew_drivers_license' do
```
jamisonordway, line 99: This should also test that a drivers license is not renewed for an unqualified registrant.

jamisonordway (commented):
Good work! I left you some specific feedback about ways to make your testing more robust, and suggested a few refactors in your implementation code. Please let me know if you have any feedback about these comments.

PR #339, opened 2023-09-05
Martin Chavez

Chavezgm:


lib/facility.rb
```diff
@@ -1,14 +1,63 @@
...
+
+  def renew_drivers_license(registrant)
+    #require 'pry';binding.pry
+    if services.include?("Renew License") && registrant.permit? && registrant.license_data[:written] == true && registrant.license_data[:license] == true
+      registrant.license_data[:renewed] = true
+ 
+    end 
+    registrant.license_data[:renewed]
+  end
+end
```
abdulredd, line 61: Overall this is a well-put-together class. Just a few suggestions:

There are several places within this class where you are doing `== true`. This is not required and you can simplify this condition by removing this unnecessary check, (ie. `&& registrant.permit? &&`, `registrant.license_data[:written]`)

There also seems to be some inconsistency in coding style throughout this class. On line 40 you explicity return `false` but on lines 50 and 59 you don't and instead return the value in the `license_data` hash. The `administer_written_test` method is also the only one that uses an `else` where the others do not.

My point is, no matter how small or large the codebase, the code writing should be consistent and it should look like 1 person wrote it, even in group work.

lib/facility_factory.rb
```diff
@@ -0,0 +1,29 @@
...
+        address = "#{data[:address1]} #{data[:city]} #{data[:state]} #{data[:zip]}"
+        phone = data[:phone]
+      end 
+      facility = Facility.new(name: name, address: address, phone: phone)
+      facilities << facility
+    end 
+    facilities
+    
+  end 
+end
```
abdulredd, line 29: The code in this class is good, there are just a few spacing and indentation issues. Consider using consistent indentation and spacing to improve readability.

Finally, while this logic is okay in terms of determining which type of Facility to create, I also believe it is quite fragile. I think the best way to determine what type of Facility to create would have been by using the state data like so:

```ruby
def create_facility(api_data) # This method should also be named plurally
  facilities = []

  api_data.each do |data|
    state = data[:state]

    if state == "CO"
      # Parse the CO data
    elsif state == "NY"
      # Parse the NY data
    elsif state == "MO"
      # Parse the MO data
    else
      # Do something else...
    end
  
    facility = Facility.new(name: name, address: address, phone: phone)
    facilities << facility
  end

  facilities
end
```

This small change makes your code more resistant to breaking since we're relying on the value of the state and not some arbitrary key to not be `nil`.

Also, because we are creating new Facility objects in this class, you need to require the Facility class at the top of this file.

lib/registrant.rb
```diff
@@ -0,0 +1,22 @@
...
+    end 
+
+    def earn_permit
+        @permit = true 
+    end
+    
+    def unearn_permit
+        @permit = false
+    end 
+end
```
abdulredd, line 22: 👍 

Spacing and indentation please. 2 spaces, not 4.

lib/vehicle.rb
```diff
@@ -22,4 +28,5 @@ def antique?
   def electric_vehicle?
     @engine == :ev
   end
-end
+  
+end
```
abdulredd, line 32: 👍

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,16 @@
...
+      model = registration[:model]
+      vin = registration[:vin_1_10]
+      model_year = registration[:model_year]
+
+      vehicle = Vehicle.new(engine: , make: , model:, vin: , model_year: )
+      vehicles << vehicle
+    end 
+    vehicles
+  end 
+end
```
abdulredd, line 16: In the loop, you have assigned values to `engine`, `make`, `model`, `vin`, and `model_year`, but you haven't provided values for these attributes when creating a `Vehicle` object. Ensure that you pass the appropriate values to `Vehicle.new` for each attribute.

Also, since we are creating new Vehicle objects in this class, you must also require the Vehicle class at the top of this file.

abdulredd (commented):
.

PR #340, opened 2023-09-05
Sooyung Kim

skim1027:


lib/dmv.rb
```diff
@@ -1,16 +1,51 @@
...
+            update_phone = phone_data.insert(0, "(").insert(4, ")").insert(5, ' ').insert(-5, '-')
+          end
+        facility_details = { name: office_name, address: full_address, phone: update_phone }
+      else
+        address_data = office[:address1], office[:city], office[:state], office[:zipcode]
+        full_address = address_data.join(' ')
+        facility_details = { name: office[:name], address: full_address, phone: office[:phone] }
+      end
+    end
+  end
```
jamisonordway, line 44: I agree with what you said in your submission form about this method, there's some room to refactor as it got pretty large and complex. Some food for thought:
1. It doesn't seem like the local variable `facility_details` gets used again. Lines 23, 37, and 41 could just return a hash without assigning it to a var.
2. You could _abstract_ some of this functionality to helper methods and call on them according to the value of `office[:state]`.
```
if office[:state] == 'CO'
  co_office_data(office)
elsif office[:state] == 'NY'
  ny_office_data(office)
else
  mo_office_data(office)
end
```
```
# using CO functionality as an example
def co_office_data(office)
  address_data = office[:address_li], office[:address__1], office[:city], office[:state], office[:zip]
  full_address = address_data.join(" ")
  { name: office[:dmv_office], address: full_address, phone: office[:phone] }
end
```
3. I also think lines 34-36 could be moved to a helper method, making 33 something like:
```
phone_data = updated_phone(office[:public_phone_number])
```

lib/facility.rb
```diff
@@ -1,14 +1,75 @@
...
   end
-
-  def add_services(service)
+  
+  def add_service(service)
     @services << service
   end
+
+  def collected_fees
+    @collected_fees = 0
```
jamisonordway, line 21: Since this variable isn't used outside of the `collected_fees` method, it should be a local variable, not an ivar. 
A different approach would be to store `@collected_fees` as an attribute in the initialize, and only do a `@collected_fees +=` at the time of vehicle registration. Your solution works well, just wanted to mention an alternative since we're on the subject of instance variables vs local variables.

lib/facility.rb
```diff
@@ -1,14 +1,75 @@
...
+      @registered_vehicles << car
+    end
+    car.registration_date = Date.today
+      if car.antique? == true
+        car.plate_type = :antique
+      elsif car.engine == :ev
+        car.plate_type = :ev
+      else
+        car.plate_type = :regular
+      end
```
jamisonordway, line 45: Lines 38-45 should probably be nested under the line 35 conditional. Otherwise, these attributes will be updated for the car, even if the facility should not provide registration services.
To avoid nesting conditional statements, you could use a guard clause to exit the method early if the service is not available:
```
def register_vehicle(car)
  @registered_vehicles << car
  car.registration_date = Date.today
  if car.antique?
    ...
  end
end
```

lib/facility.rb
```diff
@@ -1,14 +1,75 @@
...
+  end
+
+  def administer_written_test(registrant)
+    if @services.include?('Written Test') 
+      if registrant.age >= 16 && registrant.permit? 
+        registrant.license_data[:written] = true
+      end
+    end 
+    registrant.license_data[:written]
+  end
```
jamisonordway, line 55: For this method and the similar ones below, you could refactor to 'flatten' the conditional logic:
```
if @services.include?('Written Test') && registrant.age >= 16 && registrant.permit?
end
registrant.license_data[:written]
```
Of course, that line gets long. Maybe it warrants a helper method.
```
def written_eligible?(registrant)
  registrant.age >= 16 && registrant.permit?
end

  if @services.include?('Written Test') && written_eligible?(registrant)
    ...
  end
  registrant.license_data[:written]
end
```

lib/registrant.rb
```diff
@@ -0,0 +1,28 @@
+class Registrant
+  attr_reader :name, :age, :permit, :earn_permit
```
jamisonordway, line 2: `earn_permit` is a method, not an instance variable/attribute, so it doesn't need an `attr_reader`.

lib/registrant.rb
```diff
@@ -0,0 +1,28 @@
...
+    @permit = permit
+    @license_data = {
+      :written=>false, 
+      :license=>false, 
+      :renewed=>false
+    }
+  end
+
+  def permit?
+    if @permit == true || @earn_permit == true
```
jamisonordway, line 18: I'm not sure this line of code is working the way you intended. The ivar `@earn_permit` will always return `nil` because it isn't defined anywhere. 
Digging a little deeper on this logic, we don't need to consider the `earn_permit` method in this conditional statement at all. This is because once the `earn_permit` method is called, it will update the `@permit` attribute, which is what's being checked here anyway.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,37 @@
...
+  def car_model
+    list_model = @create_vehicle.map do |vehicle|
+      vehicle.model
+    end
+    counts = Hash.new 0
+    list_model.each do |model|
+      counts[model] += 1
+    end
+    counts.max_by{ |model, value| value }
+  end
```
jamisonordway, line 36: This is a great use of the map and max_by enumerables!

spec/dmv_spec.rb
```diff
@@ -32,12 +36,66 @@
...
+      expect(ny_facility_info.class).to eq(Array)
+      expect(ny_facility_info.first).to eq({name: 'EVANS COUNTY OFFICE', address: '6853 ERIE RD DERBY NY 14006', phone: '(716) 858-7450'})
+    end
+
+    it 'converts MO facility information from API correct format' do
+      mo_facility_info = @dmv.create_facility_helper(@mo_dmv_office_locations)
+      expect(mo_facility_info.class).to eq(Array)
+      expect(mo_facility_info.first).to eq({name: 'OAKVILLE', address: '3164 TELEGRAPH ROAD ST LOUIS MO 63125', phone: '(314) 887-1050'})
+    end
+  end
```
jamisonordway, line 66: These are great tests. One suggestion I have is to verify (in each example) that _every_ element of the returned array has the expected data types:
```
mo_facility_info = @dmv.create_facility_helper(@mo_dmv_office_locations)
mo_facility_info.each do |facility|
  expect(facility).to be_a(Hash)
  expect(facility[:name]).to be_a(String)
  # etc
end
```

spec/dmv_spec.rb
```diff
@@ -32,12 +36,66 @@
...
+    it 'converts mo facility information from API' do
+      @dmv.create_facility_helper(@mo_dmv_office_locations)
+      expect(@dmv.create_facility.count).to eq(178)
+      expect(@dmv.create_facility.class).to eq(Array)
+      expect(@dmv.create_facility.first.class).to eq(Facility)
+      expect(@dmv.create_facility.first.name).to eq('OAKVILLE')
+      expect(@dmv.create_facility.first.phone).to eq('(314) 887-1050')
+      expect(@dmv.create_facility.first.address).to eq('3164 TELEGRAPH ROAD ST LOUIS MO 63125')
+    end
+  end
```
jamisonordway, line 98: Again, great tests, and you could also make them more robust by iterating within the test examples.
```
@dmv.create_facility_helper(@mo_dmv_office_locations)
facility_objects = @dmv.create_facility
facility_objects.each do |facility|
  expect(facility).to be_a(Facility)
  expect(facility.name).to be_a(String)
end
```
Alternatively, you could assert the specific attributes of `@dmv.create_facility.last` in addition to the first ones.

spec/facility_spec.rb
```diff
@@ -2,25 +2,217 @@
...
+      @facility_1.add_service('Road Test')
+      @facility_1.add_service('Renew License')
+      @registrant_2.earn_permit
+      @facility_1.administer_written_test(@registrant_2)
+      @facility_1.administer_road_test(@registrant_2)
+      @facility_1.renew_drivers_license(@registrant_2)
+      expect(@facility_1.renew_drivers_license(@registrant_2)).to be true
+      expect(@registrant_2.license_data).to eq({:written=>true, :license=>true, :renewed=>true})
+    end
+  end
```
jamisonordway, line 216: Really nice job testing various scenarios regarding registrant and facility eligibility for these services.

jamisonordway (commented):
Great job on this project! I left some comments, mostly ideas about how you could refactor complex methods and add even more value to your test suite. Please let me know if you have any follow up questions about this feedback.

PR #341, opened 2023-09-05
Miranda Delaney

delaneymiranda1:


lib/dmv_factory.rb
```diff
@@ -0,0 +1,24 @@
+class DmvFactory 
+  attr_reader :office_list
+  def initialize
+    @office = nil
+  end
```
abdulredd, line 5: The instance variable `@office` is not being used anywhere in the code and could probably be removed.

lib/dmv_factory.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def create_offices_helper(api_data)
+    @office_list = api_data.map do |office|
+     office_data = { 
+      name: "#{office[:dmv_id]} #{office[:dmv_office]}", 
+      address: "#{office[:address_li]} #{office[:address__1]} 
+                #{office[:city]} #{office[:state]} #{office[:zip]}", 
+      phone: office[:phone], services: office[:services_p] } 
+    end
+  end
```
abdulredd, line 15: I think this method could use some additional formatting to make it easier to read.

```ruby
def create_offices_helper(api_data)
  @office_list = api_data.map do |office|
    office_data = { 
      name: "#{office[:dmv_id]} #{office[:dmv_office]}", 
      address: "#{office[:address_li]} #{office[:address__1]} #{office[:city]} #{office[:state]} #{office[:zip]}", 
      phone: office[:phone], 
      services: office[:services_p] 
    } 
  end
end
```

You also do not need to be grabbing `services` data from the hash since a new Facility object will start with no services.

lib/dmv_factory.rb
```diff
@@ -0,0 +1,24 @@
...
+
+
+  def create_offices(api_data)
+    self.create_offices_helper(api_data)
+    offices = @office_list.map do |office|
+      facility = Facility.new(office)
+    end
+  end
```
abdulredd, line 23: I don't think you need both the `create_offices_helper` and `create_offices` method. They could both be combined. This would make the code a lot easier to read.

```ruby
def create_offices_helper(api_data)
  api_data.map do |office|
    office_data = { 
      name: "#{office[:dmv_id]} #{office[:dmv_office]}", 
      address: "#{office[:address_li]} #{office[:address__1]} #{office[:city]} #{office[:state]} #{office[:zip]}", 
      phone: office[:phone], 
      services: office[:services_p] 
    } 

    Facility.new(office_data)
  end
end
```

You also don't need to store the result of the enumerable into a variable like you are doing on line 8 and line 20.

lib/facility.rb
```diff
@@ -1,14 +1,65 @@
 class Facility
-  attr_reader :name, :address, :phone, :services
+  attr_accessor :name, 
+              :address, 
+              :phone, 
+              :services,
+              :registered_vehicles, 
+              :collected_fees
```
abdulredd, line 7: We should only be using `attr_accessor` in instances where we intend for an instance variable to be read AND overwritten from outside of the class. None of the instance variables meet this criterion and only need to be read from outside the class, therefore we should be using an `attr_reader` instead.

lib/facility.rb
```diff
@@ -1,14 +1,65 @@
...
+        vehicle.plate_type = :ev
+        @collected_fees += 200
+      else
+        vehicle.plate_type = :regular 
+        @collected_fees += 100
+      end
+      vehicle.registration_date = Date.today
+      @registered_vehicles << vehicle
+    end
+  end
```
abdulredd, line 37: In this method, I think we are missing the check to see if this Facility even offers vehicle registrations before performing it.

Be careful when using a `while` loop as it may lead to an infinite loop. Since the `vehicle` argument that we are passing in will only ever be a single vehicle and not an array or something else that needs to be iterated over in a loop I think a normal `if` statement is the better choice in this scenario.

lib/facility.rb
```diff
@@ -1,14 +1,65 @@
...
+ 
+  def administer_road_test(registrant)
+    if @services.include?('Road Test') and registrant.license_data[:written] = true
+      registrant.license_data[:license] = true
+      true
+    else 
+      false
+    end
+  end
```
abdulredd, line 55: I think `registrant.license_data[:written] = true` is missing another `=` to make it a conditional statement. Currently, it is a variable assignment and doesn't belong in an `if` statement.

lib/facility.rb
```diff
@@ -1,14 +1,65 @@
...
+
+  def renew_drivers_license(registrant)
+    if @services.include?('Renew License') and registrant.license_data[:license] == true
+      registrant.license_data[:renewed] = true
+      true
+    else
+      false
+    end
+  end
```
abdulredd, line 64: You can omit the explicit `== true` check when checking for boolean values and just simplify it to this:

```ruby
if @services.include?('Renew License') and registrant.license_data[:license]
```

`registrant.license_data[:license]` will either be `true` or `false`.

lib/facility.rb
```diff
@@ -1,14 +1,65 @@
...
 end
```
abdulredd, line 65: Although in Ruby `and` and `&&` seem very similar, we recommend sticking to `&&` to avoid any surprises.

lib/registrant.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def permit?
+    @permit
+  end
+
+  def earn_permit
+    @permit = true
+  end
+end
```
abdulredd, line 22: This class is great. The only thing I will add is that you've defined an `attr_accessor` for `@license_data`, which allows direct modification of the license_data hash from outside the class. This is only necessary if you want to allow `@license_data` to point to something other than the hash it is currently pointing to.

Because `@license_data` is pointing to a hash and hashes can be updated without the `attr_writer`, line 5 can be removed and you would need to add `:license_data` to the `attr_reader` list.

lib/vehicle.rb
```diff
@@ -1,18 +1,22 @@
...
-              :make,
-              :model,
-              :engine
+  attr_reader :vin, 
+              :year, 
+              :make, 
+              :model, 
+              :engine, 
+              :registration_date
+  attr_accessor :plate_type, :registration_date
```
abdulredd, line 10: It's redundant to specify both `attr_reader` and `attr_accessor` for the same attribute, `registration_date`. If you want it to be both readable and writable, you can use `attr_accessor` alone.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,17 @@
+class VehicleFactory
+  def initialize
+   
+  end
```
abdulredd, line 4: The `initialize` method is currently empty. If you don't have any specific initialization logic to perform, you can omit this method. It is not required to have an `initialize` method unless there is logic you want to put inside of it. An instance object can still be instantiated without explicitly defining it.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,17 @@
...
+    @factory_vehicles = []
+    api_data.each do |data|
+      vehicle_data = { vin: data[:vin_1_10], make: data[:make], 
+                      model: data[:model], year: data[:model_year], 
+                      engine: 'ev' }
+      vehicle = Vehicle.new(vehicle_data)
+      @factory_vehicles << vehicle
+    end
+    @factory_vehicles
+  end
```
abdulredd, line 16: When declaring a hash, it is much easier to read when all the keys are on their own line.

```ruby
def create_vehicles(api_data)
  @factory_vehicles = []
  api_data.each do |data|
    vehicle_data = {
      vin: data[:vin_1_10], 
      make: data[:make], 
      model: data[:model], 
      year: data[:model_year], 
      engine: 'ev' # This is supposed to be :ev, not 'ev'
    }
    vehicle = Vehicle.new(vehicle_data)
    @factory_vehicles << vehicle
  end
  @factory_vehicles
end
```

Finally, instead of manually creating an array, `@factory_vehicles`, to store the vehicles and then returning it at the end of the method, you can simplify the code by using the `map` method to transform the `api_data` into an array of `Vehicle` objects and directly return it:

```ruby
def create_vehicles(api_data)
  api_data.map do |data|
    vehicle_data = {
      vin: data[:vin_1_10], 
      make: data[:make], 
      model: data[:model], 
      year: data[:model_year], 
      engine: 'ev' 
    }
    
    Vehicle.new(vehicle_data)
  end
end
```

abdulredd (commented):
I know you asked for specific feedback on your Facility tests and I think they are great. If there was something in particular you were unsure about please let me know but I think what you have is great.

PR #343, opened 2023-09-05
Billy Burch

billyburch:
Iteration 3 should be complete! All spec tests in the directory came back as passing. Thank you.

### DIFF
--- a/lib/facility_creator.rb
+++ b/lib/facility_creator.rb
@@ -1,0 +1,22 @@
+require 'date'
+
+class FacilityCreator
+  def initialize
+    @new_facilities = []
+  end
+
+  def create_facilities(remote_data)
+    remote_data.each do |data|
+      facility = Facility.new{
+        name:data[:dmv_office],
+        address:"#{data[:address_li]}, #{data[:address__1]}, #{data[:city]}, #{data[:state]}, #{data[:zip]}",
+        phone:data[:phone],
+        services:data[:services_p],
+        collected_fees:0,
+        registered_vehicles:[]}
+      @new_facilities << facility
+    end
+    @new_facilities
+  end
+end
+
--- a/lib/facility_creator.rb
+++ b/lib/facility_creator.rb
@@ -5,15 +5,43 @@ class FacilityCreator
     @new_facilities = []
   end
 
-  def create_facilities(remote_data)
-    remote_data.each do |data|
-      facility = Facility.new{
+  def create_co_facilities(co_remote_data)
+    co_remote_data.each do |data|
+      facility = Facility.new({
         name:data[:dmv_office],
         address:"#{data[:address_li]}, #{data[:address__1]}, #{data[:city]}, #{data[:state]}, #{data[:zip]}",
         phone:data[:phone],
         services:data[:services_p],
         collected_fees:0,
-        registered_vehicles:[]}
+        registered_vehicles:[]})
+      @new_facilities << facility
+    end
+    @new_facilities
+  end
+
+  def create_ny_facilities(ny_remote_data)
+    ny_remote_data.each do |data|
+      facility = Facility.new({
+        name:data[:office_name],
+        address:"#{data[:street_address_line_1]}, #{data[:city]}, #{data[:state]}, #{data[:zip_code]}",
+        phone:data[:public_phone_number],
+        services:[],
+        collected_fees:0,
+        registered_vehicles:[]})
+      @new_facilities << facility
+    end
+    @new_facilities
+  end
+
+  def create_mo_facilities(mo_remote_data)
+    mo_remote_data.each do |data|
+      facility = Facility.new({
+        name:data[:name],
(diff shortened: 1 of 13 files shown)


lib/facility_creator.rb
```diff
@@ -0,0 +1,50 @@
...
+        address:"#{data[:address1]}, #{data[:city]}, #{data[:state]}, #{data[:zipcode]}",
+        phone:data[:phone],
+        services:[],
+        collected_fees:0,
+        registered_vehicles:[]})
+      @new_facilities << facility
+    end
+    @new_facilities
+  end
+end
```
abdulredd, line 49: I think there might have been a little confusion surrounding the expectation for iteration 3. It was intended for you to just have one method that would behave differently depending on the data that it got sent from either the CO, NY or MO endpoints. Currently, you have 3 different methods within this class.

The code overall is great and each method does what it is supposed to do but doesn't quite follow the specification.

One last thing, since we are instantiating new `Facility` objects we should `require` in the Facility class into this file.

lib/registrant.rb
```diff
@@ -0,0 +1,18 @@
+class Registrant
+  attr_reader :name, :age, :permit
+  attr_accessor :license_data
+  def initialize(name, age, permit = false, license_data = {:written=>false, :license=>false, :renewed=>false})
+    @name = name
+    @age = age
+    @permit = permit
+    @license_data = license_data
+  end
```
abdulredd, line 9: I think one common thing I see with new developers is that there is some confusion surrounding when something should be set as a default parameter vs a default value in the `initialize` method. One way to remember this is to think about how many potential arguments are going to be passed to the method in question. 

In this example, there will never be an instance where license_data will be set to anything other than `{ :written => false, :license => false, :renewed => false }`, never. And since this is the case this should be set at the default within the method.

This is different than `permit` since there will be instances where a new `Registrant` will be instantiated with a value of `true` as the 3rd optional argument.

```ruby
  @name = name
  @age = age
  @permit = permit
  @license_data = { :written => false, :license => false, :renewed => false }
end
```

One other thing, since you are using Symbols as keys, you do not need the hash rocket.

```ruby
@license_data = { written: false, license: false, renewed: false }
```

Lastly, you've defined an `attr_writer` for `@license_data`, which allows direct modification of the license_data hash from outside the class. This is only necessary if you want to allow `@license_data` to point to something other than the hash it is currently pointing to.

Because `@license_data` is pointing to a hash and hashes can be updated without the `attr_writer`, line 3 can be removed.

lib/vehicle.rb
```diff
@@ -5,14 +5,20 @@ class Vehicle
               :year,
               :make,
               :model,
-              :engine
+              :engine,
+              :registration_date,
+              :plate_type
+              
+  attr_accessor :registration_date, :plate_type
```
abdulredd, line 12: It's redundant to specify both `attr_reader` and `attr_accessor` for the same attributes (`registration_date` and `plate_type`). If you want them to be both readable and writable, you can use `attr_accessor` alone.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,20 @@
...
+        vin: data[:vin_1_10],
+        year: data[:model_year],
+        make: data[:make],
+        model: data[:model],
+        engine: :ev)
+      @new_vehicles << vehicle
+    end
+    @new_vehicles
+  end
+end
```
abdulredd, line 20: Great job here. I see on line `18` is returning the `@new_vehicles` array as the final operation in the method. You can eliminate the need to do this if you switch out `each` for `map` like so:

```ruby
require "./lib/vehicle"

class VehicleFactory
  def create_vehicles(remote_data)
    remote_data.map do |data|
      Vehicle.new(
        vin: data[:vin_1_10],
        year: data[],
        make: data[:make],
        model: data[:model],
        engine: ev
      )
    end
  end
end
```

1. You'll see I completely got rid of the `initialize` and the `@new_vehicles` array altogether
2. Converted `each` to `map` since `map` transforms every element in an array from one thing to something else, in this case -- `Vehicle` objects
3. Don't forget to require the `Vehicle` class at the top of this file since you are instantiating new vehicles

spec/facility_spec.rb
```diff
@@ -23,4 +24,252 @@
...
   end
+
+  describe '#add facility' do
+    it 'can add facilities' do
+      facility_1 = Facility.new({name: 'DMV Tremont Branch', address: '2855 Tremont Place Suite 118 Denver CO 80205', phone: '(720) 865-4600'})
+      facility_2 = Facility.new({name: 'DMV Northeast Branch', address: '4685 Peoria Street Suite 101 Denver CO 80239', phone: '(720) 865-4600'})
+      expect(facility_1).to be_a Facility
+      expect(facility_2).to be_a Facility
+    end
+  end
```
abdulredd, line 35: I'm a little confused as to what we are testing in this block. The description says that you're testing the ability to add facilities but I don't see that reflected in the block. These tests look exactly the same as the assertion on line `10`.

PR #345, opened 2023-09-05
Nurudeen Sulaimon

nurudeensulaimon:


lib/facility.rb
```diff
@@ -1,14 +1,89 @@
...
     @services << service
   end
+
+  def registered_vehicles 
+    @r_vehicles
+  end 
+
+  def collected_fees
+    @total
+  end
```
megstang, line 28: If you created `attr_readers` for these attributes, you would not need to create these methods. You would, however, need to change the name of your instance variables in your initialize method to be named the way you want the method to be named.

lib/facility.rb
```diff
@@ -1,14 +1,89 @@
...
+      @r_vehicles 
+    elsif vehicle.vehicle_type == :regular
+      @total += 100
+      @r_vehicles 
+     # binding.pry
+    end 
+  else  
+      nil 
+   end 
+  end
```
megstang, line 49: This looks good!

lib/facility.rb
```diff
@@ -1,14 +1,89 @@
...
+    if @services.include?('Written Test')
+      if registrant.age >= 16 && registrant.permit == true
+        registrant.license_data[:written] = true
+      else
+        false
+      end
+    else
+      false
+    end
+  end
```
megstang, line 62: (this comment applies to the two mehtods below this as well)

We can combine some logic to slim up this method. We could refactor this to look something like: 

```ruby
    if @services.include?('Written Test') && registrant.age >= 16 && registrant.permit == true
     else
        false
    end
end
```

lib/facility_factory.rb
```diff
@@ -0,0 +1,59 @@
...
+        @zip = details[:zip_code]
+      else
+        # Missouri Reference
+        @dmv_office = details[:type]
+        @address_li = details[:county]
+        @phone = details[:phone]
+        @address__1 = details[:address1]
+        @city = details[:city]
+        @state = details[:state]
+        @zip = details[:zipcode]
```
megstang, line 53: I'm curious why you have your `DMV` class in this file? I don't believe this is necessary.  You already have a DMV class defined in '/lib/dmv.rb'.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,30 @@
...
+      Vehicle.new({
+        vin: vehicle_data[:vin_1_10],
+        year: vehicle_data[:model_year],
+        make: vehicle_data[:make],
+        model: vehicle_data[:model],
+        engine: :ev
+      })
+    end
+  end
+end
```
megstang, line 30: You do not need any of these attr_accessors or attr_readers. The interaction pattern mentioned that you just need a `create_vehicles` method that returns a collection of vehicle objects. We don't need any other methods to return something else in this class. So, we can slim this class up to look something like: 

```ruby 
class VehicleFactory 

  def create_vehicles(details)
    details.map do |vehicle_data|
      Vehicle.new({
        vin: vehicle_data[:vin_1_10],
        year: vehicle_data[:model_year],
        make: vehicle_data[:make],
        model: vehicle_data[:model],
        engine: :ev
      })
    end
  end
end
```

megstang (commented):
@nurudeensulaimon Good job on this project. I left some comments for you to take a look over. I'll follow up in slack within a day or so with more feedback. Let me know if you have any questions!

PR #347, opened 2023-09-05
Shawn Carpenter

shawndcarpenter:


lib/facility_factory.rb
```diff
@@ -0,0 +1,42 @@
...
+        new_facility = Facility.new({
+          name: item[:name], 
+          address: "#{item[:address1]}, #{item[:city]}, #{item[:state]} #{item[:zipcode]}", 
+          phone: item[:phone],
+          hours: item[:daysopen],
+          closed: item[:holidaysclosed]})
+        facilities << new_facility
+      end
+    end
+  end
```
abdulredd, line 39: I think first and foremost, the `create_facilities` method is quite long. I would encourage you to keep your methods somewhere between 7 - 10 lines in length and refactor into helper methods to make the code more readable and easier to understand.

```ruby
def create_facilities(items)
  items.map do |item|
    state = item[:state]
    
    case state
    when "CO"
      create_co_facility(item)
    when "NY"
      create_ny_facility(item)
    when "MO"
      create_mo_facility(item)
    end
    
  end
end
```

I would also replace `items` with `dataset` or something similar, and `item` to `data` or something similar.

The last thing I want to mention is that I see that you are using a `map` but also shoveling the `new_facility` into a `facilities` array. This defeats the purpose of using a `map` and these lines can be removed since no shoveling needs to take place. A `map` will transform every element in the array from one thing to something else, which in this example are `Facility` objects.

lib/registrant.rb
```diff
@@ -0,0 +1,24 @@
...
+  attr_reader :name, :age, :permit, :license_data
+  attr_accessor :license_data
+
+  def initialize(name, age, permit = false)
+    @name = name
+    @age = age
+    @permit = permit
+    @license_data = license_data
+    assign_license_data
+  end
```
abdulredd, line 11: In the `initialize` method, you have `@license_data = license_data`, which is setting the instance variable `@license_data` to itself and not using the default values provided by `assign_license_data`. You should remove this line and initialize `@license_data` by assigning it to the hash directly.

```ruby
@license_data = { written: false, license: false, renewed: false }
```

The `assign_license_data` method can be removed.

Lastly, you've defined an `attr_writer` for `@license_data`, which allows direct modification of the license_data hash from outside the class. This is only necessary if you want to allow `@license_data` to point to something other than the hash it is currently pointing to.

Because `@license_data` is pointing to a hash and hashes can be updated without the `attr_writer`, line 3 can be removed.

lib/vehicle.rb
```diff
@@ -1,18 +1,18 @@
 require 'date'
 
 class Vehicle
-  attr_reader :vin,
-              :year,
-              :make,
-              :model,
-              :engine
+  attr_reader :vin, :year, :make, :model, :engine, :registration_date, :county
+  attr_accessor :registration_date, :plate_type, :vin, :year, :make, :model, :engine, :county
```
abdulredd, line 5: Similar to what I mentioned earlier, having the same instance variables in the `attr_reader` and `attr_accessor` is redundant. If you intent to have reading and writing of an instance variable then only an `attr_writer` is necessary.

I am also unsure of what we need to create the ability to write to the instance variables from outside of this class, since the only variables then get written to from outside of the class are `registration_date` and `plate_type`. These other variables in your `attr_accessor` can probably be removed.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,67 @@
...
+    registrations.map do |registration|
+      new_vehicle = Vehicle.new({
+      vin: registration[:vin],
+      year: registration[:model_year],
+      make: registration[:make],
+      model: "unknown",
+      engine: :BOAT, county: registration[:county]})
+      vehicles << new_vehicle
+    end
+  end
```
abdulredd, line 30: I mentioned this earlier but it's worth mentioning again, `map` does not require a shoveling at the end.

PR #348, opened 2023-09-05
Sam Tran

Sykogst:
FInal

### DIFF
--- a/lib/dmv.rb
+++ b/lib/dmv.rb
@@ -1,5 +1,5 @@
 class Dmv
-
+  attr_reader :facilities
   def initialize
     @facilities = []
   end
@@ -9,7 +9,7 @@ class Dmv
   end
 
   def facilities_offering_service(service)
-    @facilities.find do |facility|
+    @facilities.find_all do |facility|
       facility.services.include?(service)
     end
   end
--- a/lib/dmv.rb
+++ b/lib/dmv.rb
@@ -13,4 +13,48 @@ class Dmv
       facility.services.include?(service)
     end
   end
+
+  def add_multiple_facilities(state_facilities_data)
+    state_facilities_data.each do |facility|
+      # For CO facilities
+      if facility[:state] == 'CO'
+        facility_details = {
+          name: facility[:dmv_office],
+          address: "#{facility[:address_li]} #{facility[:address__1]} #{facility[:city]} #{facility[:state]} #{facility[:zip]}",
+          phone: facility[:phone], 
+        }
+        # CO specifies services provided, all only provide registration
+        new_facility = Facility.new(facility_details)
+        new_facility.services << 'Vehicle Registration'
+        @facilities << new_facility
+      # For NY facilities
+      elsif facility[:state] == 'NY'
+        if facility[:street_address_line_2] != nil
+          address_full = "#{facility[:street_address_line_1]} #{facility[:street_address_line_2]} #{facility[:city]} #{facility[:state]} #{facility[:zip_code]}"
+        else
+          address_full = "#{facility[:street_address_line_1]} #{facility[:city]} #{facility[:state]} #{facility[:zip_code]}"
+        end
+        phone_extract = facility[:public_phone_number].to_s
+
+        facility_details = {
+          name: facility[:office_name],
+          address: address_full,
+          phone: "(#{phone_extract[0..2]}) #{phone_extract[3..5]}-#{phone_extract[6..9]}"
+        }
+        # NY has no specified services for facilities
+        @facilities << Facility.new(facility_details)
+      # For MO facilities
+      elsif facility[:state] == 'MO'
+        facility_details = {
+          name: facility[:name],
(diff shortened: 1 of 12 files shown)


lib/facility.rb
```diff
@@ -1,14 +1,59 @@
...
+    end
+  end
+
+  def renew_drivers_license(registrant)
+    if @services.include?('Renew License') && registrant.license_data[:license] == true
+      registrant.license_data[:renewed] = true
+    else
+      false
+    end
+  end
```
megstang, line 58: This logic and syntax looks really great! Nice job here

lib/facility.rb
```diff
@@ -1,14 +1,59 @@
 class Facility
-  attr_reader :name, :address, :phone, :services
-
-  def initialize(name, address, phone)
-    @name = name
-    @address = address
-    @phone = phone
+  attr_reader :name, :address, :phone, :registered_vehicles, :collected_fees
+  attr_accessor :hours, :services, :closed
```
megstang, line 3: I don't believe other classes ever call to change a Facility object's `services` attribute. For that reason, I think you can keep it as an attr_reader rather than an attr_accessor.

lib/registrant.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def permit?
+    @permit
+  end
+
+  def earn_permit
+    @permit = true
+  end
+end
```
megstang, line 21: THis all looks great 👍

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,20 @@
...
+        vin: vehicle[:vin_1_10], 
+        year: vehicle[:model_year], 
+        make: vehicle[:make], 
+        model: vehicle[:model], 
+        engine: :ev
+      }
+      @factory_vehicles << Vehicle.new(vehicle_details)
+    end
+  end
+end
```
megstang, line 20: In our Interaction Pattern, I don't believe it mentions that we need to be able to call `factory_vehicles` on a `VehicleFactory` object. For that reason, I don't think we need an attr_reader for that attribute. I would even go a bit further to say you don't need any attribute to get set upon initialization. That's because your `create_vehicles` method is meant to iterate and calculate those vehicles for you. We could refactor this class to look something like: 


```ruby 
class VehicleFactory

  def create_vehicles(state_registrations_data)
    state_registrations_data.map do |vehicle|
      vehicle_details = {
        vin: vehicle[:vin_1_10], 
        year: vehicle[:model_year], 
        make: vehicle[:make], 
        model: vehicle[:model], 
        engine: :ev
      }
    end
  end
end
```

lib/dmv.rb
```diff
@@ -9,8 +9,59 @@ def add_facility(facility)
...
+        new_facility = Facility.new(facility_details)
+        new_facility.hours = facility[:daysopen]
+        new_facility.closed = "#{facility[:holidaysclosed]} #{facility[:additionaldaysclosed]}"
+        @facilities << new_facility
+      else
+        false
+      end
+    end
+  end
 end
```
megstang, line 67: Woowee this is a big method!! I was really glad to read in your review that you were able to get through most of this on your own. I also saw that you mentioned you'd like to refactor this with more time. I definitely agree it could be refactored, but I'm mostly glad to see that you got it working! I like that you are looking at the state to be able to filter what you do with each set of data that you are creating Facility objects out of. You might consider creating different methods to format each state's data. It could be something like:

```ruby 
if facility[:state] == "CO"
   format_colorado_facility_data(facility)
elsif facility[:state] = "MO"
   format_mo_facility_data(facility)
elsif ...
...
...
```

spec/dmv_spec.rb
```diff
@@ -23,6 +23,74 @@
...
+      expect(@dmv.facilities_offering_service('Renew License').length).to be 0
+
+      expect(@dmv.facilities[0].name).to eq('DMV Tremont Branch')
+      expect(@dmv.facilities[0].address).to eq('2855 Tremont Place Suite 118 Denver CO 80205')
+      expect(@dmv.facilities[0].phone).to eq('(720) 865-4600')
+      expect(@dmv.facilities[4].services).to eq(['Vehicle Registration'])
+      expect(@dmv.facilities[0].registered_vehicles).to eq([])
+      expect(@dmv.facilities[0].collected_fees).to eq(0)
+      expect(@dmv.facilities[3].hours).to eq('Mon, Tue, Thur, Fri  8:00 a.m.- 4:30 p.m. / Wed 8:30 a.m.-4:30 p.m.')
+    end
```
megstang, line 45: This test and the tests that are similar to it are all okay. One thing you could do in addition to it is check that each element in the array that's returned is a Facility object. You could do something like:

```ruby
facilities = @dmv.add_multiple_facilities(co_dmv_facilities)
facilities.each do |facility|
   expect(facility).to be_an_instance_of(Facility)
end

megstang (commented):
@Sykogst Really nice work on this project! I left a few comments for you to take a look over. I'll follow up over slack in about a day to get you a bit more feedback. Let me know if you have any questions!

PR #349, opened 2023-09-05
Heather McNeil

hdmcneil:


lib/dmv_facility.rb
```diff
@@ -0,0 +1,58 @@
+DATA_SOURCES = {
+  co: { name:}
+  ny: { }
+  mo: { }
+}
```
jamisonordway, line 5: I think a constant variable like this could be useful, but since this isn't being used anywhere else in your code, it's best to remove it. Also, if you kept it, I'd expect it to be defined inside of the class (probably right below the initialize method)

lib/facility.rb
```diff
@@ -1,14 +1,87 @@
...
+              :address, 
+              :phone, 
+              :services,
+              :registered_vehicles, :collected_fees
 
-  def initialize(name, address, phone)
-    @name = name
-    @address = address
-    @phone = phone
+  def initialize(info, collected_fees = 0)
```
jamisonordway, line 8: Nice use of a default argument 🎉

lib/facility.rb
```diff
@@ -1,14 +1,87 @@
...
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
+    
   end
+
+  def register_vehicle(vehicle)
+    if available_service?("Vehicle Registration")
```
jamisonordway, line 23: You could flatten the conditional below by making this line into a guard clause, which will exit the method early if a certain condition isn't met.
```
unavailable_msg = "This facility does not offer vehicle registration services."
return unavailable_msg if available_service?("Vehicle Registration")
```

lib/facility.rb
```diff
@@ -1,14 +1,87 @@
...
+  def administer_written_test(registrant)
+    if available_service?("Written Test")
+      if registrant.permit? && registrant.age >= 16
+        registrant.license_data[:written] = true
+      else
+        registrant.license_data[:written] = false
+      end
+    else
+      registrant.license_data[:written] = false 
+      puts "This facility does not offer written test services."
```
jamisonordway, line 53: Generally, my advice is to avoid using `puts` outside of runner files.

lib/vehicle.rb
```diff
@@ -22,4 +26,13 @@ def antique?
...
     @engine == :ev
   end
+
+  def registration_date=(date)
+    @registration_date = date
+  end
+
+  def plate_type=(type)
+    @plate_type = type
+  end
```
jamisonordway, line 36: Yeah! Setter methods! 🎉

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,33 @@
...
+        vin: data[:vin_1_10],
+        make: data[:make],
+        model: data[:model],
+        year: data[:model_year],
+        engine: @vehicle_data[:engine]
+      )
+      @factory_vehicles << vehicle
+    end
+  end
+  @factory_vehicles
```
jamisonordway, line 30: I wonder if this would be better as a map, though I suppose if there were already elements in `@factory_vehicles`, using a map here wouldn't return those as well as all of the new ones. No strong opinion or suggestion here, just something to think about.

spec/facility_spec.rb
```diff
@@ -23,4 +24,62 @@
...
+      @camaro = Vehicle.new({vin: '1a2b3c4d5e6f', year: 1969, make: 'Chevrolet', model: 'Camaro', engine: :ice} )
+      end
+
+    it 'can add services to a facility' do
+      expect(@facility_1.add_service('Vehicle Registration')).to eq(["Vehicle Registration"])
+    end
+
+    it 'can read registration date' do
+      expect(@cruz.registration_date).to eq(nil)
+    end
```
jamisonordway, line 43: I don't think this needs to be its own test, though we might want to include the line as part of the `#register_vehicle` test later on

spec/facility_spec.rb
```diff
@@ -23,4 +24,62 @@
...
+    it 'has collected fees default to 0' do
+      expect(@facility_1.collected_fees).to eq(0)
+    end
+
+    it 'can register a vehicle' do
+      @facility_1.add_service("Vehicle Registration")
+      @facility_1.register_vehicle(@cruz)
+      expect(@cruz.registration_date).to eq(Date.today)
+      expect(@cruz.plate_type).to eq(:regular)
+      expect(@facility_1.registered_vehicles).to include(@cruz)
```
jamisonordway, line 58: This whole test example block is great! For this line, you could be even more specific and `expect(@facility_1.registered_vehicles).to eq([@cruz])`, verifying that it is the only vehicle in the array

spec/facility_spec.rb
```diff
@@ -23,4 +24,62 @@
...
+     it 'can register multiple vehicles' do
+      @facility_1.add_service("Vehicle Registration")
+      @facility_1.register_vehicle(@cruz)
+      @facility_1.register_vehicle(@camaro)
+      @facility_1.register_vehicle(@bolt)
+      expect(@camaro.registration_date).to eq(Date.today)
+      expect(@camaro.plate_type).to eq(:antique)
+      expect(@bolt.registration_date).to eq(Date.today)
+      expect(@bolt.plate_type).to eq(:ev)
+      p @facility_1.registered_vehicles
```
jamisonordway, line 71: I would stay away from `p` and `puts` in spec files. To test this in an RSpec way, you might use something like
```
expect(@facility_1.registered_vehicles).to eq([@cruz, @camaro, @bolt])
```

spec/facility_spec.rb
```diff
@@ -23,4 +24,62 @@
...
+      expect(@facility_1.collected_fees).to eq(325)
+     end
+
+     it 'works across other facilities' do
+      expect(@facility_2.registered_vehicles).to eq([])
+      expect(@facility_2.services).to eq([])
+      expect(@facility_2.register_vehicle(@bolt)).to eq(nil)
+      expect(@facility_2.registered_vehicles).to eq([])
+      expect(@facility_2.collected_fees).to eq(0)
+     end
```
jamisonordway, line 81: As part of this test, I would also include some lines where you add the 'Register Vehicle' service and register at least 1 vehicle.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+    end
+
+    it 'earns a permit' do
+      @registrant_2.earn_permit
+      expect(@registrant_2.permit?).to eq(true)
+    end
+  end
+end
+
+RSpec.describe 'getting a drivers license' do
```
jamisonordway, line 35: Conventionally, each RSpec file will only have one _outer_ `RSpec.describe` like this. You could definitely nest this group of tests within the block that ends on your current line 33, though, with a `describe 'getting a drivers license'`

spec/registrant_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+      @registrant_2.earn_permit
+      expect(@facility_1.administer_written_test(@registrant_2)).to be(true)
+      expect(@registrant_2.license_data).to eq({:written=>true, :license=>false, :renewed=>false})
+      expect(@registrant_3.age).to eq(15)
+      expect(@registrant_3.permit?).to eq(false)
+      expect(@facility_1.administer_written_test(@registrant_3)).to eq(false)
+      @registrant_3.earn_permit
+      expect(@facility_1.administer_written_test(@registrant_3)).to be(false)
+      expect(@registrant_3.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+    end
```
jamisonordway, line 65: This test is very thorough, nice job!
One suggestion I have is to break it up into multiple 'it' blocks, for better maintainability and readability in the long run. The tests would be the same, but spread out across examples that might look something like
```
it 'does not administer written test when facility does not offer the service' do
  ...
end

it 'administers a written test to an eligible registrant when the facility offers the service' do
  ...
end

it 'does not administer written test to registrants without a permit' do
  ...
end

it 'does not administer written test to registrants under 16' do
  ...
end
```

Alternatively, you could further organize your tests using more `describe` blocks.
```
describe 'facility offers written test service' do
  describe 'eligible registrants' do
    it 'administers a written test' do
      ...
    end
  end

  describe 'ineligible registrants' do
    it 'does not administer written test if registrant has no permit' do
      ...
    end

    it 'does not administer written test if registrant is under 16' do
      ...
    end
  end
end
```
And so on.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+      @facility_1.administer_written_test(@registrant_2)
+      @facility_1.administer_road_test(@registrant_2)
+      @facility_1.renew_drivers_license(@registrant_2)
+      expect(@facility_1.renew_drivers_license(@registrant_2)).to eq(true)
+      expect(@registrant_2.license_data).to eq(@registrant_2.license_data)
+
+    end
+  end
+  end
+end
```
jamisonordway, line 118: Overall, great job testing all the eligibility scenarios for the written/road/license behavior.

jamisonordway (commented):
Most of my feedback here has to do with Ruby convention, refactoring suggestions, and recommendations for your tests (especially `facility_spec.rb` since you mentioned that in your submission). Please let me know if you have any follow up questions about these comments.

PR #350, opened 2023-09-05
Brendan Bondurant

brendan-bondurant:
Brendan Bondurant DMV


lib/dmv.rb
```diff
@@ -1,16 +1,26 @@
...
+  end
+
   def facilities_offering_service(service)
-    @facilities.find do |facility|
+    @facilities.select do |facility|
       facility.services.include?(service)
     end
   end
 end
+
```
abdulredd, line 26: Great job with this class. I think having this `import_facility` method makes a lot of sense here. I think I still would have preferred to see this behavior in a separate class as we did with the `VehicleFactory`.

The only other thing would be to make sure to require the `Facility` class at the top of this file since you are instantiating the facility objects from within the file inside the class.

lib/facility.rb
```diff
@@ -1,14 +1,141 @@
...
+      @registered_vehicles << vehicle
+      vehicle.plate_type = :ev
+    elsif @services.include?('Vehicle Registration') == true
+      vehicle.registration_date = Date.today
+      @collected_fees += 100
+      @registered_vehicles << vehicle
+      vehicle.plate_type = :regular
+    else @services.include?('Vehicle Registration') == false
+    end
+  end
```
abdulredd, line 108: This one method is doing a lot of things. It's registering the vehicle AND collecting the fees AND setting the vehicle plate. This goes against SRP and you can consider breaking this method up into separate helper methods:

```ruby
def register_vehicle(vehicle)
    vehicle.register
    self.collect_fees(vehicle)
    self.set_plate_type(vehicle)
    
    return true
  end

  false
end
```

Then you would have 2 helper methods that would collect the fees, and set the plate type based on the engine.

You'll also notice that I dropped all the `== true` since this is unnecessary. Because we are using Boolean logic, the result will either be `true` or `false` there is not need to do the comparison.

Lastly, `else` does not take a conditional so when formatting the `else` branch in your `if` statements, make sure the logic goes underneath and not next to the `else`.

This:

```ruby
else
  return false
end
```

Not this:

```ruby
else @services.include?('Vehicle Registration') == false
end
```

lib/facility.rb
```diff
@@ -1,14 +1,141 @@
...
+  def renew_drivers_license(registrant)
+    if @services.include?('Renew License') == true && registrant.license_data[:license] == true
+      return registrant.license_data[:renewed] = true
+    else
+      false
+    end
+  end
 end
+
+
```
abdulredd, line 135: For `administer_written_test`, `administer_road_test`, `renew_drivers_license` there is not need for the explicit `return` if you are using the else. Removing the explicit `return` would produce the same result. You would only need it if you did not have the else.

```ruby
  if @services.include?('Written Test') && registrant.permit && registrant.age >= 16
  else
    false
  end
end
```

And this is the same:

```ruby
  if @services.include?('Written Test') && registrant.permit && registrant.age >= 16
    return registrant.license_data[:written] = true
  end

  false
end
```

lib/registrant.rb
```diff
@@ -0,0 +1,18 @@
...
+              :permit,
+              :license_data
+  attr_writer :license_data
+  
+  def initialize(name, age, permit = false)
+    @name = name
+    @age = age
+    @license_data = { written: false, license: false , renewed: false }
+    @permit = permit
+  end
```
abdulredd, line 13: You've defined an `attr_writer` for `@license_data`, which allows direct modification of the license_data hash from outside the class. This is only necessary if you want to allow `@license_data` to point to something other than the hash it is currently pointing to. 

Because `@license_data` is pointing to a hash and hashes can be updated without the `attr_reader`, line `6` can be removed.

lib/vehicle.rb
```diff
@@ -22,4 +29,28 @@ def antique?
...
+  end
+  
+  def parse_engine(data)
+    if data[:engine]
+      data[:engine]
+    else
+      :ev
+    end
+  end
 end
```
abdulredd, line 56: 👍

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,13 @@
...
+  end
+
+  def create_vehicles(api_data)
+    api_data.each do |data|
+      vehicle = Vehicle.new(data)
+      @factory_vehicles << vehicle
+    end
+    @factory_vehicles
+  end
+end
```
abdulredd, line 13: Great job here. I see on line `11` is returning the `@factory_vehicles` array at the final operation in the method. You can eliminate the need to do this if you switch out `each` for `map` like so:

```ruby
require "./lib/vehicle"

class VehicleFactory
  def create_vehicles(api_data)
    api_data.map do |data|
      Vehicle.new(data)
    end
  end
end
```

1. You'll see I completely got rid of the `initialize` and the `@factory_vehicles` array altogether
2. Converted `each` to `map` since `map` transforms every element in an array from one thing to something else, in this case -- `Vehicle` objects
3. Don't forget to require the `Vehicle` class at the top of this file since you are instantiating new vehicles

spec/registrant_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      expect(registrant_1.permit).to eq(true)
+      expect(registrant_2.permit).to eq(false)
+    end
+    
+    it 'has license data' do
+      registrant_1 = Registrant.new('Bruce', 18, true )
+      registrant_2 = Registrant.new('Penny', 15 )
+      expect(registrant_1.license_data).to be_an_instance_of(Hash)
+      expect(registrant_2.license_data).to be_an_instance_of(Hash)
+    end
```
abdulredd, line 39: Having gone through all your tests, this is the only thing I want to explicitly call out. Your license data tests should be testing the actual value of calling `license_data` on a registrant. Not just making sure it's a hash.

```ruby
expect(registrant_1.license_data).to eq({ written: false, license: false , renewed: false })
```

"Data Type" tests should only be reserved for API's or when something might be difficult to test.

PR #353, opened 2023-09-06
Anthea Yur

chisPmama:
Pull request with my final submission of DMV project! Sorry for the delay!


lib/facility.rb
```diff
@@ -1,14 +1,84 @@
...
-    @address = address
-    @phone = phone
+  def initialize(hash, collected_fees = 0)
+    @name = hash[:name]
+    @address = hash[:address]
+    @phone = hash[:phone]
     @services = []
+    @collected_fees = collected_fees
+    @registered_vehicles = []
   end
```
megstang, line 16: I like seeing that you've found out how to set default values upon initialization. Looks good!

lib/facility.rb
```diff
@@ -1,14 +1,84 @@
...
 
-  def add_services(service)
+  def add_service(service)
     @services << service
+    @services = @services.uniq
   end
```
megstang, line 21: This looks great! Alternatively, we could also only shovel the service into the `@service` array IF it didn't already exist in the array. 

Something like: 

```ruby
@services << service if !@services.include?(service)
```

megstang: To be clear, either way would be totally fine.. but sometimes it's fun to see a different approach ;)

lib/facility.rb
```diff
@@ -1,14 +1,84 @@
...
+        elsif vehicle.electric_vehicle?
+          vehicle.plate_type = :ev
+          @collected_fees+=200
+        else
+          vehicle.plate_type = :regular
+          @collected_fees+=100
+        end
+        vehicle
+      else
+        "Oops! Looks like this vehicle has already been registered on #{vehicle.registration_date}."
```
megstang, line 40: fun! I like this edge case that you considered!

lib/facility.rb
```diff
@@ -1,14 +1,84 @@
...
+      end
+     end
+   end
```
megstang, line 43: Your logic here looks really good! Something that could be fun to look into is creating helper methods to slim up this method, and align more with SRP (single responsibility principle). One thing I might do to refactor this method is create 3 more methods specifically for registering different types of vehicles. If you pulled out some of this logic into three other methods, your code could look like: 

```ruby
        if vehicle.antique?
          register_antique_vehicle(vehicle)
          register_ev_vehicle(vehicle)
        else
          register_reg_vehicle(vehicle)
        end
```

lib/facility.rb
```diff
@@ -1,14 +1,84 @@
...
+      true
+    else
+      false
+    end
+  else
+    "Sorry, this location does not service License Renewals."
+  end
+ end
+
+end
```
megstang, line 84: Logic looks really great on these! Again, I really like how you're thinking about usability, and returning helpful messages when conditions aren't met as expected (referring to lines 54, 67, and 80)

lib/facility_creator.rb
```diff
@@ -0,0 +1,31 @@
...
+      facilities << Facility.new({
+      name: facility[:dmv_office], 
+      address: facility[:address_li] + " " + facility[:city] + " " + facility[:state]+ " " + facility[:zip], 
+      phone: facility[:phone]})
+    end
+    end
+  facilities
+  end
+end
+
```
megstang, line 31: This looks good.. I'm curious if you could use the `:state` value in the API response to help dictate your conditional. If you can, I'd probably prefer to go that route rather than checking to see what keys are present/not present. 

So, something like:

```ruby
if location[:state] == "CO"
...
elsif location[:state] == "NY"
...
elsif location[:state] == "MO"
...
end
```

lib/registrant.rb
```diff
@@ -0,0 +1,20 @@
+class Registrant
+  attr_reader :name, :age
+  attr_accessor :license_data, :permit
```
megstang, line 3: I don't think you'll need to list `:permit` as an attr_accessor here. If you look throughout you're code, you never need to call `.permit` and assign it to a new value in another class. I actually think you could get rid of this altogether (no need for `:permit` as a reader or an accessor), because you can check the status of the permit by calling `permit?`, and you can change the status of the permit by calling `earn_permit`.

lib/facility.rb
```diff
@@ -1,14 +1,84 @@
...
+        vehicle
+      else
+        "Oops! Looks like this vehicle has already been registered on #{vehicle.registration_date}."
+      end
+     end
+   end
+
+   def administer_written_test(registrant)
+    if @services.include?("Written Test")
+      if registrant.age >= 16 && registrant.permit == true
```
megstang, line 47: I think we can use `registrant.permit?` here since that will either return true or false automatically.

lib/facility.rb
```diff
@@ -1,14 +1,84 @@
...
+        false
+      end
+    else
+      "Sorry, this location does not service written tests."
+   end
+  end
+
+  def administer_road_test(registrant)
+  if @services.include?("Road Test")
+    if registrant.permit == true && registrant.age >=16
```
megstang, line 60: same thing here... could use `registrant.permit?`

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,19 @@
...
+    registrations.each do |registration|
+      vehicles << Vehicle.new({
+        vin: registration[:vin_1_10], 
+        year: registration[:model_year], 
+        model: registration[:model],
+        make: registration[:make],
+        engine: :ev})
+    end
+    vehicles
+  end
```
megstang, line 14: Nice! This could also be a good use of `map`. That way, you wouldn't need a placeholder on line 4, and you wouldn't need to return `vehicles` on line 13.

spec/facility_creator_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  it 'can accept information from Missouri submissions' do
+    missouri_facilities = []
+    missouri_facilities <<  DmvDataService.new.mo_dmv_office_locations[0]
+    @check_results << @creator.create_facility(missouri_facilities) 
+    expect(DmvDataService.new.mo_dmv_office_locations[0].class).to be(Hash)
+    expect(@check_results[3].to_s).to include("MO")
+    expect(@check_results[3].class).to be(Array)
+  end
```
megstang, line 59: (This comment can be applied to a few tests you have)

This test is okayyyyy. I don't love that you are shoveling values into an array to check in your specs. This could cause some headaches later when you need to add more specs. What I'd suggest doing for a test like this is something like: 

```ruby
mo_dmv_api_response = DmvDataService.new.mo_dmv_office_locations
mo_facilities = @creator.create_facility(mo_dmv_api_response)
expect(mo_facilities).to be_an_instance_of(Array)
mo_facilities.each do |mo_facility|
   expect(mo_facility).to be_an_instance_of(Facility)
end 
```

megstang (commented):
@chisPmama Really nice work on this! I will follow up with you within the day to get you more feedback. In the meantime, please take a look at my comments and let me know if you have any questions!

PR #355, opened 2023-09-07
Matt Darlington

mdarl17:
delete scratch.rb, analytics_spec.rb; clean up files


lib/vehicle_factory.rb
```diff
@@ -0,0 +1,9 @@
+require './spec/spec_helper'
```
abdulredd, line 1: No necessary. This file should be requiring the Vehicle class file and nothing else.

abdulredd:
It's very obvious your comfortability with code, but this often leads to instances where some of your logic might be a little over-engineered and could maybe be both simplified in complexity and readability. With time, anyone can understand the logic, but I think there are still improvements that can be made in making the code easier to onboard.

PR #356, opened 2023-10-19
Fix bugs found via spec files

SamPuttman:
This fixes the undefined method and hash bugs found via the spec files.


SamPuttman:
Looks good Sam! Great work!

PR #372, opened 2023-10-24
Martin Chavez 

Chavezgm:


lib/facility.rb
```diff
@@ -1,14 +1,59 @@
...
+        @collected_fees += 25
+      elsif vehicle.electric_vehicle? == true
+        vehicle.plate_type = :ev 
+        @collected_fees += 200
+      else 
+        vehicle.plate_type = :regular
+        @collected_fees += 100
+      end
+    end
+  end
```
megstang, line 33: This method looks great!

lib/facility.rb
```diff
@@ -1,14 +1,59 @@
...
+
+  def renew_drivers_license(registrant)
+    if @services.include?('Renew License') && registrant.license_data[:written] == true && registrant.license_data[:license] == true
+      registrant.license_data[:renewed] = true
+    else 
+      false
+    end
+  end
 end
```
megstang, line 59: All of these methods look great. Good job with the conditionals in each of them. You could possibly slim them up a bit by taking out conditions like `== true` because those values should either return true or false automatically. So, something like this: 

```ruby 
    if @services.include?('Written Test') && registrant.permit? && registrant.age >= 16
    else 
      false 
    end
  end

  def administer_road_test(registrant)
    if @services.include?('Road Test') && registrant.license_data[:written]
      registrant.license_data[:license] = true 
    else 
      false 
    end
    # registrant.license_data[:license]
  end

  def renew_drivers_license(registrant)
    if @services.include?('Renew License') && registrant.license_data[:written] && registrant.license_data[:license]
      registrant.license_data[:renewed] = true
    else 
      false
    end
  end
```

lib/registrant.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  def permit?
+    @permit 
+  end
+
+  def earn_permit
+    @permit = true
+  end
+end
```
megstang, line 17: 👍

lib/facility_factory.rb
```diff
@@ -0,0 +1,28 @@
...
+      else 
+        name = data[:name]
+        address = "#{data[:address1]} #{data[:city]} #{data[:state]} #{data[:zipcode]} #{data[:county]}"
+        phone = data[:phone]
+      end
+      # require 'pry'; binding.pry
+      facility = Facility.new(name: name, address: address, phone: phone)
+    end
+  end
+end
```
megstang, line 28: There really isn't a need for an initialize method in this class since we don't need to store any state/attributes.

lib/vehicle.rb
```diff
@@ -6,13 +6,18 @@ class Vehicle
               :make,
               :model,
               :engine
+  attr_accessor :registration_date,
+                :plate_type
+
```
megstang, line 11: Nice, I like that these are the only accessor's that you've set.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,14 @@
...
+      vin = registration[:vin_1_10]
+      model_year = registration[:model_year]
+      make = registration[:make]
+      model = registration[:model]
+
+      vehicle = Vehicle.new(engine: engine, vin: vin, year: model_year, make: make, model: model )
+    end
+    # require 'pry'; binding.pry
+  end
+end
```
megstang, line 14: Looks good!

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      expect(factory).to be_a(FacilityFactory)
+      facilities = factory.create_facility(@Colorado)
+      expect(facilities).to be_a(Array)
+      expect(@Colorado[0][:dmv_office]).to be_a(String)
+      expect(facilities[0]).to be_a(Facility)
+      expect(@Colorado[0][:address_li]).to be_a(String)
+      expect(@Colorado[0][:address__1]).to be_a(String)
+      expect(@Colorado[0][:city]).to be_a(String)
+      expect(@Colorado[0][:state]).to be_a(String)
+      expect(@Colorado[0][:state]).to eq("CO")
```
megstang, line 22: Right here, you're testing the data from teh api which isn't really necessary. We mainly need you to test the data that's getting returned to you from your `create_facility` method.

megstang: It looks like your return value for that method is stored in `facilities`. You could do several things. For starters, I'd suggest just pulling out the first element in that array (`facilities.first`) and making sure it's a Facility object, and then making sure that each of it's expected attributes are not nil (that woul dmean that your attribute assignments were going well).

megstang (commented):
@Chavezgm I left some comments for you to look over. Please let me know if you have any questions. I'll follow up in slack within a day to get you more feedback.

PR #374, opened 2023-10-24
Nathan Lambertson

lambo1986:


lib/dmv.rb
```diff
@@ -1,5 +1,9 @@
-class Dmv
+require './lib/facility'
+require './lib/dmv'
 
+class Dmv
+attr_reader :facilities
+attr_accessor :facilities
```
megstang, line 6: You only need one of these. Both of these lines will create a `facilities` method that returns `@facilities`. But the attr_accessor will allow other classes to change the `@facilities` attribute. Looking through your code, I don't think you need to do that, so an attr_reader should be just fine.

lib/dmv_data_service.rb
```diff
@@ -11,6 +11,10 @@ def wa_ev_registrations
     @wa_ev_registrations ||= load_data('https://data.wa.gov/resource/rpr4-cgyd.json')
   end
 
+  def ny_registrations
+    @ny_registrations ||= load_data('https://data.ny.gov/resource/w4pv-hbkt.json')
+  end
```
megstang, line 16: 👍

lib/facility.rb
```diff
@@ -1,14 +1,189 @@
...
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+  #refactor
+  def register_vehicle(car)
+    if services.find {|service| service.include?('Vehicle Registration')}
+    current_time = DateTime.now
+    cdt = current_time.strftime "%d/%m/%Y %H:%M"
+    car.registration_date = cdt
```
megstang, line 39: Cool idea to format the Date object. I like it!

lib/facility.rb
```diff
@@ -1,14 +1,189 @@
...
+    @collected_fees = 0
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+  #refactor
+  def register_vehicle(car)
+    if services.find {|service| service.include?('Vehicle Registration')}
```
megstang, line 36: you could probably slim this line down a bit by doing something like: 

`if @services.include?("Vehicle Registration")`

lib/facility.rb
```diff
@@ -1,14 +1,189 @@
...
+        car.plate_type = :ev
+      end
+
+      if car.plate_type == :antique
+        @collected_fees += 25
+      elsif car.engine == :ev
+        @collected_fees += 200
+      else
+        @collected_fees =+ 100
+      end
```
megstang, line 56: you could probably assign collected feeds while going through your first conditional. 

```ruby
      if Date.today.year - car.year > 25
        car.plate_type = :antique
        @collected_fees += 25
      elsif Date.today.year - car.year <= 25 && car.engine == :ice
        car.plate_type = :regular
        @collected_fees += 100
      else
        car.plate_type = :ev
        @collected_fees += 200
      end
```

lib/facility.rb
```diff
@@ -1,14 +1,189 @@
...
+    elsif list.count == 172 || list.find {|omv| omv[:state] == "NY"} #NY
+      list.each do |omv|
+        new_omv = {
+          name: omv[:office_name],
+          address: omv[:street_address_line_1],
+          phone: omv[:public_phone_number]
+        }
+        nf_arr << Facility.new(new_omv)
+      end
+    else
```
megstang, line 128: I don't love how you're checking the count of the list. What if NY suddenly opened 7 more DMV's? Then they'd get dropped into the block on line 110-118 because it now has 179 DMV offices. I do, however, like your idea of checking the state. That seems like a better approach for formatting rather than the count of the offices.

lib/facility.rb
```diff
@@ -1,14 +1,189 @@
...
+      hour_array = []
+      source.each do |omv|
+        office_hours = {
+          omv[:name] => omv[:daysopen]
+        }
+        hour_array << office_hours
+      end
+      hour_array
+    end
+  end
```
megstang, line 173: You could also update your `create_facilities` method to pull hours out of the data and pass that information along to the FAcility class. Then, you would need to update your facility class to pull out an attribute for hours, and have a way to read that data (`attr_reader`)

megstang: I like the above approach because then you could work with FAcility objects to find out the hour information rather than just raw Hash data.

lib/registrant.rb
```diff
@@ -0,0 +1,30 @@
...
+      true
+    else
+      false
+    end
+  end
+
+  def earn_permit
+    @permit = true
+  end
+end
```
megstang, line 30: You don't need line 7 because we never need to call `.permit` on a Registrant object. In your `permit?` method, you could simplify this a bit to look like: 

```ruby 
def permit? 
   @permit 
end 
```

since `@permit` will automatically return true or false.

lib/vehicle.rb
```diff
@@ -5,14 +5,24 @@ class Vehicle
...
+              :engine,
+              :registration_date,
+              :plate_type
+  attr_accessor :vin,
+                :year,
+                :make,
+                :model,
+                :engine,
+                :registration_date,
+                :plate_type
```
megstang, line 17: Same comment as before -- selectively set attributes as accessors. I believe the only attributes you need to be able to update outside of this class are `registration_Date` and `plate_type` so those should be the only accessors, and the rest could be readers.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,344 @@
...
+          make: car[:make],
+          body_type: car[:body_type],
+          engine: :ev}
+        nc_arr << Vehicle.new(new_car)
+      end
+      nc_arr
+    else
+      "N/A"
+    end
+  end
```
megstang, line 41: 👍

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,344 @@
...
+    chevy_bolt_euv = []
+    ford_fusion = []
+    volvo_xc60 = []
+    volvo_xc40 = []
+    hyundai_ioniq5 = []
+    ford_transit = []
+    audi_a3 = []
+    hyundai_ioniq = []
+    bmw_x5 = []
+    porsche_cayenne = []
```
megstang, line 95: I don't think you need any of this with the approach you're taking. Rather, I'd suggest doing something like this in the map below: 

```ruby 
 cars_by_model = Hash.new()
 by_make_and_model.map do |car|
   cars_by_model[car]  += 1
 end 
...
```

this would create keys for the car's make/model, and increment one every time that make/model came through.

megstang: You could apply this idea to a lot of the code you have for Iteration 4.

spec/facility_spec.rb
```diff
@@ -1,26 +1,181 @@
...
+      colorado_facilities = DmvDataService.new.co_dmv_office_locations
+      new_york_facilities = DmvDataService.new.ny_dmv_office_locations
+      missouri_facilities = DmvDataService.new.mo_dmv_office_locations
+
+      facility_list.create_facility(colorado_facilities)
+      facility_list.create_facility(new_york_facilities)
+      facility_list.create_facility(missouri_facilities)
+
+      expect(facility_list.create_facility(colorado_facilities)).to be_a(Array)
+    end
```
megstang, line 158: For tests like this, I'd also suggest going into that array a bit and testing that the elements are in fact fAcility objects and that their attributes are not nil.

For starters, you could just pull the first element out of the array and test that it's the right kind of object, and that when you call to get the attributes, you aren't getting returned `nil`.

megstang (commented):
@lambo1986 I left some comments for you to look over. Please let me know if you have any questions. I'l follow up in slack by EOD today.

lib/facility.rb
lambo1986: Thanks I had to use google to figure that out. Pretty neat that time is a class in ruby!

lambo1986: or Date, rather

lib/facility.rb
lambo1986: Ahhh I didn't think about what would happen if they had the same count!

lib/vehicle.rb
lambo1986: Yes I should have cleaned this up. Sorry!

PR #375, opened 2023-10-24
Sam Puttman Project

SamPuttman:
Hi! I hope this works! When I initially created my repo, I cloned it instead of forking it, so it was not connected. The link to my original repo is: https://github.com/SamPuttman/the_dmv

### DIFF
--- a/lib/dmv.rb
+++ b/lib/dmv.rb
@@ -1,4 +1,3 @@
-#fixed bugs
 class Dmv
 attr_reader :facility, :service
   
--- a/lib/dmv.rb
+++ b/lib/dmv.rb
@@ -1,7 +1,12 @@
 class Dmv
-
+attr_reader :facility, :service
+  
   def initialize
     @facilities = []
+  end
+
+  def facilities
+    @facilities
   end
 
   def add_facility(facility)
@@ -9,7 +14,7 @@ class Dmv
   end
 
   def facilities_offering_service(service)
-    @facilities.find do |facility|
+    @facilities.select do |facility|
       facility.services.include?(service)
     end
   end
--- a/lib/facility.rb
+++ b/lib/facility.rb
@@ -1,14 +1,14 @@
 class Facility
   attr_reader :name, :address, :phone, :services
 
-  def initialize(name, address, phone)
-    @name = name
-    @address = address
-    @phone = phone
+  def initialize(facility_details)
+    @name = facility_details[:name]
+    @address = facility_details[:address]
+    @phone = facility_details[:phone]
     @services = []
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
 end
--- a/lib/facility.rb
+++ b/lib/facility.rb
@@ -1,3 +1,5 @@
+require 'date'
+
 class Facility
   attr_reader :name, :address, :phone, :services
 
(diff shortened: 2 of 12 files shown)


lib/dmv.rb
```diff
@@ -1,15 +1,20 @@
 class Dmv
-
+attr_reader :facility, :service
```
megstang, line 2: I'm curious about both of these attr_readers. I don't see any attributes named `@facility` or `@service`, so there doesn't seem to be a reson to have them.

lib/facility.rb
```diff
@@ -1,14 +1,89 @@
...
+    return vehicle.plate_type
+  end
+
+  def registered_vehicles
+    @registered_vehicles
+  end
+
+  def collected_fees
+    @collected_fees
+  end
```
megstang, line 46: you could use `attr_reader`s for these methods if you wanted.

lib/facility.rb
```diff
@@ -1,14 +1,89 @@
...
+      return false
+    elsif registrant.permit? == false
+      return false
+    elsif registrant.age < 16
+      return false
+    else @services.include?('Written Test') and registrant.permit? == true
+      registrant.license_data = { written: true, license: false, renewed: false }
+      return true
+    end
+  end
```
megstang, line 59: This piece of feedback could go for the other two methods below this method. I think you could slim this method up a bit by checking conditions on one line. For example: 

```ruby
    if !@services.include?('Written Test') || !registrant.permit? || registrant.age < 16
      return false
    else @services.include?('Written Test') and registrant.permit? == true
      registrant.license_data = { written: true, license: false, renewed: false }
      return true
    end

```

lib/facility_factory.rb
```diff
@@ -0,0 +1,17 @@
...
+  def create_facilities(facilities_data)
+    @facility_instances = facilities_data.map do |facility|
+      Facility.new(
+        name: facility[:dmv_office] || facility[:office_name] || facility[:name],
+        address: facility[:address_li] || facility[:street_address_line_1] || facility[:address1],
+        phone: facility[:phone] || facility[:public_phone_number] || facility[:phone]
+      )
+    end
+  end
+end
```
megstang, line 17: I don't think you need the `:facility_instances` method ... or the instance variable. We really only need the `create_facilities` method to return a collection of Facility objects. No need to store state, since this response will change anytime the data passed in is changed.

lib/vehicle.rb
```diff
@@ -1,18 +1,20 @@
 require 'date'
 
 class Vehicle
-  attr_reader :vin,
-              :year,
-              :make,
-              :model,
-              :engine
+  attr_accessor :vin, :year, :make, :model, :engine, :registration_date, :plate_type
```
megstang, line 5: similarly here.. consider which of these attributes can be readers, and which of them should be accessors.

megstang: It looks like you change `plate_type` and `registration_date` in your Facility class, so those will probably be the only accessor's that you need to list, and the rest can be readers.

lib/vehicle.rb
```diff
@@ -1,18 +1,20 @@
...
     @make = vehicle_details[:make]
     @model = vehicle_details[:model]
     @engine = vehicle_details[:engine]
+    @plate_type = nil
+    @registration_date = nil
+  end
+
+  def registration_date
+    @registration_date
   end
```
megstang, line 19: Since you have a method in your attr_accessor's for `registration_date`, there's no need for this method to return the attribute.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,50 @@
+class VehicleFactory
+  attr_reader :vehicle_instances
+
+  def initialize
+    @vehicle_instances = []
+  end
```
megstang, line 6: Similar to your FacilityFactory.. no need for you to store state for the vehicle objects.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,50 @@
...
+    model_count = {}
+
+    @vehicle_instances.each do |vehicle|
+      make_count[vehicle.make] =+ 1
+      model_count[vehicle.model] =+ 1
+    end
+    most_popular_make = find_most_popular(make_count)
+    most_popular_model = find_most_popular(model_count)
+
+    return most_popular_make, most_popular_model
```
megstang, line 31: ooh, fun! It's cool that you found a way to return two values at once!

lib/facility_factory.rb
megstang: I know you asked for feedback specifically on lines 8-14. THis approach looks good to me! It should work for this set of data othat you're working with. It's possible that once another set of data is added for another state, you might need to refactor this, but for now, looks great!

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    facility_factory = FacilityFactory.new
+    facilities_data = DmvDataService.new.co_dmv_office_locations
+    facility_factory.create_facilities(facilities_data)
+
+    expect(facility_factory.facility_instances).not_to eq(nil)
+    expect(facility_factory.facility_instances.empty?).not_to eq(true)
+
+    facility_factory.facility_instances.map do |facility|
+      expect(facility).to be_an_instance_of(Facility)
+    end
```
megstang, line 33: You should also test the attributes that you'd expect to get back for each of tehse facility objects. Make sure that all of their addresses come back and are not nil, etc.

PR #376, opened 2023-10-24
Yain's DMV Project

yainporter:
This was a fun project!


lib/dmv.rb
```diff
@@ -9,7 +9,7 @@ def add_facility(facility)
   end
 
   def facilities_offering_service(service)
-    @facilities.find do |facility|
+    @facilities.find_all do |facility|
```
KatBrandt, line 12: Nice!

lib/facility.rb
```diff
@@ -1,14 +1,66 @@
...
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(car)
+    if @services.include?("Vehicle Registration")
+      car.new_registration_date
+      collect_fees(car)
+      @registered_vehicles << car
+    end
```
KatBrandt, line 26: Nice job with this method.

lib/facility.rb
```diff
@@ -1,14 +1,66 @@
...
+  end
+
+  def administer_written_test(registrant)
+    if @services.include?("Written Test") && registrant.permit?
```
KatBrandt, line 30: nice use of logical operators to avoid nested if statements

lib/facility.rb
```diff
@@ -1,14 +1,66 @@
...
+  def renew_drivers_license(registrant)
+    if @services.include?("Rewnew License") && registrant.license_data[:license] == true
+      registrant.license_data[:renew] = true
+      true
+    else
+      false
+    end
+  end
+
+  ################Helper Methods#############
```
KatBrandt, line 56: Nice, I like the helper method! You can put the keyword `private` on line 56 and that makes these methods only accessible from inside the class. usually we put helper methods there.

lib/facility_locations.rb
```diff
@@ -0,0 +1,105 @@
...
+    end
+  end
+
+  def address(dmv_data)
+    dmv_data.each do |facility|  
+      old_address = facility.select {|key,value| key.name.include?("address")}
+      old_city = facility.select {|key,value| key.name.include?("city")}
+      old_state = facility.select {|key,value| key.name.include?("state")}
+      old_zip = facility.select {|key,value| key.name.include?("zip")}
+      facility[:address] = [old_address, old_city, old_state, old_zip]
```
KatBrandt, line 47: Nice work with this method. I like how you pulled it apart

lib/facility_locations.rb
```diff
@@ -0,0 +1,105 @@
...
+  end
+
+  def create_facilities(dmv_data)
+      dmv_data.each do |facility| 
+      hours(dmv_data)
+      phone(dmv_data)
+      address(dmv_data)
+      name(dmv_data)
+      holidays_closed(dmv_data)
+      facility = Facility.new(facility)
```
KatBrandt, line 15: I like the idea you're on here, with making all these helper methods to do specific things. The only problem I see is that basically, you're passing around this giant array of hashes to each of these methods, iterating through it, changing it and manipulating it, and if that array had a million or more elements in it, that would be extremely costly and slow. Instead, we can take pieces out of the array and then when we make changes, we can store that updated piece in a variable. 
For example, line 10 here, maybe it's something like `facility_hours = hours(dmv_data)` and then instead of returning that whole entire hash again, we just return the section of data that relates to the facility hours. Then we can sent those variables into Facility.new in a hash. 
Hoping that makes sense!

lib/facility_locations.rb
```diff
@@ -0,0 +1,105 @@
...
+  end
+
+  def hours(dmv_data)
+    dmv_data.each do |facility| 
+      # Turns :hours => hash
+      check_for_multiple_hours(facility)
+      # Turns :hours => string
+      hours_of_facility(facility)
+      # Checks for closed down facilities
+      check_for_closed_notice(facility)
```
KatBrandt, line 67: Really great job using all these helper methods!

lib/facility_locations.rb
```diff
@@ -0,0 +1,105 @@
...
+    end
+  end
+
+# Helper method for finding closed offices
+  def check_for_closed_notice(facility)
+    if facility[:hours] == nil
+      facility.select do |key, value|
+        if value != nil && (value.include?("CLOSED UNTIL") || value.include?("closed until"))
```
KatBrandt, line 75: Careful here, you're using a .select so you don't need conditionals inside the select. Potentially could have used a different enumerable.

lib/facility_locations.rb
```diff
@@ -0,0 +1,105 @@
...
+  end
+# Helper method to find the :hours of a facility
+  def hours_of_facility(facility)
+    if facility.has_key?(:hours) == false
+      old_hours = facility.keys.select {|key| key.name.include?("hour")|| key.name.include?("daysopen")}[0]
+      facility[:hours] = facility.delete old_hours 
+    end
+  end
+
+end
```
KatBrandt, line 105: Overall, really great job with this class. Generally speaking, we want to maintain the integrity of our original data and any manipulation or changing we do, we can store that new info in our own variables. That way if there are every any discrepancies, we have our original data to go back to.

lib/registrant.rb
```diff
@@ -0,0 +1,21 @@
+class Registrant
+  attr_accessor :name, :age, :permit, :license_data
```
KatBrandt, line 2: I'm curious why these are all attr_accessors. Rule of thumb: keep everything an attr_reader until you absolutely must have an attr_accessor. We want to limit the ability of outside classes to make changes to these attributes.

lib/vehicle.rb
```diff
@@ -1,25 +1,51 @@
...
     Date.today.year - @year > 25
   end
 
   def electric_vehicle?
     @engine == :ev
   end
+
+  # Change this method name
+  def new_plate_type
+    if antique? == true
```
KatBrandt, line 39: `antique?` already returns true or false, so you don't need to do the comparison to true or false. can just have `if antique?`

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,107 @@
...
+    @vehicles.each do |vehicle|
+      make_and_model[vehicle.make].push(vehicle.model)
+    end
+    make_and_model
+  end
+
+# Filter out the boats and other vehicles
+  def sort_out_boats_and_trl(api)
+    api.select! do |vehicle|
+      if vehicle.values.include?("BOAT") == false && vehicle.values.include?("TRL") == false
```
KatBrandt, line 90: Same thing here, you don't need the `== false` because these values will either return nil or false and anything else would be "truthy".

Also, careful with the bang operator at the end of select as it permanently changes the original data and as I mentioned before, generally we want to maintain that original data. Nice work removing those boats and trls though!

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,107 @@
...
+    @vehicles = []
+  end
+
+  def create_vehicles(api)
+  # Helper method here
+    sort_out_boats_and_trl(api)
+    api.each do |vehicle|
+      if vehicle[:vin].nil?
+        vehicle[:vin] = vehicle.delete :vin_1_10
+      end
```
KatBrandt, line 14: Add a space after this end to visually separate your code blocks.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,107 @@
...
+    # The model variable finds the most popular model
+    model = ev_counter.key(ev_counter.values.sort[-1])
+    # Now iterate through the make_and_models to find the make of the model
+    make_and_models.each do |key, value|
+      if value.include?(model)
+        make = key
+      end
+    end
+    "The most popular make and model is the #{make} #{model}!"
+  end
```
KatBrandt, line 45: Great job with these analytics. Could also create an analytics class and have these be methods in that class. Remember that even classes should adhere to Single Responsibility Principle.

spec/facility_locations_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+      co_offices.facilities.each do |facility|
+        expect(facility).to be_an_instance_of(Facility)
+        expect(facility.name.nil?).to eq(false)
+        expect(facility.name).to be_an_instance_of(String)
+        expect(facility.address.nil?).to eq(false)
+        expect(facility.address).to be_an_instance_of(Array)
+        expect(facility.address[-2][:state]).to eq("CO")
+        expect(facility.phone.nil?).to eq(false)
+        expect(facility.phone).to be_an_instance_of(String)
+      end
```
KatBrandt, line 28: Nice job! Love that you used the enumerable in your test to test every element of the array.

spec/facility_locations_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+
+      new_york_facilities = DmvDataService.new.ny_dmv_office_locations
+      ny_offices = FacilityLocations.new
+      ny_offices.create_facilities(new_york_facilities)
+  
+      ny_offices.facilities.each do |facility|
+        expect(facility.hours.nil?).to eq(false)
+        expect(facility.hours).to be_an_instance_of(Hash)
+      end
```
KatBrandt, line 78: I'm not sure exactly what method you're testing here. You can do unit tests for the `hours` method in the `facility_locations` class and test just that method itself. This one is probably more of an integration test which is fine, but for a method like `hours` that is doing quite a bit, it can be helpful to have a unit test for just that method.

spec/facility_spec.rb
```diff
@@ -17,10 +25,148 @@
...
+    it 'changes the cost to register a vehicle depending on age' do
+      @facility.add_service('Vehicle Registration')
+      expect(@facility.collected_fees).to eq(0)
+  
+      @facility.register_vehicle(@cruz)
+      expect(@facility.collected_fees).to eq(100)
+      @facility.register_vehicle(@camaro)
+      expect(@facility.collected_fees).to eq(125)
+      @facility.register_vehicle(@bolt)
+      expect(@facility.collected_fees).to eq(325)
```
KatBrandt, line 69: Great job testing all the different paths of execution here.

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+
+  describe '#create_vehicles' do
+    it 'can create vehicles with an API' do
+      wa_ev_registrations = DmvDataService.new.wa_ev_registrations
+      @factory.create_vehicles(wa_ev_registrations)
+
+      expect(@factory.vehicles).to include(Vehicle)
+      expect(@factory.vehicles.length).to be > 100
+
+        @factory.vehicles.each do |car|
```
KatBrandt, line 22: Nice

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+      wa_ev_registrations = DmvDataService.new.wa_ev_registrations
+      @factory.create_vehicles(wa_ev_registrations)
+      @factory.most_popular_ev
+
+      expect(@factory.most_popular_ev).to eq("The most popular make and model is the NISSAN Leaf!")
+    end
+  end
+
+  describe 'EV Registrations - #registered_evs_for_model_year' do
+    it 'can count the number of registered evs for a model year' do
```
KatBrandt, line 57: Great job testing your analytics methods!

PR #379, opened 2023-10-24
Isaac Mitchell

tmitchellisaac:
Submission for DMV project

### DIFF
--- a/lib/facility.rb
+++ b/lib/facility.rb
@@ -2,14 +2,19 @@ require "date"
 
 
 class Facility
-  attr_accessor :name, :address, :phone, :services, :registered_vehicles, :collected_fees
-
-  # CurrentDate = Date.new(2023,10,19)
+  attr_accessor :name, 
+                :address, 
+                :phone, 
+                :services,
+                :hours, 
+                :registered_vehicles, 
+                :collected_fees
 
   def initialize(input)
     @name = input[:name]
     @address = input[:address]
     @phone = input[:phone]
+    @hours = input[:hours]
     @services = []
     @registered_vehicles = []
     @collected_fees = 0
--- a/lib/facility.rb
+++ b/lib/facility.rb
@@ -1,14 +1,40 @@
+require "date"
+
+
 class Facility
-  attr_reader :name, :address, :phone, :services
+  attr_accessor :name, :address, :phone, :services, :registered_vehicles, :collected_fees
+
+  CurrentDate = Date.new(2023,10,19)
 
   def initialize(input)
     @name = input[:name]
     @address = input[:address]
     @phone = input[:phone]
     @services = []
+    @registered_vehicles = []
+    @collected_fees = 0
   end
 
   def add_service(service)
     @services << service
+  end
+
+  def register_vehicle(vehicle)
+    vehicle.registration_date = CurrentDate
+    if vehicle.registered == false
+      if vehicle.engine == :ev
+        @collected_fees += 200
+        vehicle.plate_type =:ev
+      elsif vehicle.antique? == true
+        @collected_fees += 25
+        vehicle.plate_type = :antique
+      else
+        @collected_fees += 100
+        vehicle.plate_type = :regular
+      end
+    @registered_vehicles << vehicle
+    vehicle.registered = true
+    vehicle
+    end
   end
 end
--- a/lib/facility.rb
+++ b/lib/facility.rb
@@ -11,6 +11,13 @@ class Facility
                 :collected_fees,
                 :holidays_closed
 
+
+  LIST_OF_SERVICES = ["New Drivers License", 
+                      "Renew Drivers License", 
+                      "Vehicle Registration", 
+                      "Written Test",
+                      "Road Test"]
+
   def initialize(input)
     @name = input[:name]
     @address = input[:address]
@@ -25,8 +32,10 @@ class Facility
   end
 
   def add_service(service)
-    @services << service
-    @services
+    if LIST_OF_SERVICES.include?(service)
+      @services << service
+      @services
+    end
   end
 
   def register_vehicle(vehicle)
@@ -51,7 +60,7 @@ class Facility
   def administer_written_test(registrant)
     if (@services.include?("Written Test")) && (registrant.age >= (16)) && (registrant.permit? == (true))
       registrant.license_data[:written] = true
-          end
+    end
     registrant.license_data[:written]
   end
 
@@ -63,7 +72,7 @@ class Facility
   end
 
   def renew_drivers_license(registrant)
-    if (@services.include?("Renew License")) && (registrant.license_data[:license] == true)
+    if (@services.include?("Renew Drivers License")) && (registrant.license_data[:license] == true)
       registrant.license_data[:renewed] = true
     end
     registrant.license_data[:renewed]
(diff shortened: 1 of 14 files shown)


lib/facility.rb
```diff
@@ -1,14 +1,81 @@
...
 class Facility
-  attr_reader :name, :address, :phone, :services
+  attr_accessor :name, 
+                :address, 
+                :phone, 
+                :services,
+                :hours, 
+                :registered_vehicles, 
+                :collected_fees,
+                :holidays_closed
```
megstang, line 12: Similarly here.. it doesn't look like you are trying to update these attributes outside of this class, so I thin kthey can all be set as readers.

lib/facility.rb
```diff
@@ -1,14 +1,81 @@
...
+        vehicle.plate_type = :antique
+      else
+        @collected_fees += 100
+        vehicle.plate_type = :regular
+      end
+    @registered_vehicles << vehicle
+    vehicle.registered = true
+    vehicle
+    end
+  end
```
megstang, line 58: We will want to test that the facility has the "Vehicle Registration" service before we do anything to that vehicle object. The rest of the logic in this method looks good though!

lib/facility.rb
```diff
@@ -1,14 +1,81 @@
...
+    registrant.license_data[:license]
+  end
+
+  def renew_drivers_license(registrant)
+    if (@services.include?("Renew Drivers License")) && (registrant.license_data[:license] == true)
+      registrant.license_data[:renewed] = true
+    end
+    registrant.license_data[:renewed]
+  end
+
```
megstang, line 80: This logic all looks good. You don't necessarily need to have `== true ` in any of the places where you have it, because those values will automatically return `true` or `false` depending on what's stored as their values.

lib/facility_factory.rb
```diff
@@ -0,0 +1,54 @@
...
+
+class FacilityFactory
+
+  def initialize
+  end
+
+  def create_facility(input_api)
+    co_total_facility_array = []
+    ny_total_facility_array = []
+    mo_total_facility_array = []
```
megstang, line 11: You don't necessarily need separate arrays for each data set that could come in. This method will get run for only one set of data at a time, so just having one placeholder for `facilities` should be plenty. You'd just shovel all new facilities made into `facilities` rather than a specific array.

lib/facility_factory.rb
```diff
@@ -0,0 +1,54 @@
...
+      input_api.each do |facility_raw_info|
+        input_hash = {}
+        input_hash[:name] = facility_raw_info[:dmv_office]
+        input_hash[:address] = "#{facility_raw_info[:address_li]}, #{facility_raw_info[:address__1]}, #{facility_raw_info[:city]}, #{facility_raw_info[:state]} #{facility_raw_info[:zip]}"
+        input_hash[:phone] = facility_raw_info[:phone]
+        input_hash[:hours] = facility_raw_info[:hours]
+        co_total_facility_array << Facility.new(input_hash)
+      end
+    return co_total_facility_array
+
```
megstang, line 23: I like that you're explicitly returning the array after it's been filled to avoid your method continuing to run.

lib/facility_factory.rb
```diff
@@ -0,0 +1,54 @@
...
+      input_api.each do |facility_raw_info|
+        input_hash = {}
+        input_hash[:name] = facility_raw_info[:name]
+        input_hash[:address] = "#{facility_raw_info[:address1]}, #{facility_raw_info[:city]}, #{facility_raw_info[:state]} #{facility_raw_info[:zipcode]}"
+        input_hash[:phone] = facility_raw_info[:phone]
+        input_hash[:hours] = facility_raw_info[:daysopen]
+        input_hash[:holidays_closed] = facility_raw_info[:holidaysclosed]
+        mo_total_facility_array << Facility.new(input_hash)
+      end
+    return mo_total_facility_array
```
megstang, line 51: Great job on this. You could possibly create helper methods that would help you to format specific state's data. That would help to slim up this method a bit.

lib/registrant.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def permit?
+    @permit == true
+  end
+  
+  def earn_permit
+    @permit = true
+  end
+
```
megstang, line 21: 👍 This class looks good. I would, however, remove `:permit` from the accessors since we don't need to ever call `.permit` on a Registrant object. On that subject also, I think only `license_data` is the attribute that should be an accessor since it can be updated from other classes. All others (name and age) can just be set as readers.

lib/vehicle.rb
```diff
@@ -6,13 +6,17 @@ class Vehicle
               :make,
               :model,
               :engine
+  attr_accessor :plate_type, :registration_date, :registered
```
megstang, line 9: Good call here on making these items accessor's since you'll need to change them later.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,89 @@
...
+      make_model = []
+      make_model << "#{vehicle.make}, #{vehicle.model}"
+      all_makes_models << make_model
+    end
+        
+    models_count = Hash.new(0)
+    all_makes_models.each do |model| 
+      models_count[model] += 1
+    end
+    @most_popular_make_model = (models_count.sort_by { |model,number| number}.last[0]).join
```
megstang, line 52: Really fun to see this! You could pull this into a helper method also rather than having it be a part of your create method.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,89 @@
...
+    counties_count = Hash.new(0)
+    total_counties.each do |county|
+      counties_count[county] += 1
+    end
+
+    @most_common_county = counties_count.sort_by {|county, number| number}.last[0]
+    
+    #last line of method return:
+    created_vehicles
+  end
```
megstang, line 83: these are all really great too! Fun approaches to solving this, I like it!

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+  it "has capability to use NY vehicle database" do
+    factory_2 = VehicleFactory.new
+    new_york_vehicle_registrations = DmvDataService.new.ny_state_registrations
+    factory_2.create_vehicles(new_york_vehicle_registrations)
+    
+    expect(factory_2.most_common_county).to eq("ERIE")
+    expect(factory_2.most_common_year).to eq(1949)
+    expect(factory_2.year_counter(1936)).to eq(9)
+    expect(factory_2.year_counter(1999)).to eq(4)
+  end
```
megstang, line 52: These tests would be a great time to use fixture data. We haven't talked about this yet, but will briefly touch on it in the mocks/stubs class next week. Since the data that is returned from the API is going to change eventually, it's very possible that these tests will no longer pass. After about a year, it's likely that the most_common_year  or year_counter will change. So, for that reason, we could create our own subset of data that won't change, just to be sure that our methods work. What I'd do for that is create a `fixtures` folder in your spec folder, and in there, house a few files of data. Then, instead of using the `DmvDataService.new.wa_ev_registrations` data, you could use the data from the fixture file.

megstang (commented):
@tmitchellisaac I left some comments for you to review. Take a look and let me know if you have any questions. I'll follow up by end of day today with a bit more feedback.

PR #381, opened 2023-10-24
DMV Project Submission

s2an:
The actual Project Submission Repository


lib/facility.rb
```diff
@@ -1,14 +1,69 @@
...
     @services = []
+    @registered_vehicles = []
+    @collected_fees = 0
+  end
+
+  def add_service(service)
+    unless @services.include?(service)
+      @services << service
+    end
+  end
```
megstang, line 24: Nice! I like this edge case to check if it's included first before adding the service.

lib/facility.rb
```diff
@@ -1,14 +1,69 @@
...
+          @collected_fees += 200
+        elsif model.antique?
+          @collected_fees += 25
+        else
+          @collected_fees += 100
+        end
+      end
+    end
+    registered_vehicles
   end
```
megstang, line 42: Line 30 will not set that vehicles registration date, but rather assign today's date into a variable called `registration_date`. I see in your vehicle class that you created a method called `set_registration_date`. You could use that here and just say `model.set_registration_date`.

megstang: You could also do something similar for `plate_type`.

lib/facility.rb
```diff
@@ -1,14 +1,69 @@
...
+    end
   end
+
+  def administer_road_test(registrant)
+    if registrant.license_data[:written] == true
+      registrant.license_data[:license] = true
+    else
+      false
+    end
+  end
```
megstang, line 59: We'd also want to test here that the Facility includes "Road Test" in it's services

lib/facility.rb
```diff
@@ -1,14 +1,69 @@
...
+
+  def renew_drivers_license(registrant)
+    if registrant.license_data[:written] == true && registrant.license_data[:license] == true
+      registrant.license_data[:renewed] = true
+    else
+      false
+    end
+  end
```
megstang, line 67: we'd also want to check here that the facility includes "Renew License" in it's services.

lib/registrant.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  def permit?
+    @permit
+  end
+
+  def earn_permit
+    @permit = true
+  end
+
+end
```
megstang, line 19: 👍 This looks good, however there is no need to list `:permit` as an attr_Reader since we never need to call `.permit` on a `Registrant` object.

lib/vehicle.rb
```diff
@@ -22,4 +26,21 @@ def antique?
...
+     @plate_type = if antique?
+        :antique
+      elsif electric_vehicle?
+        :ev
+      else
+        :regular
+      end
+    end
+  end
 end
```
megstang, line 46: Looping back around to talk about these methods ( I had commented on them earlier). These methods will only update the plate type and registration date once they were called. The functionality we were looking for was that the plate type and the registration date were set AS SOON as the car was registered.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,59 @@
...
+    @created_vehicles = state_registration.map do |vehicle|
+      Vehicle.new(
+        vin: vehicle[:vin_1_10], 
+        year: vehicle[:model_year], 
+        make: vehicle[:make], 
+        model: vehicle[:model], 
+        engine: :ev)
+    end 
+  end
+
```
megstang, line 18: your logic in this method looks good. There was no need for an attribute, since we only cared about teh return value of this `create_vehicles` method. This return value might change depending on what the API returns, so that's why we don't need to store the state in an attribute, and we can focus on just returning the vehicles that were created when the `create_vehicles()` method was called

lib/facility_factory.rb
```diff
@@ -0,0 +1,20 @@
...
+      Facility.new(
+        name: facility[:dmv_office] || facility[:office_name] || facility[:name],
+        address: [ facility[:address_li] || facility[:address__1] || facility[:street_address_line_1] || facility[:address1],
+        facility[:city],
+        facility[:state],
+        facility[:zip] || facility[:zip_code] || facility[:zipcode] ].join(" "),
+        phone: facility[:phone] || facility[:public_phone_number],
+        services: facility[:services_p])
+    end
+  end
```
megstang, line 19: You did not need to add the services to the facility because our Facility class has that functionality already. I'd also spend a bit more time taking a look at how you're grabbing the address. I'd suggest making different methods that would pull out the address data depending on what state the data set was from.

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  
+  it 'creates a MO facility' do
+    factory = FacilityFactory.new
+    missouri_facilities = DmvDataService.new.mo_dmv_office_locations
+    mo_dmv_office_locations = DmvDataService.new.mo_dmv_office_locations
+
+    expect(factory.create_facilities(mo_dmv_office_locations)).to be_an_instance_of(Array)
+  end
+
+end
```
megstang, line 41: We also want to get into that array that the method should be returning, and check that it is truly creating Vehicle objects as you'd expect it to. For starters, we could look at the first element of the array that was returned and make sure it was a vehicle object, and that each of it's attributes were not nil.

spec/facility_spec.rb
```diff
@@ -1,26 +1,198 @@
...
+  it 'can renew a license after that service is added' do
+    expect(@facility_1.renew_drivers_license(@registrant_1)).to eq(false)
+    
+    @facility_1.add_service('Written Test') #I don't like that I need to fake add these services
+    @facility_1.add_service('Road Test') #I don't like that I need to fake add these services, and readminister the 
+    @facility_1.add_service('Renew License')
+    @facility_1.administer_written_test(@registrant_1)
+    @facility_1.administer_road_test(@registrant_1)
+    
+    expect(@facility_1.services).to eq(['Written Test', 'Road Test', 'Renew License'])
```
megstang, line 174: You've already tested this behavior on line 47

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    
+    expect(wa_ev_registrations = DmvDataService.new.wa_ev_registrations).to be_an_instance_of(Array)
+  end
+
+  it 'takes the registered vehicles and arrays it into a Vehicle' do
+    factory = VehicleFactory.new
+    wa_ev_registrations = DmvDataService.new.wa_ev_registrations
+
+    expect(factory.create_vehicles(wa_ev_registrations)).to be_an_instance_of(Array)
+  end
```
megstang, line 23: Apply the same feedback I gave on the FAcilityFactory spec to this spec too. It'd be best to go into those arrays and test what they're made up of.

megstang (commented):
@s2an I left some comments for you to take a look over. Please let me know if you have any questions. I'll follow up in slack within a day to give you more feedback.

PR #382, opened 2023-10-24
DMV Project Submission

megroth2:
Completed Iterations 1, 2, and 3.


lib/facility.rb
```diff
@@ -1,14 +1,67 @@
...
+        @facility.collected_fees += 25
+      elsif @vehicle.electric_vehicle?
+        @vehicle.plate_type = :ev
+        @facility.collected_fees += 200
+      else
+        @vehicle.plate_type = :regular
+        @facility.collected_fees += 100
+      end
+    end
+  end
```
megstang, line 38: Referring to lines 20 and 21, I see you doing this a lot in your methods -- reassigning values to be instance variables. Unless those values need to be called outside of that method, there's no need to make them instance variables. You could redo this entire method to look something like: 

```ruby 
def register_vehicle(vehicle)

      if vehicle.antique?
        vehicle.plate_type = :antique
        @collected_fees += 25
        vehicle.plate_type = :ev
        @collected_fees += 200
      else
        vehicle.plate_type = :regular
        @collected_fees += 100
      end
    end
  end 
```

lib/facility.rb
```diff
@@ -1,14 +1,67 @@
...
+  def renew_drivers_license(registrant)
+    @facility = self
+    @registrant = registrant
+
+    @registrant.license_data[:renewed] = true if @facility.services.include?('Renew License') && @registrant.license_data[:license] == true
+    
+    @registrant.license_data[:renewed]
+  end
+
 end
```
megstang, line 67: These all look great! Apply the same feedback I gave previously on the `register_vehicle` method.

lib/registrant.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def permit?
+    @permit
+  end
+
+  def earn_permit
+    @permit = true
+  end
+
+end
```
megstang, line 23: 👍 This all looks great!

lib/vehicle.rb
```diff
@@ -1,25 +1,24 @@
...
   def antique?
-    Date.today.year - @year > 25
+    Date.today.year - @year >= 25
   end
 
   def electric_vehicle?
     @engine == :ev
   end
+
 end
```
megstang, line 24: We really only want to create attr_accessor's for variables if we want those variables to be able to get updated outside of the class. The only attributes that I think might get updated outside of this class are the `@registration_date` and `@plate_type`, so those will be the only attr_accessor's we'll need. The others can be attr_readers. That said, I don't think you have an attribute called `vehicle_details`, so you could get rid of that (it would probably return `nil` at this moment if it were to get called.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,22 @@
...
+      model = registration[:model]
+
+      new_vehicle = Vehicle.new({vin: vin, year: year, make: make, model: model, engine: :ev})
+
+      vehicles << new_vehicle
+    end
+    vehicles
+  end
+  
+end
```
megstang, line 22: Awesome work here. No need for the `initialize` method though. We only need to create an `initialize` method if we want to set attributes upon initialization of an object. Since we don't need to store any state on a VehicleFactory object once it's created, we don't even need to include this method.

lib/facility_factory.rb
```diff
@@ -0,0 +1,52 @@
...
+      phone = location[:phone]
+
+      new_facility = Facility.new({name: name, address: address, phone: phone})
+
+      facilities << new_facility
+    end
+    facilities
+  end
+  
+end
```
megstang, line 52: Really awesome work here! I don't have much at all to add. I guess you could have some fun with refactoring and use other enums... maybe `map` would be good!

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    expect(factory.create_vehicles(wa_ev_registrations)).to be_instance_of(Array)
+    expect(factory.create_vehicles(wa_ev_registrations).first).to be_instance_of(Vehicle)
+    expect(factory.create_vehicles(wa_ev_registrations).first.vin).not_to be(nil)
+    expect(factory.create_vehicles(wa_ev_registrations).first.year).not_to be(nil)
+    expect(factory.create_vehicles(wa_ev_registrations).first.make).not_to be(nil)
+    expect(factory.create_vehicles(wa_ev_registrations).first.model).not_to be(nil)
+  end
+
+end
```
megstang, line 23: Great work with testing here. This one was a bit tricky to figure out how to test it. There's a lot of different approaches you could take on it, but I really like that you're making sure that it's an array, that it's made of Vehicle objects, and that the data isn't empty.

megstang (commented):
@megroth2 I left some comments for you to look over. Please let me know if you have any questions! I'll follow up within a day to get you more feedback over slack.

lib/facility.rb
megstang: In your submission, you asked about how you could avoid nested if statements in this method. It is totally okay to have nested if's in a method, but something you could do is create a helper method that would handle the functionality for assigning plate_type and collecting fees.

```ruby
def register_vehicle(vehicle)
       assign_plate_type(vehicle)
       collect_fees(vehicle)
    end
  end 

def assign_plate_type(vehicle)
  if vehicle.antique?
        vehicle.plate_type = :antique
        vehicle.plate_type = :ev
      else
        vehicle.plate_type = :regular
      end
end 

def collect_fees(vehicle)
  if vehicle.antique?
        @collected_fees += 25
        @collected_fees += 200
      else
        @collected_fees += 100
      end
end
```

PR #383, opened 2023-10-24
Iteration 2

mel-langhoff:


lib/dmv.rb
```diff
@@ -9,7 +10,7 @@ def add_facility(facility)
   end
 
   def facilities_offering_service(service)
-    @facilities.find do |facility|
+    @facilities.select do |facility|
       facility.services.include?(service)
     end
   end
```
megstang, line 16: good use of select!

lib/facility.rb
```diff
@@ -1,14 +1,21 @@
 class Facility
-  attr_reader :name, :address, :phone, :services
+  attr_reader :name, :address, :phone, :services, :vehicle, :registered_vehicle
```
megstang, line 2: what is this attr_reader for :vehicle for? I don't see anywhere where you need to call `.vehicle` on a Facility object, so you can remove that.

lib/registrant.rb
```diff
@@ -0,0 +1,31 @@
+class Registrant
+    attr_reader :age,
+                :license_data,
+                :name,
+                :permit
+
```
megstang, line 6: I don't think you need `permit` listed here since you never need to call `.permit` on a Registrant object.

lib/registrant.rb
```diff
@@ -0,0 +1,31 @@
...
+            renewed: false
+            }
+        @permit = permit
+               
+    end
+
+    def permit?
+        @permit == @permit
+    end
+
```
megstang, line 22: This will always be true since youre checking if one thing is equal to itself ... which will always be true. You can just return `@permit` because that should store either true or false.

```ruby 
def permit? 
 @permit 
end 
```

lib/registrant.rb
```diff
@@ -0,0 +1,31 @@
...
+    def earn_permit
+        if @permit == false
+            @permit != false
+        end
+    end
```
megstang, line 27: This method is not going to change the permit value. You'll need to reassign the value of `@permit` if the condition on line 24 evaluates to `true`. To do that, on line 25, you'll need it to say `@permit = true`

spec/facility_spec.rb
```diff
@@ -23,4 +29,11 @@
       expect(@facility.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
     end
   end
+
+  describe "#register_vehicle" do
+    it "can register a vehicle" do
+      expect(@facility.registered_vehicle).to eq(Vehicle.new)
+    end
```
megstang, line 36: We don't want this method to return a vehicle object, but rather an array of vehicle objects. In this case though, we haven't added any vehicles, so it should return an empty array.

megstang (commented):
@mel-langhoff I left some comments for you to take a look over. Please let me know if you have any questions. I'll follow up on slack with more feedback within a day.

PR #384, opened 2023-10-24
Get all initial tests passing

Alaina-Noel:
This is a pull request


Alaina-Noel:
Yea this was me doing iteration one of the project on my own for fun (haven't written ruby since January!) And then pushing to the forked repo instead of my own. Old habits die hard apparently.

PR #386, opened 2023-10-24
Logan

LoganFinnegan:


lib/dmv.rb
```diff
@@ -9,7 +9,7 @@ def add_facility(facility)
   end
 
   def facilities_offering_service(service)
-    @facilities.find do |facility|
+    @facilities.select do |facility|
       facility.services.include?(service)
     end
   end
```
megstang, line 15: This is a great use of `select`!

lib/facility.rb
```diff
@@ -1,14 +1,42 @@
...
+  def determine_fee(vehicle)
+    if vehicle.plate_type == :antique
+      @collected_fees += 25
+    elsif vehicle.plate_type == :ev
+      @collected_fees += 200
+    else
+      @collected_fees += 100
+    end 
+  end
 end
```
megstang, line 42: I really like that you broke this `register_vehicle` method out into a helper method. This slims up your code and helps you stick to one responsibility per method.

lib/registrant.rb
```diff
@@ -0,0 +1,20 @@
...
+    end
+
+    def permit?
+        @permit
+    end
+
+    def earn_permit
+      @permit = true
+    end
+end
```
megstang, line 20: 👍 This looks good, but I would change a few things about this class: 

1. We only want to use `attr_accessor`s if we want that attribute to be able to be changed by another class. Since you don't need to do that, you could set all of these as `attr_reader`s which would only allow these attributes to be read outside of this class, but not changed.
2. There is no need for a `permit` method. You created one by listing the `:permit` as an attr_accessor. If you look in the interaction pattern, we never need to call `.permit` on a `Registrant` instance, so for that reason we don't need it listed as an attr_accessor. BUT, we'd still want to keep it as an attribute in our initialize method since our `permit?` method relies on that state.

lib/vehicle.rb
```diff
@@ -22,4 +24,18 @@ def antique?
...
+
+  def plate_type
+    if Date.today.year - @year >= 25
+      :antique
+    elsif @engine == :ev
+      :ev
+    else 
+      :regular
+    end
+  end
```
megstang, line 40: 👍

spec/facility_spec.rb
```diff
@@ -2,25 +2,82 @@
...
+      expect(@facility_1.registered_vehicles).to eq([])
+    end
+  end
+
+  describe '#register_vehicle' do
+    it 'can register a vehicle' do
+      @facility_1.add_service('Vehicle Registration')
+
+      expect(@facility_1.register_vehicle(@cruz)).to eq([@cruz])
     end
```
megstang, line 43: We could also have a test where we don't add the service, and test the return value of `register_vehicle`. In your code, it looks like it should return `nil` if the service isn't listed. So that's what you could test.

megstang: In a similar vein, the interaction pattern shows that if you called `registration_date` on a vehicle that has yet to be registered, that it should return `nil`. You should have a test for that too.

megstang (commented):
@LoganFinnegan I left a few comments for you to look over. Let me know if you have any questions. I will follow up in slack within the next two days with more feedback.

PR #387, opened 2023-10-24
Began work on administer_written_test

DHowell1150:
- Added registrant_3 in registrant_spec to all tests.
- Began "can administer test" in registrant_spec
- Began `administer_written_test` in Facility.rb
- This method does not work so I commented it out both the method and the test for it.


lib/dmv.rb
```diff
@@ -9,7 +9,7 @@ def add_facility(facility)
   end
 
   def facilities_offering_service(service)
-    @facilities.find do |facility|
+    @facilities.find_all do |facility|
       facility.services.include?(service)
     end
```
megstang, line 14: Great use of `find_all`!

lib/facility.rb
```diff
@@ -1,14 +1,42 @@
...
-  def initialize(name, address, phone)
-    @name = name
-    @address = address
-    @phone = phone
+  def initialize(facility)
+    @name = facility[:name]
+    @address = facility[:address]
+    @phone = facility[:phone]
     @services = []
+    @facility = []
```
megstang, line 9: we shouldn't need this attribute, nor will we need an attr_reader for it. I say this because I don't see anywhere in the interaction pattern where we call `.facility` on a `Facility` object.

lib/facility.rb
```diff
@@ -1,14 +1,42 @@
...
+      @collected_fees += 100
+      vehicle.plate_type = :regular
+    end
+  end
+
+  # def administer_written_test
+  #   permit? == true
+  #   age >= 16
+  # end
 end
```
megstang, line 38: You're close with this method! I know you mentioned this was a really tough one for you. We only want to register a vehicle IF the list of services contains "Vehicle Registration". So, we'll want to add some type of conditional into this method such that IF "Vehicle Registration" is in the list of services, we do everything between lines 19 and 31, otherwise, we do nothing.

lib/vehicle.rb
```diff
@@ -1,18 +1,23 @@
...
 
 class Vehicle
+  attr_accessor :plate_type
   attr_reader :vin,
               :year,
               :make,
               :model,
-              :engine
+              :engine,
+              :registration_date
```
megstang, line 11: Good call on using an `attr_accessor` for `:plate_type` since you needed your `Facility` class to be able to update that attribute. You could have also done the same thing for `registration_date`, and then you wouldn't have needed to make the `new_registration_date` method down below.

lib/vehicle.rb
```diff
@@ -22,4 +27,13 @@ def antique?
   def electric_vehicle?
     @engine == :ev
   end
+
+  def registered?    
+    !!plate_type && !!registration_date
+  end
```
megstang, line 33: I'm curious about this method. I didn't see anywhere in the interaction pattern where you needed to create this method, and it doesn't look like it's being used as a helper method anywhere else. You could get rid of it. 

That said, as a general practice, it doesn't make too much sense to throw a double bang (!!) in front of something, since that essentially negates the negation (like a double negative), so it just makes it what it was originally.

lib/vehicle.rb
```diff
@@ -22,4 +27,13 @@ def antique?
...
+  
+  def new_registration_date
+    @registration_date = (Time.now)
+  end
```
megstang, line 37: Since we're working with Dates, it would be good to use a `Date` object. Something like `Date.today` would work great here.

spec/facility_spec.rb
```diff
@@ -1,26 +1,73 @@
...
+      expect(@facility_1.registered_vehicles).to eq([])
+      expect(@facility_1.collected_fees).to eq(0)
+      expect(@cruz.registered?).to eq(false)
+
+      @facility_1.register_vehicle(@cruz)
+      expect(@cruz.plate_type).to eq(:regular)
+      expect(@facility_1.collected_fees).to eq(100)
+      expect(@cruz.registration_date).to be_an_instance_of(Time)
+      expect(@cruz.registered?).to eq(true)
+    end
```
megstang, line 53: we might also want to test that the registration_date is nil until `registered_vehicles` is called. So maybe instead of line 46, you could have `expect(@cruz.registration_date).to eq(nil)`

megstang: Nevermind, I see in your `vehicle_spec` that you test that  attribute. :)

spec/vehicle_spec.rb
```diff
@@ -15,6 +16,22 @@
...
+      expect(@bolt.engine).to eq(:ev)
+      expect(@bolt.registration_date).to eq(nil)
+
+      expect(@camaro).to be_an_instance_of(Vehicle)
+      expect(@camaro.vin).to eq('1a2b3c4d5e6f')
+      expect(@camaro.year).to eq(1969)
+      expect(@camaro.make).to eq('Chevrolet')
+      expect(@camaro.model).to eq('Camaro')
+      expect(@camaro.engine).to eq(:ice)
+      expect(@camaro.registration_date).to eq(nil)
```
megstang, line 34: I would also test in your `vehicile_spec` your `new_registration_date` method since that's a method that you defined in the Vehicle class.

megstang (commented):
@DHowell1150 I left some comments for you to look over. Please let me know if you have any questions. I'll follow up in slack within the next two days with more feedback.

lib/dmv.rb
DHowell1150: Thanks!  I might have had Kat hold my hand into that one ;)

PR #388, opened 2023-10-24
feat/dylan-perry-dmv

Dylan-Perry:


lib/facility.rb
```diff
@@ -1,14 +1,72 @@
 class Facility
-  attr_reader :name, :address, :phone, :services
+  attr_accessor :name, :address, :phone, :services, :registered_vehicles, :collected_fees, :hours, :closed_holidays
```
KatBrandt, line 2: Wondering why you made all of these accessors. My general rule of thumb is to keep them all as readers, then IF you need any as accessors, pull those out one by one. We want to restrict access to outside classes being able to change these values as much as possible.

lib/facility.rb
```diff
@@ -1,14 +1,72 @@
...
+      if vehicle.antique?
+        @collected_fees += 25
+        vehicle.plate_type = :antique
+      elsif vehicle.electric_vehicle?
+        @collected_fees += 200
+        vehicle.plate_type = :ev
+      else
+        @collected_fees += 100
+        vehicle.plate_type = :regular
+      end
```
KatBrandt, line 31: This works and looks good overall. Ideally though, we want to avoid nested if statements when possible (maybe not always possible), simply because they take up a lot of cpu power.

lib/facility.rb
```diff
@@ -1,14 +1,72 @@
...
+  def administer_written_test(registrant)
+    if @services.include?("New Drivers License")
+      if !registrant.permit? || registrant.age < 16
+        "Error: Written tests can only be administered to registrants with a permit, and who are at least 16 years of age."
+      else
+        registrant.license_data[:written] = true
+      end
+    else
+      "Error: #{@name} does not offer Written Test services."
+    end
```
KatBrandt, line 46: Same thing here. Could use more logical operators to flatten this out. But overall, great job with the logic.

lib/facility_list.rb
```diff
@@ -0,0 +1,40 @@
...
+
+            elsif facility_api[:state] == "MO"
+                address_1, city, state, zip = facility_api.values_at(:address1, :city, :state, :zipcode)
+                facility_ours.address = "#{address_1} #{city} #{state} #{zip}"
+                facility_ours.address.gsub(/  /, " ")
+                facility_ours.hours = facility_api[:daysopen]
+                facility_ours.closed_holidays = facility_api[:holidaysclosed]
+
+            end
+            @facilities << facility_ours
```
KatBrandt, line 37: I like that you put these into an instance variable so we only have to run all this parsing once. You could refactor this by abstracting each state into it's own method, then having your if/else statement simply call the method for that state. I see potential for lots of helper methods in this big method! Great job overall!

spec/facility_list_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+        list = FacilityList.new
+        co_facilities_data = DmvDataService.new.co_dmv_office_locations
+        list.parse_facilities(co_facilities_data)
+
+        expect(list.facilities.first).to be_a(Facility)
+        expect(list.facilities.first.name).not_to be_nil
+        expect(list.facilities.first.address).not_to be_nil
+        expect(list.facilities.first.phone).not_to be_nil
+        expect(list.facilities.first.services).not_to be_nil
+        expect(list.facilities.first.hours).not_to be_nil
```
KatBrandt, line 22: Yup, this is good. This is really all we can test with an Api where we don't know the data.

spec/facility_spec.rb
```diff
@@ -23,4 +26,119 @@
...
+
+      expect(@facility.collected_fees).to eq(300)
+
+      @facility.register_vehicle(@camaro)
+
+      expect(@facility.collected_fees).to eq(325)
+
+      expect(@cruz.plate_type).to eq(:regular)
+      expect(@bolt.plate_type).to eq(:ev)
+      expect(@camaro.plate_type).to eq(:antique)
```
KatBrandt, line 55: Great test, testing all the possible control flows in the method.

spec/facility_spec.rb
```diff
@@ -23,4 +26,119 @@
...
+    end
+
+    it "errors out if facility does not offer vehicle registration" do
+      expect(@facility.register_vehicle(@cruz)).to eq("Error: DMV Tremont Branch does not offer Vehicle Registration services.")
```
KatBrandt, line 59: Awesome. Love that you provided an error message.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+        end
+
+        it "instantiates variables and defaults @permit to false" do
+            registrant_1 = Registrant.new("Bruce", 18, true)
+            registrant_2 = Registrant.new("Penny", 15)
+
+            expect(registrant_1.name).to eq("Bruce")
+            expect(registrant_1.age).to eq(18)
+            expect(registrant_1.permit).to eq(true)
+            expect(registrant_1.license_data).to be_a Hash
```
KatBrandt, line 19: Could go a step further here and instead of just testing it's a Hash, test explicitly what that hash should look like.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+        end
+    end
+
+    describe "#permit?" do
+        it "returns true/false for registrant with/without permit" do
+            registrant_1 = Registrant.new("Bruce", 18, true)
+            registrant_2 = Registrant.new("Penny", 15)
+
+            expect(registrant_1.permit?).to eq(true)
+            expect(registrant_2.permit?).to eq(false)
```
KatBrandt, line 30: Good, whenever we have that default value we want to test instantiation with and without it. Nice work.

KatBrandt (commented):
Great job!

PR #389, opened 2023-10-25
Barry Allahyar

BarryA:


lib/dmv.rb
```diff
@@ -9,7 +10,7 @@ def add_facility(facility)
   end
 
   def facilities_offering_service(service)
-    @facilities.find do |facility|
+    @facilities.select do |facility|
       facility.services.include?(service)
     end
   end
```
megstang, line 16: 👍 good use of `select`

lib/registrant.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def permit?
+    @permit
+  end
+
+  def earn_permit
+    @permit = true
+  end
+end
```
megstang, line 24: I don't believe any of these attributes need to be listed as attr_accessors because we don't need to change their values outside of this class. Also, I don't think you need any type of reader or accessor for `permit` since you never need to call `.permit` on a REgistrant object. 

Make sure to indent lines 11-14

megstang (commented):
I left a few comments on your code for you to take a look over. Let me know if you have any questions on anything I wrote. I will follow up in slack later today to get more feedback.

PR #391, opened 2023-10-25
Quin Nordmark

n0rdie:
Did not finish

### DIFF
--- a/lib/dmv.rb
+++ b/lib/dmv.rb
@@ -1,11 +1,12 @@
 class Dmv
+  attr_reader :facilities
 
   def initialize
     @facilities = []
   end
 
   def add_facility(facility)
-    @facilities << facility
+    @facilities.append(facility)
   end
 
   def facilities_offering_service(service)
--- a/lib/dmv.rb
+++ b/lib/dmv.rb
@@ -10,10 +10,10 @@ class Dmv
   end
 
   def facilities_offering_service(service)
-    finalized = false
+    finalized = []
     @facilities.each do |facility|
       if facility.services.include?(service)
-        finalized = true
+        finalized.append(facility)
       end
     end
     finalized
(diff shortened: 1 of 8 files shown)


lib/vehicle.rb
```diff
@@ -5,14 +5,24 @@ class Vehicle
...
+              :plate_type
 
   def initialize(vehicle_details)
     @vin = vehicle_details[:vin]
     @year = vehicle_details[:year]
     @make = vehicle_details[:make]
     @model = vehicle_details[:model]
     @engine = vehicle_details[:engine]
+    @registration_date = nil
+    @plate_type = :regular
```
KatBrandt, line 19: This works, but I would like to see this abstracted into it's own method. We want to keep the initialize as clean as possible.

spec/dmv_spec.rb
```diff
@@ -37,6 +37,7 @@
       @dmv.add_facility(@facility_2)
       @dmv.add_facility(@facility_3)
 
+      p @dmv.facilities_offering_service('Road Test')
```
KatBrandt, line 40: You'll want to remove `p` and `puts` from your tests in the future when you're all wrapped up with a project. :)

lib/dmv.rb
```diff
@@ -1,16 +1,21 @@
...
   def facilities_offering_service(service)
-    @facilities.find do |facility|
-      facility.services.include?(service)
+    finalized = []
+    @facilities.each do |facility|
+      if facility.services.include?(service)
+        finalized.append(facility)
+      end
     end
+    finalized
```
KatBrandt, line 19: This does work, but all we had to do to fix the broken code was change the `find` to a `find_all` or a `select`. I encourage you to review the enumerables lesson to start using more enumerables other than .each.

KatBrandt (commented):
Good start Quin.

PR #412, opened 2023-12-12
Joey Reyes DMV Project Iteration 1,2,3 complete 

JRIV-10:
My submission for the DMV project all iterations up to 3 are completed.

### DIFF
--- a/lib/dmv.rb
+++ b/lib/dmv.rb
@@ -1,5 +1,7 @@
 class Dmv
-
+  attr_reader :name, :address, :phone
+  attr_accessor :facilities
+  
   def initialize
     @facilities = []
   end
@@ -9,7 +11,7 @@ class Dmv
   end
 
   def facilities_offering_service(service)
-    @facilities.find do |facility|
+    @facilities.find_all do |facility|
       facility.services.include?(service)
     end
   end
--- a/lib/facility.rb
+++ b/lib/facility.rb
@@ -1,10 +1,10 @@
 class Facility
   attr_reader :name, :address, :phone, :services, :registered_vehicles, :collected_fees
 
-  def initialize(hash)
-    @name = hash[:name]
-    @address = hash[:address]
-    @phone = hash[:phone]
+  def initialize(facility_details)
+    @name = facility_details[:name]
+    @address = facility_details[:address]
+    @phone = facility_details[:phone]
     @services = []
     @registered_vehicles = []
     @collected_fees = 0
--- a/lib/facility.rb
+++ b/lib/facility.rb
@@ -1,14 +1,14 @@
 class Facility
   attr_reader :name, :address, :phone, :services
 
-  def initialize(name, address, phone)
-    @name = name
-    @address = address
-    @phone = phone
+  def initialize(hash)
+    @name = hash[:name]
+    @address = hash[:address]
+    @phone = hash[:phone]
     @services = []
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
 end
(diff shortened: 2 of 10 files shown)


lib/dmv.rb
```diff
@@ -1,5 +1,7 @@
 class Dmv
-
+  attr_reader :name, :address, :phone
+  attr_accessor :facilities
```
KatBrandt, line 3: Don't need this to be an attr_accessor, just an attr_reader. :)

lib/facility.rb
```diff
@@ -1,14 +1,67 @@
...
+      else 
+        vehicle.plate_type = :regular
+      end 
+      vehicle.plate_type
+  end 
+
+  def administer_written_test(registrant) 
+    services.include?("Written Test") &&
+    registrant.age >= 16 && 
+    registrant.permit? &&
```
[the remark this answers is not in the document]
KatBrandt: Same comment for the following two methods

spec/facility_spec.rb
```diff
@@ -2,25 +2,275 @@
...
+    
+    it 'facility provides the service needed' do 
+      expect(@facility_1.services).to eq([])
+
+      @facility_1.add_service('Vehicle Registration')
+
+      expect(@facility_1.services).to eq(['Vehicle Registration'])
+    end 
+
+    it 'does not provide service if the facility does not offer it' do
```
KatBrandt, line 45: This it block isn't really necessary, it's not testing anything

spec/facility_spec.rb
```diff
@@ -2,25 +2,275 @@
...
+  
+    it 'can give a date a vehicle is registered' do
+      cruz = Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice} )
+
+            
+      @facility_1.add_service('Vehicle Registration')
+      @facility_1.register_vehicle(cruz)
+
+      expect(@facility_1.registered_vehicles.first.registration_date).to eq(Date.today)
+    end
```
KatBrandt, line 70: Great organization to this describe block for the register_vehicle method. We can also include all the side effects in this block, including that a  plate type is changed and fees are collected. That would be the integration test. Then when you test those methods by themselves in their own it blocks, those would be the unit tests.

spec/facility_spec.rb
```diff
@@ -2,25 +2,275 @@
...
+
+  describe '#administer_written_test(person)' do 
+    it 'does not administer a written if service is not provided' do 
+      expect(@facility_1.services).to eq([])
+
+      registrant_1 = Registrant.new('Bruce', 18, true )
+
+      @facility_1.administer_written_test(registrant_1)
+
+      expect(@facility_1.administer_written_test(registrant_1)).to eq(false)
```
KatBrandt, line 145: good testing that it does NOT administer this test. I'm wondering if this test passed though? It looked like your code didn't actually ever return a `false` value.

spec/facility_spec.rb
```diff
@@ -2,25 +2,275 @@
...
+      @facility_1.administer_written_test(registrant_3)
+
+      expect(@facility_1.administer_written_test(registrant_3)).to eq(false)
+      
+      @facility_1.administer_road_test(registrant_3)
+
+      expect(@facility_1.administer_road_test(registrant_3)).to eq(false)
+      expect(@facility_1.renew_drivers_license(registrant_3)).to eq(false)
+    end 
+  end
```
KatBrandt, line 275: Overall good job with testing all the different potential outcomes of these methods.

spec/facility_spec.rb
JRIV-10: Yes my test did pass and i was able to return a false value!

PR #413, opened 2023-12-12
Jack Sweeney: The DMV Project

JackCSweeney:
Here is my pull request for the DMV project


lib/analyzer.rb
```diff
@@ -0,0 +1,52 @@
...
+        @vehicle_models
+    end
+
+    def model_counter(vehicles)
+        vehicle_models(vehicles)
+        @vehicle_models.tally
+    end
+
+    def most_popular_model(vehicles)
+        most_popular_model = model_counter(vehicles).max_by do |model, count|
```
KatBrandt, line 21: Great use of max_by! No need for the variable though since you're not doing anything with that data other than returning it.

lib/analyzer.rb
```diff
@@ -0,0 +1,52 @@
+class Analyzer
+
+    def initialize
```
KatBrandt, line 3: Since all of your methods are taking the same argument, it might be better design to instead initialize an Analyzer object with this argument of `vehicles`. Then you can populate the instance variables (ivars) in the initialize so line 4 might look like `@vehicle_models = vehicle_models(vehicles)` that way your instance variable's data is populated from the get go so all other methods can simply call on that ivar.

lib/analyzer.rb
```diff
@@ -0,0 +1,52 @@
...
+        @vehicle_models = []
+        @most_popular_vehicle = [] 
+    end
+
+    def vehicle_models(vehicles)
```
KatBrandt, line 8: Remember that method names should be verbs, so something like `add_vehicle_models` would be more appropriate. Variable names should be nouns.

lib/analyzer.rb
```diff
@@ -0,0 +1,52 @@
...
+            count
+        end
+        most_popular_model
+    end
+
+    def most_popular_vehicle(vehicles)
+        vehicle_object = vehicles.find do |vehicle|
+            vehicle.model == most_popular_model(vehicles).first
+        end
+        @most_popular_vehicle << vehicle_object.make
```
KatBrandt, line 31: typcially if we have instance variables, we would want them declared in the initialize, otherwise we can ask ourselves, "should this be in ivar or a local variable?" For this method, I think a local variable would be appropriate.

lib/analyzer.rb
```diff
@@ -0,0 +1,52 @@
...
+        most_popular_model = model_counter(vehicles).max_by do |model, count|
+            count
+        end
+        most_popular_model
+    end
+
+    def most_popular_vehicle(vehicles)
+        vehicle_object = vehicles.find do |vehicle|
+            vehicle.model == most_popular_model(vehicles).first
+        end
```
KatBrandt, line 30: I'm not quite sure what this method is doing or returning. Part of the confusion could be in how things are being named and what they're returning. For the `most_popular_model` method, I would expect that method to return a single string listing the most popular model, so then what would `most_popular_vehicle` represent? How is this different than the `most_popular_model` method? I will look at your tests and see if there is documentation there to provide clues! 
I do like your experimentation though!

lib/facility.rb
```diff
@@ -1,14 +1,60 @@
...
+    @registered_vehicles = []
+    @collected_fees = 0
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
```
KatBrandt, line 24: Great job with this method. Love all the helper methods called in it. Makes it very readable and clear what the side effects are.

lib/facility.rb
```diff
@@ -1,14 +1,60 @@
...
+      @collected_fees += 25
+    elsif vehicle.plate_type == :ev
+      @collected_fees += 200
+    else
+      @collected_fees += 100
+    end
+  end
+
+  def administer_written_test(registrant)
+    if @services.include?("Written Test") && registrant.permit? && registrant.age > 15
```
KatBrandt, line 44: Nice job using the logical && operators to keep this if statement clean

lib/facility_maker.rb
```diff
@@ -0,0 +1,79 @@
...
+        @dmv_data = dmv_data
+        @new_facilities = []
+        @new_facility_data = {}
+    end
+
+    def facility_maker_co
+        if @dmv_data.first[:state] == "CO"
+            data_organizer_co
+        end
+    end
```
KatBrandt, line 15: Really great use of helper methods here. I wonder if you could combine this method and the `facility_maker_ny` methods though and use the if/else statement to determine which state's data is being fed in and then call the appropriate method.

lib/facility_maker.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    def facility_maker_co
+        if @dmv_data.first[:state] == "CO"
+            data_organizer_co
+        end
+    end    
+
+    def data_organizer_co
+        @dmv_data.each do |location_data|
+            @new_facility_data[:name] = location_data[:dmv_office]
```
KatBrandt, line 19: I would just keep this as a local variable. We only need instance variables if it's going to be stored data and other methods will use that data. So I like `@new_facilities` as an ivar since that's data we'll want to hold onto, but we don't need `@new_facility_data` as an ivar, that's just local to this method.

lib/facility_maker.rb
KatBrandt: and the MO one!

lib/facility_maker.rb
```diff
@@ -0,0 +1,79 @@
...
+                    @new_facility_data[:phone] = "No Public Phone Number"
+                end
+            @new_facilities << Facility.new(@new_facility_data)
+        end
+    end
+
+    def facility_maker
+        facility_maker_co
+        facility_maker_mo
+        facility_maker_ny
```
KatBrandt, line 76: oooo, I see, you did combine them all here. I think we can par this down a little and just have the one method `facility_maker` and then in this method have an if/else statement that then determines if we are making a CO, MO, or NY facility.

lib/registrant.rb
```diff
@@ -0,0 +1,22 @@
+class Registrant
+
+    attr_reader :name, :age, :permit, :license_data
```
KatBrandt, line 3: Careful here, you have exposed a method called `.permit`, but we never need that method and don't actually want that method.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,45 @@
+class VehicleFactory
+
+    def initialize
+        @created_vehicles = []
+        @vehicle_data = {}
+    end
+
+    def create_vehicles(registrations)
+       registrations.each do |vehicle_hash|
+        @vehicle_data[:engine] = :ev
```
KatBrandt, line 10: Same thing here, don't need the vehicle data as an ivar. Otherwise, great job with this method!

lib/facility.rb
```diff
@@ -1,14 +1,60 @@
...
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    if @services.include?('Vehicle Registration')
+      vehicle.add_registration_date(Date.today)
+      vehicle.plate_maker
```
KatBrandt, line 27: I like that you made this method in the Vehicle class, let that class be responsible for changing the attribute.

spec/analyzer_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+    end
+
+    describe '#most_popular_vehicle(vehicles)' do
+        it 'can analyze an array of one vehicle and return the make and model' do
+            cruz = Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice})
+            analyzer = Analyzer.new
+
+            vehicles = [cruz]
+
+            expect(analyzer.most_popular_vehicle(vehicles)).to eq(['Chevrolet', 'Cruz'])
```
KatBrandt, line 19: We'd want a more robust test to make sure this method actually figures out the most popular vehicle. Having just one vehicle in the array doesn't give me confidence that this method can actually find the right value when there are 50 items in the array.

spec/analyzer_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+        it 'can analyze multiple vehicles and return the make and model of the most popular' do
+            cruz_1 = Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice})
+            cruz_2 = Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice})
+            camaro = Vehicle.new({vin: '1a2b3c4d5e6f', year: 1969, make: 'Chevrolet', model: 'Camaro', engine: :ice})
+            bolt = Vehicle.new({vin: '987654321abcdefgh', year: 2019, make: 'Chevrolet', model: 'Bolt', engine: :ev})
+            analyzer = Analyzer.new
+
+            vehicles = [cruz_1, cruz_2, camaro, bolt]
+
+            expect(analyzer.most_popular_vehicle(vehicles)).to eq(['Chevrolet', 'Cruz'])
```
KatBrandt, line 31: I see, you did it a bit more robust down here. I wouldn't even have that first test then, just this one.

spec/analyzer_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+        it 'can count how many of each model of vehicle there are and return a hash with the model as a key and count as value' do
+            cruz_1 = Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice})
+            cruz_2 = Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice})
+            camaro = Vehicle.new({vin: '1a2b3c4d5e6f', year: 1969, make: 'Chevrolet', model: 'Camaro', engine: :ice})
+            bolt = Vehicle.new({vin: '987654321abcdefgh', year: 2019, make: 'Chevrolet', model: 'Bolt', engine: :ev})
+            analyzer = Analyzer.new
+
+            vehicles = [cruz_1, cruz_2, camaro, bolt]
+
+            expect(analyzer.model_counter(vehicles)).to eq({"Cruz"=> 2, "Camaro"=> 1, "Bolt"=>1})
```
KatBrandt, line 59: Nice!

lib/analyzer.rb
```diff
@@ -0,0 +1,52 @@
...
+    def vehicle_models(vehicles)        
+        vehicles.each do |vehicle|
+            @vehicle_models << vehicle.model
+        end
+        @vehicle_models
+    end
+
+    def model_counter(vehicles)
+        vehicle_models(vehicles)
+        @vehicle_models.tally
```
KatBrandt, line 17: `.tally`!!! Thanks for teaching me a new method! :)

spec/analyzer_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+        it 'returns an array with the model and count of the most popular vehicle' do
+            cruz_1 = Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice})
+            cruz_2 = Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice})
+            camaro = Vehicle.new({vin: '1a2b3c4d5e6f', year: 1969, make: 'Chevrolet', model: 'Camaro', engine: :ice})
+            bolt = Vehicle.new({vin: '987654321abcdefgh', year: 2019, make: 'Chevrolet', model: 'Bolt', engine: :ev})
+            analyzer = Analyzer.new
+
+            vehicles = [cruz_1, cruz_2, camaro, bolt]
+
+            expect(analyzer.most_popular_model(vehicles)).to eq(["Cruz", 2])
```
KatBrandt, line 73: When you have data like this, I think storing it in a hash makes more sense than an array with 2 elements.

spec/facility_maker_spec.rb
```diff
@@ -0,0 +1,180 @@
...
+            expect(facility_maker.new_facilities.last).not_to be_a(Facility)
+        end
+
+        it 'stores data for colorado facilities in new facility objects' do
+            co_dmv_office_locations = DmvDataService.new.co_dmv_office_locations
+            facility_maker = FacilityMaker.new(co_dmv_office_locations)
+
+            facility_maker.facility_maker_co
+
+            expect(facility_maker.new_facilities.first.name).not_to eq(nil)
```
KatBrandt, line 43: We can also test that indeed the elements of this array are Facility objects.

spec/facility_spec.rb
```diff
@@ -2,25 +2,293 @@
...
-      expect(@facility.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
+      expect(@facility_1.services).to eq([])
+      @facility_1.add_service('New Drivers License')
+      @facility_1.add_service('Renew Drivers License')
+      @facility_1.add_service('Vehicle Registration')
+      expect(@facility_1.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
+    end
+  end
+
+  describe '#plate_maker' do
```
KatBrandt, line 31: Since this is a method that lives in the Vehicle class, this method should be tested in the Vehicle class. This is a good unit test though! Just needs to be in a different file.

spec/facility_spec.rb
```diff
@@ -2,25 +2,293 @@
...
+
+    it 'gives a vehicle a plate type of :ev when registered if vehicle has :ev engine:' do
+      bolt = Vehicle.new({vin: '987654321abcdefgh', year: 2019, make: 'Chevrolet', model: 'Bolt', engine: :ev})
+
+      @facility_1.add_service('Vehicle Registration')
+      @facility_1.register_vehicle(bolt)
+
+      expect(bolt.plate_type).to eq(:ev)
+    end
+  end
```
KatBrandt, line 115: Really great describe block for this one register_vehicle method with all the it blocks that test each of the side effects. This is great integration testing.

spec/facility_spec.rb
```diff
@@ -2,25 +2,293 @@
...
+      cruz = Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice})
+
+      @facility_1.add_service('Vehicle Registration')
+      @facility_1.register_vehicle(bolt)
+      @facility_1.register_vehicle(camaro)
+      @facility_1.register_vehicle(cruz)
+
+      expect(@facility_1.collected_fees).to eq(325)
+    end
+  end
```
KatBrandt, line 161: Great tests for the fees!

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+            expect(factory.create_vehicles(wa_ev_registrations)).to be_a(Array)
+        end
+    end
+
+    describe '#create_vehicles_ny' do
+        it 'can create vehicles from NY external data' do
+            factory = VehicleFactory.new
+            ny_vehicle_registrations = DmvDataService.new.ny_vehicle_registrations 
+            
+            expect(factory.create_vehicles_ny(ny_vehicle_registrations).first).to be_a(Vehicle)
```
KatBrandt, line 28: Yup, good test.

spec/vehicle_spec.rb
```diff
@@ -33,4 +33,42 @@
...
+    end
+  end
+
+  describe '#add_registration_date' do
+    it 'can add a registration date to a vehicle' do
+      expect(@cruz.add_registration_date(Date.today)).to eq(@cruz.registration_date)
+    end
+  end
+
+  describe '#plate_maker' do
```
KatBrandt, line 49: You do have the test for this here! Ok, no need to also have a test for this method in the Facility spec file.

KatBrandt (commented):
Overall, really great work with this! 

In regards to your question about commits, I don't think you have too many! If you think you do, you can definitely be more aware and intentional about when you commit, but for this project, 87 commits isn't a crazy amount.

PR #414, opened 2023-12-12
Mel Langhoff

mel-langhoff:


lib/facility.rb
```diff
@@ -1,14 +1,62 @@
...
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    if @services.include?("Vehicle Registration")
+      @registered_vehicle << vehicle
+      vehicle.car_registration_date
+      collect_fees(vehicle)
+      @registration_date = vehicle.car_registration_date
```
KatBrandt, line 22: Don't need to create an instance variable here for @registration_date, calling just vehicle.car_registration_date will accomplish what you need. In terms of naming, remember that method names should be verbs, so maybe calling this method something like vehicle.set_registration_date would have been more indicative of what it does. 
Overall, great job using all these helper methods with this register_vehicle method.!

lib/facility.rb
```diff
@@ -1,14 +1,62 @@
...
+    end
+  end
+
+  def administer_written_test(registrant)
+    if registrant.age >= 16 && @services.include?("Written Test") && registrant.permit?
```
KatBrandt, line 27: Nice job with this conditional logic.

lib/facility.rb
```diff
@@ -1,14 +1,62 @@
...
+      registrant.license_data[:written] = true
+      true
+    else
+      false
+    end
+  end
+
+
+  def administer_road_test(registrant)
+    if registrant.age >= 16 && @services.include?("Road Test") && registrant.permit? && registrant.license_data[:written] == true
```
KatBrandt, line 37: don't need the `== true` at the end since the expression `registrant.license_data[:written]` will already return a true or false.

lib/office_locations.rb
```diff
@@ -0,0 +1,128 @@
...
+
+    def create_facility(source)
+        source.each do |facility|
+            if facility.keys.include?(:dmv_office)
+                co_create_facility(source)
+            elsif facility.keys.include?(:office_name)
+                ny_create_facility(source)
+            else facility.keys.include?(:name)
+                mo_create_facility(source)
+            end
```
KatBrandt, line 11: Nice job separating all this work into helper methods. Makes it very readable.

lib/office_locations.rb
```diff
@@ -0,0 +1,128 @@
...
+                    name: name,
+                    address: address,
+                    phone: phone,
+                    services: services,
+                    registered_vehicle: [],
+                    collected_fees: 0
+                })
+            co_dmv_facilities_array << facility
+        end
+        co_dmv_facilities_array
```
KatBrandt, line 75: Awesome job with this Mel!

lib/registrant.rb
```diff
@@ -0,0 +1,28 @@
+class Registrant
+    attr_reader :age,
+                :license_data,
+                :name,
+                :permit
```
KatBrandt, line 5: Careful here, we don't need a method called `.permit` so we don't want to write a method we don't need.

lib/registrant.rb
```diff
@@ -0,0 +1,28 @@
...
+            written: false,
+            license: false,
+            renewed: false
+            }
+        @permit = permit
+               
+    end
+
+    def permit?
+        @permit == @permit
```
KatBrandt, line 20: This would actually always give a `true`. You're comparing the same thing on both sides of the equation so there's no way this would ever give a `false`.  I'm curious what the tests look like for this method, will look in a second. We'd instead just want this method to be `@permit` inside so it's just reading the value of the instance variable.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+        expect(registrant_1).to be_an_instance_of(Registrant)
+        expect(registrant_2).to be_an_instance_of(Registrant)
+    end
+
+    it '#permit?' do
+        registrant_1 = Registrant.new('Bruce', 18, true)
+        registrant_2 = Registrant.new('Penny', 15)
+
+        expect(registrant_1.permit).to eq true
+        expect(registrant_2.permit).to eq false
```
KatBrandt, line 19: ahhh, see, you're testing that other method `.permit`, not the method `.permit?` here, that's why you didn't catch the error.

lib/registrant.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    def earn_permit
+        @permit = true if @permit == false
```
KatBrandt, line 24: This works, can also do `@permit = true if !@permit` :)

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,106 @@
...
+                    make: make,
+                    model: model,
+                    engine: :ev,
+                    registration_date: nil,
+                    plate_type: nil
+                })
+            @wa_ev_vehicles_array << wa_ev_vehicle
+        end
+        @wa_ev_vehicles_array
+    end
```
KatBrandt, line 58: Great job with all these Vehicle making methods

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    def most_popular_make
+        hash_grouped_by_make = @wa_ev_vehicles_array.group_by do |vehicle|
```
KatBrandt, line 61: Great use of group_by, nice!

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,106 @@
...
+            vehicle.make
+        end
+        most_popular_make = hash_grouped_by_make.max_by do |make, vehicles_array| # .max_by finds largest array
+            vehicles_array.size
+        end
+        most_popular_make.first
```
KatBrandt, line 67: Awesome, love the use of enumerables here. Care with your variable names though, we dont' want to make the variable name the same as the name of the method. Can cause confusion.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,106 @@
...
+                year: year,
+                make: make,
+                model: model,
+                engine: nil,
+                registration_date: registration_date,
+                plate_type: nil
+            })
+            @ny_vehicles_array << ny_vehicle
+        end
+        @ny_vehicles_array
```
KatBrandt, line 35: I would have an initialize method that actually declares these instance variables, so instead of having line 16 in this method, have it in the initialize. This is helpful because when we always declare our instance variables in the initialize method, we always know where to look for those instance variables.

spec/facility_spec.rb
```diff
@@ -16,11 +22,134 @@
...
+    it "has a plate type" do
+      facility_1 = Facility.new({name: 'DMV Tremont Branch', address: '2855 Tremont Place Suite 118 Denver CO 80205', phone: '(720) 865-4600'})
+      facility_2 = Facility.new({name: 'DMV Northeast Branch', address: '4685 Peoria Street Suite 101 Denver CO 80239', phone: '(720) 865-4600'})
+      cruz = Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice} )
+      bolt = Vehicle.new({vin: '987654321abcdefgh', year: 2019, make: 'Chevrolet', model: 'Bolt', engine: :ev} )
+      camaro = Vehicle.new({vin: '1a2b3c4d5e6f', year: 1969, make: 'Chevrolet', model: 'Camaro', engine: :ice} )
+
+      cruz.determine_plate_type
+      camaro.determine_plate_type
+      bolt.determine_plate_type
```
KatBrandt, line 71: This method should actually just be tested in the Vehicle spec since it's a method that lives in the Vehicle class. You can run the Facility method of register_vehicle and ensure that the plate_type was set though, since that's one of the side effects of the register_vehicle method.

spec/facility_spec.rb
```diff
@@ -16,11 +22,134 @@
...
+
+      facility_1.add_service('Vehicle Registration')
+      facility_1.register_vehicle(cruz)
+
+      expect(facility_1.collected_fees).to eq(100)
+
+      facility_1.register_vehicle(camaro)
+      facility_1.register_vehicle(bolt)
+
+      expect(facility_1.collected_fees).to eq(325)
```
KatBrandt, line 93: Great job showing the accumulation of fees and testing those.

spec/facility_spec.rb
```diff
@@ -16,11 +22,134 @@
...
+      facility_2 = Facility.new({name: 'DMV Northeast Branch', address: '4685 Peoria Street Suite 101 Denver CO 80239', phone: '(720) 865-4600'})
+
+      facility_1.add_service('Written Test')
+
+      expect(facility_1.administer_written_test(registrant_1)).to be true
+      expect(facility_1.administer_written_test(registrant_2)).to be true
+      expect(facility_1.administer_written_test(registrant_3)).to be false
+    end
+
+  it "#administer_road_test" do
```
KatBrandt, line 110: For these methods, we'd also want to test that the values stay "false" if the service of "Written Test" has not been added to the facility. Basically we want to test for each of those three conditions listed in the if statement for this method.

spec/office_locations_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+
+    it "#create_facility" do
+        office_location = OfficeLocations.new
+        co_dmv_office_locations = DmvDataService.new.co_dmv_office_locations
+        new_york_facilities = DmvDataService.new.ny_dmv_office_locations
+        missouri_facilities = DmvDataService.new.mo_dmv_office_locations
+
+        expect(office_location.create_facility(co_dmv_office_locations)).to be_a Array
+        expect(office_location.create_facility(new_york_facilities)).to be_a Array
+        expect(office_location.create_facility(missouri_facilities)).to be_a Array
```
KatBrandt, line 20: Yes, can also test that each element of these arrays are Facility objects.

KatBrandt (commented):
Overall, great job Mel! You've come a very long way!

PR #415, opened 2023-12-12
Igor Magalhaes 2311 BE

IgorrMagalhaess:


KatBrandt (commented):
Overall, really great job!!! It's clear that you understand all the fundamental concepts we've learned so far. I am glad this was still challenging for you though! That means learning is happening. 
Let's just work on some of the testing and really isolating logic to the appropriate spec files. This is a PASSING project. Great job, please let me know if you have questions.

PR #416, opened 2023-12-12
Jess Kohl

kohljd:


spec/facility_locations_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+    it 'returns facility objects' do
+        dmv = FacilityLocations.new
+        co_dmv_office_locations = DmvDataService.new.co_dmv_office_locations
+        ny_dmv_office_locations = DmvDataService.new.ny_dmv_office_locations
+        mo_dmv_office_locations = DmvDataService.new.mo_dmv_office_locations
+        
+        expect(dmv.co_dmv_offices(co_dmv_office_locations)).to all(be_an_instance_of(Facility))
+        expect(dmv.ny_dmv_offices(ny_dmv_office_locations)).to all(be_an_instance_of(Facility))
+        expect(dmv.mo_dmv_offices(mo_dmv_office_locations)).to all(be_an_instance_of(Facility))
+    end
```
jamisonordway, line 13: This is great. You could also consider testing the first or last Facility in each of these collections to make sure the attributes are getting saved:
`expect(dmv.co_dmv_offices(co_dmv_office_locations).last.name).to be_a(String)`
and so on for the rest of the attributes.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,15 @@
...
+    def initialize; end
+
+    def create_vehicles(registrations)
+        registrations.map do |car_details|
+            Vehicle.new({
+            vin: car_details[:vin_1_10],
+            year: car_details[:model_year],
+            make: car_details[:make],
+            model: car_details[:model],
+            engine: :ev
```
jamisonordway, line 11: Conventionally these lines would be indented one more tab.

lib/facility.rb
```diff
@@ -1,14 +1,64 @@
...
+        @collected_fees += 25
+        vehicle.plate_type = :antique
+      elsif vehicle.electric_vehicle?
+        @collected_fees += 200
+        vehicle.plate_type = :ev
+      else
+        @collected_fees += 100
+        vehicle.plate_type = :regular
+      end
+    end
```
jamisonordway, line 38: You could use an explicit return for a guard clause here, which would eliminate the need for a nested conditional:
```ruby
return false unless @services.include?("Vehicle Registration")
# etc
```

jamisonordway (commented):
Nice work! 🎉

PR #417, opened 2023-12-12
Faisal

mfaisalnazari:
iteration 3 complete

### DIFF
--- a/lib/facility.rb
+++ b/lib/facility.rb
@@ -1,14 +1,14 @@
 class Facility
   attr_reader :name, :address, :phone, :services
 
-  def initialize(name, address, phone)
-    @name = name
-    @address = address
-    @phone = phone
+  def initialize(attributes = {})
+    @name = attributes[:name]
+    @address = attributes[:address]
+    @phone = attributes[:phone]
     @services = []
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
 end
--- a/lib/facility.rb
+++ b/lib/facility.rb
@@ -1,14 +1,73 @@
+
 class Facility
-  attr_reader :name, :address, :phone, :services
+  
+  attr_reader :name, :address, :phone, :services, :registered_vehicles,
+  :collected_fees
 
   def initialize(attributes = {})
     @name = attributes[:name]
     @address = attributes[:address]
     @phone = attributes[:phone]
     @services = []
+    @registered_vehicles = []
+    @collected_fees = 0
   end
 
   def add_service(service)
     @services << service
+  end
+  def register_vehicle(vehicle)
+    if @services.include?('Vehicle Registration')
+      vehicle.registration_date = Time.now
+      if vehicle.antique?
+        @collected_fees += 25
+        vehicle.plate_type = :antique
+      elsif vehicle.electric_vehicle?
+          @collected_fees += 200
+          vehicle.plate_type = :ev
+      else 
+        @collected_fees += 100
+        vehicle.plate_type = :regular
+        end
+      @registered_vehicles << vehicle
+    end
+  end
+  def administer_written_test(registrant)
+    if @services.include?('Written Test')  
+      if registrant.permit? && registrant.age >= 16
+        registrant.license_data[:written] = true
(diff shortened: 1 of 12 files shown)


lib/facility.rb
```diff
@@ -1,14 +1,74 @@
...
+        vehicle.plate_type = :antique
+      elsif vehicle.electric_vehicle?
+          @collected_fees += 200
+          vehicle.plate_type = :ev
+      else 
+        @collected_fees += 100
+        vehicle.plate_type = :regular
+        end
+      @registered_vehicles << vehicle
+    end
```
jamisonordway, line 33: For nested conditionals like this one, you could use an explicit return for a guard clause that would eliminate the need for nesting:
```ruby
return false unless @services.include?('Vehicle Registration')
vehicle.registration_date = Time.now
# etc
```

lib/facility_factory.rb
```diff
@@ -0,0 +1,22 @@
+class FacilityFactory
+    
+    def create_facilities(facility_details)
+        
+        facilities_created = []
+        facility_details.each do |facility|
+          facility1 = {}
```
jamisonordway, line 7: I might call this something like `facility_data`, `facility1` feels arbitrary and less dynamic than it actually is.

lib/vehicle.rb
```diff
@@ -6,13 +6,19 @@ class Vehicle
...
+              
 
   def initialize(vehicle_details)
     @vin = vehicle_details[:vin]
     @year = vehicle_details[:year]
     @make = vehicle_details[:make]
     @model = vehicle_details[:model]
     @engine = vehicle_details[:engine]
+    @registeration_date 
+    @plate_type
```
jamisonordway, line 21: It seems like this is working as is, but I would expect these attributes to be defined with an assignment operator, even if the values start as nil or an empty string.
```ruby
@registration_date = nil
@plate_type = nil
```

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,13 @@
+class VehicleFactory
+    
+    def create_vehicles(vehicle_details)
+        vehicles_created = []
+        vehicle_details.each do |vehicle|
+            vehicle[:engine] = "ev"
+            vehicles_created << Vehicle.new(vehicle)
+        end
+        vehicles_created
```
jamisonordway, line 9: You could refactor this to use `.map` and shorten your code:
```ruby
def create_vehicles(vehicle_details)
  vehicle_details.map do |vehicle|
    vehicle[:engine] = "ev"
    Vehicle.new(vehicle)
  end
end
```

spec/facility_spec.rb
```diff
@@ -22,5 +27,64 @@
...
+      expect(@cruz.registration_date).to be_within(1.0).of(Time.now)
+      expect(@cruz.plate_type).to eq(:regular)
+      expect(@facility.registered_vehicles).to eq([@cruz])
+      expect(@facility.collected_fees).to eq(100)
+
+    end
+    
+  end
+  describe 'written test' do
+    it 'can take test' do
```
jamisonordway, line 44: Generally it's a better practice to use `true` and `false` instead of `!false` and `!true`. These tests would work the same way with this change.
If you really need to test for any false-y value (in Ruby, the only false-y values are `nil` and `false`), you could do something like `expect(<whatever>).to be_falsey`, or for the opposite, `expect(<whatever>).to be_truthy`

spec/registrant_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+    describe 'attributes' do
+        it 'has attributes' do
+            expect(@registrant_1.name).to eq("Bruce")
+            expect(@registrant_1.age).to eq(18)
+            expect(@registrant_1.permit).to eq(true)
+            expect(@registrant_1.license_data).to eq({:license=>false, :renewed=>false, :written=>false})
+        end
+    end
+    describe 'methods work' do
```
jamisonordway, line 21: I would call this describe `#earn_permit` or maybe `#earn_permit changes return value of #permit?`

PR #420, opened 2023-12-12
Odell McFarland IV

odellmac4:


spec/facility_spec.rb
```diff
@@ -21,6 +29,208 @@
...
+
+      @facility_2.fee_assignment(@camaro)
+      @facility_2.fee_assignment(@bolt)
+      expect(@facility_2.collected_fees).to eq(225)
+
+      @facility_2.plate_assignment(@camaro)
+      @facility_2.plate_assignment(@bolt)
+      expect(@camaro.plate_type).to eq (:antique)
+      expect(@bolt.plate_type).to eq (:ev)
+      #Need to figure out how a vehicle cannot be registered twice and/or at another facility
```
jamisonordway, line 108: Limiting a vehicle to one facility would be tricky and probably involve a [class method](https://backend.turing.edu/module1/lessons/class_methods). But for the other part you could add something like this to your `register_vehicle` method:
```ruby
unless @vehicles.includes?(vehicle_instance)
```

PR #421, opened 2023-12-12
Selena H

Selena730:
# The DMV
This is a program to manage some dmv services and process some data.

### DIFF
--- a/lib/facility.rb
+++ b/lib/facility.rb
@@ -1,14 +1,14 @@
 class Facility
   attr_reader :name, :address, :phone, :services
 
-  def initialize(name, address, phone)
-    @name = name
-    @address = address
-    @phone = phone
+  def initialize(info)
+    @name = info[:name]
+    @address = info[:address]
+    @phone = info[:phone]
     @services = []
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
 end
--- a/lib/facility.rb
+++ b/lib/facility.rb
@@ -1,5 +1,6 @@
 require './lib/vehicle'
 require './lib/registrant'
+require './lib/dmv_data_service'
 
 class Facility
   attr_reader :name, :address, :phone, :services, :registered_vehicles, :collected_fees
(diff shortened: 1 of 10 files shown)


lib/facility.rb
```diff
@@ -1,14 +1,109 @@
...
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    if @services.include?('Vehicle Registration')
+      vehicle.registration_date = Date.today
+      vehicle.plate_type = set_plate_type(vehicle)
```
jamisonordway, line 26: Since you're assigning the `vehicle.plate_type` on lines 40, 43, or 46, this line should probably just call the method:
```
...
set_plate_type(vehicle)
...
```

lib/facility.rb
```diff
@@ -1,14 +1,109 @@
...
+    else
+      vehicle.plate_type = :regular
+    end
+  end
+
+  def calculate_registration_fee(vehicle)
+    return 25 if vehicle.antique?
+    return 200 if vehicle.electric_vehicle?
+    100
+  end
```
jamisonordway, line 54: Great job keeping things readable and DRY 🎉

spec/facility_spec.rb
```diff
@@ -23,4 +34,165 @@
...
+  end
+
+  describe '#register_vehicle(camaro)' do
+
+    it 'registers (camaro)' do
+      @facility_1.add_service('Vehicle Registration')
+
+      expect(@facility_1.register_vehicle(@camaro)).to eq(@facility_1.registered_vehicles)
+    end
+  end
```
jamisonordway, line 63: These are great tests! You could consider registering 3 vehicles in one example block as well, that way you could test that `@facility_1.registered_vehicles` contains all of them.

spec/facility_spec.rb
```diff
@@ -23,4 +34,165 @@
...
+    it 'returns the plate type' do
+      @facility_1.registered_vehicles
+
+      expect(@facility_1.set_plate_type(@cruz)).to eq(:regular)
+      expect(@facility_1.set_plate_type(@bolt)).to eq(:ev)
+      expect(@facility_1.set_plate_type(@camaro)).to eq(:antique)
+    end
+  end
+
+  describe '#administer_written_test' do
```
jamisonordway, line 75: No concerns with these tests-- it looks like all outcomes are covered-- this is just some food for thought. 
In RSpec you can use `describe` blocks to organize and nest different scenarios. This can sometimes make large/complex tests easier to follow. For example: 
```
describe '#administer_written_test' do
  describe 'successful' do
    before(:each) do
      # setup for successful scenario
    end
    # example (it) blocks
  end

  describe 'unsuccessful' do
    before(:each) do
      # setup for successful scenario
    end
    describe 'registrant is too young' do
      # example block(s)
    end
    describe 'facility does not offer service' do
      # example block(s)
    end
  end
end
```
and so forth.

jamisonordway (commented):
Great work! Let me know if you have any follow up questions about these comments.

PR #422, opened 2023-12-12
Luis Aparicio

LuisAparicio12:


lib/facility.rb
```diff
@@ -1,14 +1,93 @@
...
+    @registered_vehicles = []
+    @collected_fees = 0
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle_helper(car)
```
jamisonordway, line 17: I'm a big fan of how you have the logic in this class organized with the use of helper methods 🎉  
This is a bit of a nitpick, but generally it's best practice to not use qualifiers like 'helper' in method names. If you wanted to polish this class even more, you could consider renaming your methods. Here are some examples/ideas:
```
#register_vehicle_helper -> #process_plate_types
#written_test_helper -> #attempt_written_test or #validate_written
#renew_drivers_license_helper -> #attempt_license_renewal or #validate_renewed_license
```

lib/facility_factory.rb
```diff
@@ -0,0 +1,47 @@
+class FacilityFactory
+
+    def initialize
+    end
```
jamisonordway, line 4: For classes that don't accept arguments upon instantiation, with an empty initialize, it's best to remove this method entirely. (`.new` will still work and assume no args will be passed)

lib/facility_factory.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    def construct_co_facility_attr_helper(office_data)
```
jamisonordway, line 6: similar to my comment above, we could just remove `_helper` from these method names in my opinion

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,22 @@
...
+        vehicle_attributes[:vin] = vehicle_data[:vin_1_10]
+        vehicle_attributes[:year] = vehicle_data[:model_year]
+
+        vehicle_attributes
+    end
+
+    def create_vehicles(data)
+        data.map do |vehicle_data|
+            Vehicle.new(construct_vehicle_attr_helper(vehicle_data))
+        end
```
jamisonordway, line 20: Keep an eye on your indentation:
```
data.map do |vehicle_data|
  Vehicle.new(...)
end
```

spec/facility_spec.rb
```diff
@@ -23,4 +23,276 @@
...
+      facility_2 = Facility.new({name: 'DMV Northeast Branch', address: '4685 Peoria Street Suite 101 Denver CO 80239', phone: '(720) 865-4600'})
+      cruz = Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice} )
+      bolt = Vehicle.new({vin: '987654321abcdefgh', year: 2019, make: 'Chevrolet', model: 'Bolt', engine: :ev} )
+      camaro = Vehicle.new({vin: '1a2b3c4d5e6f', year: 1969, make: 'Chevrolet', model: 'Camaro', engine: :ice} )
+
+      expect(facility_1.add_service('Vehicle Registration')).to eq(facility_1.services)
+      expect(facility_1.registered_vehicles).to eq([])
+      expect(facility_1.collected_fees).to eq(0)
+    
+      expect(facility_1.register_vehicle(cruz)).to eq([cruz])
```
jamisonordway, line 39: I would also like to see a test for what happens if we try to register a vehicle in the scenario where this service has not been added to the facility.

spec/facility_spec.rb
```diff
@@ -23,4 +23,276 @@
...
+      registrant_2 = Registrant.new('Penny', 16 )
+      registrant_3 = Registrant.new('Tucker', 15 )
+      facility_1 = Facility.new({name: 'DMV Tremont Branch', address: '2855 Tremont Place Suite 118 Denver CO 80205', phone: '(720) 865-4600'})
+      facility_2 = Facility.new({name: 'DMV Northeast Branch', address: '4685 Peoria Street Suite 101 Denver CO 80239', phone: '(720) 865-4600'})
+
+      registrant_1.license_data
+      registrant_1.permit?
+      facility_1.administer_written_test(registrant_1)
+      registrant_1.license_data
+      facility_1.add_service('Written Test')
```
jamisonordway, line 105: Similar to my above comment: all these tests should include a scenario where the relevant service is not available at the facility.

spec/facility_spec.rb
```diff
@@ -23,4 +23,276 @@
...
+      expect(registrant_1.license_data).to eq({:written=>true, :license=>true, :renewed=>true})
+      expect(facility_1.renew_drivers_license(registrant_3)).to eq(false)
+      expect(registrant_3.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+      expect(facility_1.renew_drivers_license(registrant_2)).to eq(true)
+      expect(registrant_2.license_data).to eq({:written=>true, :license=>true, :renewed=>true})
+    end
+  end
 end
+
+describe 'helper methods' do
```
jamisonordway, line 187: Great job **unit testing** your helper methods and using `describe` to organize this file.

PR #423, opened 2023-12-12
Iteration4_vehicle

hoadam:
Vehicles:
Add in a second data source for vehicles and create some Vehicle objects from NY State Registration Data: https://data.ny.gov/resource/w4pv-hbkt.json


lib/analytics.rb
```diff
@@ -0,0 +1,62 @@
...
+    model_count = Hash.new(0)
+
+    ev_registrations.each do |registration|
+      make_count[registration[:make]] += 1
+      model_count[registration[:model]] += 1
+    end
+
+    most_popular_make = make_count.max_by {|make,count| count }
+    most_popular_model = model_count.max_by {|make,count| count}
+    result = {most_popular_make: most_popular_make, most_popular_model: most_popular_model}
```
KatBrandt, line 39: Nice! I like that you used hashes for this and that you returned a hash as well. Good design.

lib/analytics.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  def count_registered_vehicles_by_model_year(ev_registrations)
+    model_year_count = Hash.new(0)
```
KatBrandt, line 43: Same thing here, I like the hash decision

lib/facility.rb
```diff
@@ -1,14 +1,72 @@
...
+      @collected_fees += 25
+    else
+      @collected_fees += 100
+    end
+  end
+
+  def register_vehicle(vehicle)
+    if @services.include?('Vehicle Registration')
+      @registered_vehicles << vehicle
+    else
```
KatBrandt, line 44: You don't have to have an else part of the if statement. You can just omit lines 44 and 45.

lib/facility.rb
```diff
@@ -1,14 +1,72 @@
...
+      return
+    end
+  end
+
+  def administer_written_test(registrant)
+    return false if !@services.include?('Written Test')
+    return false if !registrant.permit? || registrant.age < 16
```
KatBrandt, line 51: You can combine lines 50 and 51 actually. Just check all the conditions on one line and return false if they're not all satisfied.

lib/facility.rb
```diff
@@ -1,14 +1,72 @@
...
+
+    registrant.license_data[:written] = true
+    true
+  end
+
+  def administer_road_test(registrant)
+    return false if !@services.include?('Road Test')
+    return false if !registrant.license_data[:written]
```
KatBrandt, line 59: Same thing here. No sense in having these on two lines, can just combine them into one line using the || or && logical operators.

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'spec_helper'
+
+RSpec.describe FacilityFactory do
+  describe '#create_factory' do
+    it 'create an array of facilities from a given data' do
+      factory = FacilityFactory.new
+      data = [
```
KatBrandt, line 7: For this part of the project, we wanted to take the data that was coming in from the dmv_data_service and then write a method to take in that data. For example, if we have `co_data = DmvDataService.new.co_dmv_office_locations` and then we do `factory.create_facilities(co_data)`, we should have all the logic in that method to parse through the data that comes in from the dmv_data_service object.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,37 @@
+require './lib/vehicle'
+
+class VehicleFactory
```
KatBrandt, line 3: Yes, nice work here with this Vehicle Factory. This is how the Facility factory should be working as well, just taking in one giant set of data and then filtering through it.

spec/analytics_spec.rb
```diff
@@ -0,0 +1,58 @@
+require 'spec_helper'
+
+RSpec.describe Analytics do
+
+  describe '#most_popular_make_model' do
+    it 'return the most popular make and model from EV registration' do
+      analytics = Analytics.new
+      ev_registrations = [ {:make=>"Tesla", :model=>"3"},
```
KatBrandt, line 8: Where would this data be coming from though? 
I would think maybe we'd send in a bunch of Vehicle objects from a Facility, and then the method in the class would parse through that data. Unless you have another method that formats data like this. 
We need to think about where the data for the arguments is coming from. Usually all we have are the objects already made from previous classes and the methods already written.

spec/facility_spec.rb
```diff
@@ -1,26 +1,126 @@
...
+      expect(facility_1.services).to eq (['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
+    end
+  end
+
+  describe '#collect_fees' do
+    it 'fees when register a vehicle based on plate type' do
+      facility_1.collect_fees(cruz)
+      facility_1.collect_fees(bolt)
+      facility_1.collect_fees(camaro)
+
```
KatBrandt, line 40: Nice white space here to separate the chunks of this test. Helps the eye

spec/facility_spec.rb
```diff
@@ -1,26 +1,126 @@
...
+      expect(facility_1.registered_vehicles).to eq ([cruz, bolt, camaro])
+    end
+
+    it "can't register a vehicle at facility_2" do
+      facility_2.register_vehicle(bolt)
+
+      expect(facility_2.register_vehicle(bolt)).to eq (nil)
+      expect(facility_2.registered_vehicles).to eq ([])
+    end
+  end
```
KatBrandt, line 61: What are the other side effects of the register_vehicle method? We want to check that all those side effects were changed as expected. We want to see that fees were collected appropriately, registration dates were set, and plate types set. I would expect all that to be tested here in the register_vehicle describe block.

lib/facility.rb
```diff
@@ -1,14 +1,72 @@
...
+    if vehicle.plate_type == :ev
+      @collected_fees  += 200
+    elsif vehicle.plate_type == :antique
+      @collected_fees += 25
+    else
+      @collected_fees += 100
+    end
+  end
+
+  def register_vehicle(vehicle)
```
KatBrandt, line 41: So this method is actually supposed to be doing a bit more work. It should be adding the vehicle to the registered_vehicles array like you have in line 43, but it should be also setting the registration date of the vehicle and setting the plate_type of the vehicle.

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,48 @@
+require 'spec_helper'
+
+RSpec.describe VehicleFactory do
+  describe '#create_vehicles' do
+    it 'create an array of vehicles from the given data' do
+      factory = VehicleFactory.new
+      data = [{:make=>"Tesla", :model=>"Model Y", :vin_1_10=>"7SAYGDEE8P", :model_year=>"2023"},
```
KatBrandt, line 7: This is not quite the data we'd be feeding into these methods.  Again, this would be coming from the api data.

KatBrandt (commented):
Hey Hoa, overall it appears that you understand the concepts we've learned to far. You can work with hashes, enumerables and classes/objects just fine. One thing we'll want to work on is really understanding what the interaction pattern is asking for in terms of all the side effects and return values of a method. 
We also want to focus on, when writing tests, when we're feeding data into a method via an argument, where is that data coming from? This can be something to focus on with a rock/mentor/student support, to really focus on these areas. 
Overall, good work!

PR #424, opened 2023-12-13
Itiration 2

LJ9332:


lib/registrant.rb
```diff
@@ -0,0 +1,54 @@
...
+    attr_accessor :balance
+    def initialize(name, age, permit = false)
+        @name = name 
+        @age = age
+        @permit = permit
+        @year = year
+        @license_data = {:written=>false, :license=>false, :renewed=>false}
+        @vehicles = []
+        @balance = 0
+        @antique = antique?
```
jamisonordway, line 19: This line is causing your tests to fail, as the stack trace indicates when I run `bundle exec rspec`:
```
  1) Facility#initialize can initialize
     Failure/Error: @antique = antique?
     
     NoMethodError:
       undefined method `antique?' for #<Registrant:0x000000010fbc2158 @name="Bruce", @age=18, @permit=true, @year=nil, @license_data={:written=>false, :license=>false, :renewed=>false}, @vehicles=[], @balance=0>
     
               @antique = antique?
                          ^^^^^^^^
     # ./lib/registrant.rb:19:in `initialize'
     # ./spec/facility_spec.rb:7:in `new'
     # ./spec/facility_spec.rb:7:in `block (2 levels) in <top (required)>'
```
I think the important takeaway here is that a `Registrant` should not have any concept of `antique?` because that is an attribute of a `Vehicle`. The `antique?` method that does live in the Vehicle class looks good to me-- we should remove this one.

lib/registrant.rb
```diff
@@ -0,0 +1,54 @@
...
+    def balance
+      @balance = 0    
+         if @vehicle.antique?
+          25
+        if @vehicle.electric_vehicle?
+        200
+        else
+       100
+        end
+    end
```
jamisonordway, line 44: I don't think this method is in the interaction pattern, so it does not need to be implemented. The `Facility` should keep track of its own `collected_fees`, but the `Registrant` does not have a balance.

lib/registrant.rb
```diff
@@ -0,0 +1,54 @@
...
+   
+    def antique?
+      @year && @year >= 1998
+    end
```
jamisonordway, line 48: This method shouldn't exist, and the year we're comparing against should be dynamic, not hard-coded. The contents of your `Vehicle#antique?` method is what we would want.

lib/registrant.rb
```diff
@@ -0,0 +1,54 @@
...
+    
+    def electric_vehicle?
+        @engine == :ev
+    end
```
jamisonordway, line 52: Again, I would expect this information to live on `Vehicle` instead of `Registrant`.

spec/facility_spec.rb
```diff
@@ -23,4 +28,42 @@
...
+    end
+  end
+
+
+
+
+
+
+
+
```
jamisonordway, line 68: Be sure to remove unnecessary blank lines like these.

jamisonordway (commented):
I noticed some really good progress in this PR towards the project learning goals, and hopefully the comments help to explain the test failures you were running into. Overall, my main advice is to double check the interaction pattern and make sure you're defining attributes and methods on the correct classes.

PR #425, opened 2023-12-13
Iteration2 vehicle registration

NicoShanstrom:
pretty sure this is as far as i am able to get here. Brain stopped working, eyes are crossed. I tried, i really did


lib/dmv.rb
```diff
@@ -4,12 +4,16 @@ def initialize
...
+    @facilities
+  end
+
   def add_facility(facility)
     @facilities << facility
   end
 
   def facilities_offering_service(service)
-    @facilities.find do |facility|
+    @facilities.select do |facility|
```
KatBrandt, line 16: Yup! This is the perfect enumerable to use for this situation.

lib/facility.rb
```diff
@@ -1,14 +1,55 @@
...
+    
+    service_fee = case registered_vehicle.plate_type
+                  when :antique
+                    25
+                  when :ev
+                    200
+                  else
+                    100
+                  end
+    collected_fees(service_fee)
```
KatBrandt, line 39: Nice use of this as a helper method. I don't see case statements too often, they tend to be slightly harder to read than if/else statements, but this definitely works!

lib/facility.rb
```diff
@@ -1,14 +1,55 @@
...
+  end
+
+  def collected_fees(service_fee = 0)
+    @collected_fees += service_fee
+  end
+
+  def administer_written_test(registrant)
+      @written_test = true
```
KatBrandt, line 47: This is where it definitely gets complicated in the interaction pattern. This method should only return true IF, the "written test" service has been added to the facility AND the registrant has their permit AND the registrant is over 16. So we'd need to check all those conditions first before returning true.

lib/vehicle.rb
```diff
@@ -1,25 +1,53 @@
...
   def antique?
     Date.today.year - @year > 25
   end
 
   def electric_vehicle?
     @engine == :ev
   end
+
+  def registration_date(date = nil)
+      return nil if date.nil?
```
KatBrandt, line 30: this is called a guard clause and is helpful to check for conditions before running any additional code. Nice job adding one.

KatBrandt: For this method, just doing Date.today would have worked. :) Glad you got to play with .strftime though!

lib/vehicle.rb
```diff
@@ -1,25 +1,53 @@
...
+  # end
+
+  def plate_type
+    if antique?
+      :antique
+    elsif electric_vehicle?
+      :ev
+    else
+      :regular
+    end
```
KatBrandt, line 49: Yup, great method

spec/facility_spec.rb
```diff
@@ -2,25 +2,90 @@
...
+      @facility_1.add_service("Vehicle Registration")
+      expect(@facility_1.registered_vehicles).to eq([])
+      @facility_1.register_a_vehicle(@cruz)
+      expect(@facility_1.registered_vehicles).to eq([@cruz])
+      
+      @facility_1.register_a_vehicle(@camaro)
+      expect(@facility_1.registered_vehicles).to eq([@cruz, @camaro])
+      
+      @facility_1.register_a_vehicle(@bolt)
+      expect(@facility_1.registered_vehicles).to eq([@cruz, @camaro, @bolt])
```
KatBrandt, line 54: Great, nice job testing that we can add all three vehicles.

spec/facility_spec.rb
```diff
@@ -2,25 +2,90 @@
...
+
+      expect(@facility_2.registered_vehicles).to eq([])
+      @facility_2.register_a_vehicle(@bolt)
+      expect(@facility_2.registered_vehicles).to eq([])
```
KatBrandt, line 58: This is a good start to this test, to make sure we can not register a vehicle if the facility does not offer this service. Then, after line 60, you'd want to repeat line 57 and then make sure that you DO see @bolt when you call @facility.registered_vehicles.

PR #426, opened 2023-12-13
Lance Butler

LJ9332:


lib/vehicle.rb
```diff
@@ -5,14 +5,22 @@ class Vehicle
...
               :model,
-              :engine
+              :engine,
+              :registration_date
 
+  vehicle_details = {:vin => "vin",
+  :year => :year,
+  :make => :make,
+  :model => :model,
+  :engine => :engine}
```
jamisonordway, line 15: In Ruby, it's very rare to define a variable outside of a method like this. `vehicle_details` should return a hash which is populated with the attributes that you set up in the initialize method. Wrapped in a method, that might look like
```
def vehicle_details
  {
    vin: @vin,
    make: @make,
    model: @model,
    engine: @engine
  }
end
```

jamisonordway (commented):
I cloned your repo and ran your tests to replicate the failure you mentioned in the submission form-- it seems like the tests were not running because of this error in the `spec_helper.rb` file. 
```
Failure/Error: require 'pry'

LoadError:
  cannot load such file -- pry
  Did you mean?  pty
```
If this is what had you stuck, the good news is that this is straightforward to prevent in the future. Next time, be sure to add `gem 'pry'` to your `Gemfile`, then run `bundle install` in the terminal from the root of your project directory.

PR #429, opened 2023-12-13
Nicole Sanderson DMV

AuroraTuring:
Project submission.

### DIFF
--- a/lib/dmv.rb
+++ b/lib/dmv.rb
@@ -1,6 +1,7 @@
 class Dmv
+  attr_reader :facilities 
 
-  def initialize
+  def initialize()
     @facilities = []
   end
 
@@ -9,7 +10,7 @@ class Dmv
   end
 
   def facilities_offering_service(service)
-    @facilities.find do |facility|
+    @facilities.find_all do |facility|
       facility.services.include?(service)
     end
   end
--- a/lib/facility.rb
+++ b/lib/facility.rb
@@ -1,5 +1,5 @@
 class Facility
-  attr_reader :name, :address, :phone, :services
+  attr_reader :name, :address, :phone, :services, :registered_vehicles, :collected_fees
 
   def initialize(hash)
     # (hash) or (phonebook)
@@ -7,9 +7,39 @@ class Facility
     @address = hash[:address]
     @phone = hash[:phone]
     @services = []
+    @registered_vehicles = []
+    @collected_fees = 0
   end
 
   def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    if @services.include?("Vehicle Registration") 
+      @registered_vehicles << vehicle
+    end
+    collect_fees(vehicle)
+  end
+
+  def collect_fees(vehicle)
+    if vehicle.antique?
+      @collected_fees += 25
+    elsif vehicle.electric_vehicle? 
+      @collected_fees += 200
+    else
+      @collected_fees += 100
+    end
+  end
+  
+  def administer_written_test(registrant)
+    if !@services.include?("Written Test") 
+      false 
+    else
+      if registrant.permit? == true && registrant.age >= 16 
+        registrant.license_data[:written] = true 
+      end
+    end
+  end
+
 end
--- a/lib/facility.rb
+++ b/lib/facility.rb
@@ -1,14 +1,15 @@
 class Facility
   attr_reader :name, :address, :phone, :services
 
-  def initialize(name, address, phone)
-    @name = name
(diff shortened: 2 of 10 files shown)


lib/dmv.rb
```diff
@@ -9,7 +10,7 @@ def add_facility(facility)
   end
 
   def facilities_offering_service(service)
-    @facilities.find do |facility|
+    @facilities.find_all do |facility|
```
KatBrandt, line 13: Yup, this is the right enumerable for the job. :)

lib/facility.rb
```diff
@@ -1,14 +1,45 @@
...
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    if @services.include?("Vehicle Registration") 
+      @registered_vehicles << vehicle
+    end
+    collect_fees(vehicle)
```
KatBrandt, line 22: Nice job using a helper method here to do this logic for you! I would have put this inside the if block though because we wouldn't want to collect any fees if we can't actually register the vehicle.

lib/facility.rb
```diff
@@ -1,14 +1,45 @@
...
+  end
+
+  def collect_fees(vehicle)
+    if vehicle.antique?
+      @collected_fees += 25
+    elsif vehicle.electric_vehicle? 
+      @collected_fees += 200
+    else
+      @collected_fees += 100
+    end
```
KatBrandt, line 32: Yup! This method looks great.

lib/facility.rb
```diff
@@ -1,14 +1,45 @@
...
+  end
+  
+  def administer_written_test(registrant)
+    if !@services.include?("Written Test") 
+      false 
+    else
+      if registrant.permit? == true && registrant.age >= 16
```
KatBrandt, line 39: instead of doing a nested if statement here, we could have added this logic onto line 38 and instead of an `else`, had an `elsif`. Good job looking through those conditions though.

lib/vehicle.rb
```diff
@@ -5,14 +5,17 @@ class Vehicle
...
+              :engine, 
+              :plate_type
 
   def initialize(vehicle_details)
     @vin = vehicle_details[:vin]
     @year = vehicle_details[:year]
     @make = vehicle_details[:make]
     @model = vehicle_details[:model]
     @engine = vehicle_details[:engine]
+    @plate_type = plate_type
```
KatBrandt, line 17: This is an interesting way to code this actually, you are running the method of `plate_type` right on initialization and I think that's a great way to do this! I might change the name of the method to `set_plate_type` though, since our method names should be verbs. Great job. 

We would NOT want to do this for line 18, `@registration_date` however, since we should not be setting the registration date until the vehicle is registered.

spec/facility_spec.rb
```diff
@@ -2,25 +2,207 @@
...
+    end
+  end
+
+  describe '#add_service' do
+    xit 'can add available services' do
+      expect(@facility_1.services).to eq([])
+      @facility_1.add_service('New Drivers License')
+      @facility_1.add_service('Renew Drivers License')
+      @facility_1.add_service('Vehicle Registration')
+      expect(@facility_1.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
```
KatBrandt, line 31: yup, great test to show that we can add multiple elements to the array.

spec/facility_spec.rb
```diff
@@ -2,25 +2,207 @@
...
+    end
+  end
+
+
+  describe "#register_vehicle" do 
+    xit 'has a Vehicle Registration service' do
```
KatBrandt, line 37: I like that you did a describe block for the `register_vehicle` method, we could have then done multiple it blocks to describe every side effect that this method has: collecting fees, setting plate type, setting registration date, and adding vehicle to the @registered_vehicles array.

spec/facility_spec.rb
```diff
@@ -2,25 +2,207 @@
...
+
+     expect(@camaro.registration_date).to eq(Date.today)
+     expect(@camaro.plate_type).to eq(:antique)
+
+     @facility_1.register_vehicle(@bolt)
+
+     expect(@bolt.registration_date).to eq(Date.today) 
+     expect(@bolt.plate_type).to eq(:ev)
+     expect(@facility_1.registered_vehicles).to eq([@cruz, @camaro, @bolt])
+     expect(@facility_1.collected_fees).to eq(325)
```
KatBrandt, line 63: Great job showing the accumulation of fees!

PR #435, opened 2024-02-13
Grant Davis

grantdavis303:


lib/dmv.rb
```diff
@@ -1,16 +1,19 @@
...
+
+    def add_facility(facility)
+        @facilities << facility
+    end
 
-  def facilities_offering_service(service)
-    @facilities.find do |facility|
-      facility.services.include?(service)
+    def facilities_offering_service(service)
+        @facilities.find_all do |facility|
```
KatBrandt, line 14: Yup, nice use of find_all.

lib/facility.rb
```diff
@@ -1,14 +1,63 @@
...
+    def initialize(facility_info)
+        @name = facility_info[:name] 
+        @address = facility_info[:address]
+        @phone = facility_info[:phone]
+        @services = []
+        @registered_vehicles = []
+        @collected_fees = 0
+    end
+
+    # Add Service
```
KatBrandt, line 19: No need to keep these types of comments in a final draft. Ruby code should be mostly self-explanatory.

lib/facility.rb
```diff
@@ -1,14 +1,63 @@
...
+    def add_service(service)
+        @services << service
+    end
+
+    # Register a Vehicle
+    def register_vehicle(vehicle)
+        if @services.include? ("Vehicle Registration")
+            vehicle.change_registration_date
+            vehicle.update_plate_type
```
KatBrandt, line 28: I like that you created these 2 helper methods in the Vehicle class to update their attributes.

lib/facility.rb
```diff
@@ -1,14 +1,63 @@
...
+            @collected_fees += vehicle.fees
+            @registered_vehicles << vehicle
+        else
+            return nil
+        end
```
KatBrandt, line 33: If you're just returning nil in the `else`, then you can just leave out the else and only have an if...end statement. Then if your code doesn't go down the `if` branch, it will automatically return nil.

lib/facility.rb
```diff
@@ -1,14 +1,63 @@
...
+    end
+
+    # Administer a Road Test
+    def administer_road_test(registrant)
+        if @services.include? ("Road Test")
+            registrant.are_road_eligible?
+        else
+            return false
+        end
```
KatBrandt, line 51: This is good. Again, I like how you created a helper method in the Registrant class. You could refactor this even more by removing the if statement completely and simply having this in your method:
`@services.include?("Road Test") && registrant.are_road_eligble?` Just having that would return true or false for you.

spec/facility_spec.rb
```diff
@@ -1,26 +1,138 @@
...
+        @bolt = Vehicle.new({vin: '987654321abcdefgh', year: 2019, make: 'Chevrolet', model: 'Bolt', engine: :ev} )
+        @camaro = Vehicle.new({vin: '1a2b3c4d5e6f', year: 1969, make: 'Chevrolet', model: 'Camaro', engine: :ice} )
+
+        @registrant_1 = Registrant.new('Bruce', 18, true)
+        @registrant_2 = Registrant.new('Penny', 16)
+        @registrant_3 = Registrant.new('Tucker', 15)
+    end
+
+    describe '# initialize' do
+
```
KatBrandt, line 18: No white space needed here

spec/facility_spec.rb
```diff
@@ -1,26 +1,138 @@
...
+    describe '# registers a vehicle' do
+
+        it 'will not register a vehicle if service is not included' do
+            expect(@facility.registered_vehicles).to eq []
+            @facility.register_vehicle(@cruz)
+            expect(@facility.registered_vehicles).to eq []
+        end
+
+        it 'will register a vehicle if service is included' do
```
KatBrandt, line 51: Great job with the it statements here, very descriptive and testing for just one thing.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+            expect(@registrant_3.permit).to eq false
+        end  
+
+        it 'registrants have blank license data' do
+            expect(@registrant_1.license_data).to eq ({:written=>false, :license=>false, :renewed=>false})
+            expect(@registrant_2.license_data).to eq ({:written=>false, :license=>false, :renewed=>false})
+            expect(@registrant_2.license_data).to eq ({:written=>false, :license=>false, :renewed=>false})
+        end
+
+    end
```
KatBrandt, line 40: Remember that every test in this class should have it's own test file. You created helper methods in this class, but there are no tests for them. We want our helper methods to be tested as well, those are our Unit tests and the methods that use those helper methods, the tests for those are Integration tests.

lib/registrant.rb
```diff
@@ -0,0 +1,56 @@
+class Registrant
+
+    attr_reader :name, 
+                :age, 
+                :permit,
```
KatBrandt, line 5: Careful here. We do not want a method exposed called `.permit`, we have `.permit?`, so we do not want this attr_reader

lib/registrant.rb
```diff
@@ -0,0 +1,56 @@
+class Registrant
+
+    attr_reader :name, 
+                :age, 
+                :permit, 
+                :license_data
+
+    def initialize(name, age, permit = false, license_data = {})
```
KatBrandt, line 8: No need to have license_data as a 4th argument here. We just want it in the initialize like you have it on line 12 which will set the hash values for a new Registrant.

lib/registrant.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+
+    # Can Renew License
+    def are_renew_eligible?
+        if @license_data[:license] == true
+            @license_data[:renewed] = true
+        else
+            return false
+        end
+    end
```
KatBrandt, line 54: Great job with these helper methods. Just remember to test them. :)

lib/facility_factory.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    def create_facilities(locations)
+        if locations[0][:state] == "CO"
+            create_co_facilities(locations)
+        elsif locations[0][:state] == "NY"
+            create_ny_facilities(locations)
+        elsif locations[0][:state] == "MO"
+            create_mo_facilities(locations)
+        end
+    end
```
KatBrandt, line 11: More helper methods! I like that! Makes it very easy to read the control flow of the method.

lib/facility_factory.rb
```diff
@@ -0,0 +1,55 @@
...
+        locations.each do |location|
+            branch_details = {
+                name:location[:dmv_office],
+                address: "#{location[:address_li]} #{location[:address__1]}, #{location[:city]}, #{location[:state]} #{location[:zip]}",
+                phone:location[:phone]
+            }
+            facility = Facility.new(branch_details)
+            facilities_array << facility 
+        end
+        facilities_array
```
KatBrandt, line 24: Yup, this looks great! You could also have an instance variable holding the facilities_array stored in an initialize, that way that data is always stored and can be accessed repeatedly, instead of having to run this method everytime you want the data.

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+        @new_york_facilities = DmvDataService.new.ny_dmv_office_locations
+        @missouri_facilities = DmvDataService.new.mo_dmv_office_locations
+    end
+
+    describe '# initialize' do
+
+        it 'can initialize' do
+            expect(@factory).to be_an_instance_of FacilityFactory
+        end
+
```
KatBrandt, line 18: Too many white spaces. No white spaces between ends and don't need one between describe and next it block.

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+            it 'accurately returns data for Missouri locations' do
+                missouri_list = @factory.create_facilities(@missouri_facilities)
+
+                expect(missouri_list[2].name).to eq "CUBA"
+                expect(missouri_list[2].address).to eq "501 S FRANKLIN, CUBA, MO 65453"
+                expect(missouri_list[2].phone).to eq "(573) 885-6829"
+            end
+
+        end
+
```
KatBrandt, line 62: bye bye white spaces.... :)

lib/vehicle.rb
```diff
@@ -1,25 +1,54 @@
...
+        if antique?
+            @plate_type = :antique
+            @fees += 25
+        elsif electric_vehicle?
+            @plate_type = :ev
+            @fees += 200
+        else
+            @plate_type = :regular
+            @fees += 100
+        end
```
KatBrandt, line 41: Hmmm... I'm not sure I like the fees being stored here on the Vehicle object. I think that information would be more dependent on the Facility, like different states/counties will charge different amounts, so I'd rather have a method in the Facility class that will determine the fees based on the Vehicles plate type. Good control flow logic though.

KatBrandt (commented):
Overall, really great job here! You're doing great with your methods being very readable. Using helper methods is wonderful to accomplish this and you are doing that well! Just be sure to test all those helper methods. Your tests look great too, I like that you really organized your tests well into separate it and describe blocks. Well done.

PR #436, opened 2024-02-13
Billy Wallace - DMV

wallacebilly1:


lib/dmv.rb
```diff
@@ -9,8 +10,48 @@ def add_facility(facility)
...
+    dmv_location_data.map do |dmv_site|
+      dmv_facility_details = {
+        name: format_name(dmv_site),
+        address: format_address(dmv_site),
+        phone: format_phone(dmv_site)
+      }
+
+      Facility.new(dmv_facility_details)
+    end
+  end
```
KatBrandt, line 56: Yes, I do like how you have separated out all of these concerns into different helper methods. 
In terms of organization, we usually have the Integration methods up top (create_facilities in this case) and the helper methods down below, that way we can quickly see the meat of the logic and then if we need to dig into the helper methods, we can do that. Otherwise great job. 
Another thought is you could use Boolean Operators instead of if statements so something like this for the `format_phone` method:
`dmv_site[:phone] || dmv_site[:public_phone_number]`
This would look at the first expression and if it has a value, would return that value, if no value (nil), then it would look to the second expression and if there's a value there, would return that value. If no values are found on either side, this would return false.

spec/dmv_spec.rb
```diff
@@ -40,4 +40,78 @@
...
+      mo_facilities = @dmv.create_facilities(@mo_dmv_office_locations)
+
+      expect(mo_facilities).to be_an_instance_of Array
+      expect(mo_facilities.first).to be_an_instance_of Facility
+      expect(mo_facilities.first.name).to be_a String
+      expect(mo_facilities.first.address).to be_a String
+      expect(mo_facilities.first.phone).to be_a String
+      expect(mo_facilities.first.collected_fees).to eq 0
+      expect(mo_facilities.first.services).to eq []
+    end
```
KatBrandt, line 115: Good job with the testing for this. Not too much more you can do with this since we don't control the API data.

lib/facility.rb
```diff
@@ -1,14 +1,58 @@
...
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle) 
+    if @services.include? 'Vehicle Registration'
+      @registered_vehicles << vehicle
+      collect_fee(vehicle)
+      vehicle.register
+      vehicle.get_plate
```
KatBrandt, line 27: Nice job with all of these helper methods! You could also have encapsulated line 24 into a helper method to have the "job" of register_vehicle just be to delegate tasks to other helper methods.

lib/facility.rb
```diff
@@ -1,14 +1,58 @@
...
+
+  def collect_fee(vehicle)
+    if vehicle.electric_vehicle?
+      @collected_fees += 200
+    elsif vehicle.antique?
+      @collected_fees += 25
+    else 
+      @collected_fees += 100
+    end
+  end
```
KatBrandt, line 39: I like that the logic for this collect_fee method is here in the Facility class (as opposed to the Vehicle class)

lib/facility.rb
```diff
@@ -1,14 +1,58 @@
...
+
+  def administer_written_test(registrant)
+    if @services.include? 'Written Test'
+      registrant.pass_written_test
+    end
+  end
```
KatBrandt, line 45: I think these three methods needed more conditions. Not only did you need to check if the @services included something specific, but we needed to also check that the Registrant was a certain age AND also had their permit. How would you add these additional conditions in?

lib/facility.rb
```diff
@@ -1,14 +1,58 @@
...
+    elsif vehicle.antique?
+      @collected_fees += 25
+    else 
+      @collected_fees += 100
+    end
+  end
+
+  def administer_written_test(registrant)
+    if @services.include? 'Written Test'
+      registrant.pass_written_test
```
KatBrandt, line 43: I like the helper methods you created here, allowing the Registrant class itself to be in charge of changing it's own attributes.

spec/facility_spec.rb
```diff
@@ -23,4 +25,186 @@
...
+
+  describe '#collect_fee' do
+    before(:each) do
+      @facility.add_service('Vehicle Registration')
+      @cruz = Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice} )
+      @bolt = Vehicle.new({vin: '987654321abcdefgh', year: 2019, make: 'Chevrolet', model: 'Bolt', engine: :ev} )
+      @camaro = Vehicle.new({vin: '1a2b3c4d5e6f', year: 1969, make: 'Chevrolet', model: 'Camaro', engine: :ice} )
+    end
+
+    it 'can collect a fee (regular)' do
```
KatBrandt, line 82: Didn't we already have a test for collecting the fees on line 58? No need for the one on line 58 if you have all these ones here. 
I like that you have tested all possible values from this method

spec/facility_spec.rb
```diff
@@ -23,4 +25,186 @@
...
+    it 'cannot administer a written test if the registrant does not have a permit' do
+      @facility.add_service('Written Test')
+      @facility.administer_written_test(@registrant_1)
+      @facility.administer_written_test(@registrant_2)
+
+      expect(@registrant_1.license_data).to eq ({:written=>true, :license=>false, :renewed=>false})
+      expect(@registrant_2.license_data).to eq ({:written=>false, :license=>false, :renewed=>false})
+    end
+
+    it 'cannot administer a written test if the registrant is younger than 16' do
```
KatBrandt, line 136: hmmm... you have tests for these things, but I didn't see the conditions handled in the methods.

lib/registrant.rb
```diff
@@ -0,0 +1,57 @@
...
+
+	def earn_permit
+		@permit = true
+	end
+
+	def pass_written_test
+		if permit? && @age >= 16
+			@license_data[:written] = true
+		end
+	end
```
KatBrandt, line 32: ooooooohhhhh, I see, you put the conditions in this helper method. I would change the name of these methods to something more descriptive then like `pass_written_test?` or `can_take_written?` Also be sure to include the `?` since these methods return boolean values.

lib/registrant.rb
```diff
@@ -0,0 +1,57 @@
+require 'pry'
+
```
KatBrandt, line 2: What happened to the indentations in this file? The other files had spacing set to 2, this one has it set to 4. 
This can happen when cloning down a repo. Should be a setting in VS Code to make it consistent. We want to maintain that consistency throughout the whole project. I prefer a 2-count tab, but that's just me.

lib/registrant.rb
```diff
@@ -0,0 +1,57 @@
...
+class Registrant
+  attr_reader :name, 
+              :age, 
+              :permit, 
+              :license_data
+
+  def initialize(name, age, permit = false)
```
KatBrandt, line 9: indentation in this file off here too. The def initialize's end doesn't line up...

spec/registrant_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+    it 'can pass_written_test' do
+      expect(@registrant_1.written?).to be false
+      
+      @registrant_1.pass_written_test
+      
+      expect(@registrant_1.written?).to be true
+    end
+  end
+
+  describe 'pass_road_test' do
```
KatBrandt, line 50: Great job adding tests for your helper methods that you created.

lib/registrant.rb
```diff
@@ -0,0 +1,57 @@
+require 'pry'
+
+class Registrant
+  attr_reader :name, 
+              :age, 
+              :permit,
```
KatBrandt, line 6: Careful here, we do not want a method called `.permit` exposed. We have a method called `.permit?` so no need for this in the attr_reader.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,21 @@
...
+				year: info[:model_year], 
+				make: info[:make], 
+				model: info[:model], 
+				engine: :ev
+			}
+														
+			Vehicle.new(vehicle_details)
+		end
+	end
+end
```
KatBrandt, line 20: Yup, this method looks good. 
If you have nothing in the initialize method, you don't need to include it.

KatBrandt (commented):
Overall really great job here, Billy!

PR #437, opened 2024-02-13
Jared Hobson

JaredMHobson:


lib/facility.rb
```diff
@@ -1,14 +1,72 @@
...
     @services << service
   end
-end
+
+  def register_vehicle(vehicle)
+    unless vehicle.registered? || !services.include?('Vehicle Registration')
+      registered_vehicles << vehicle
+      vehicle.update_registration_date
+      vehicle.update_plate_type
+      add_collected_fees(vehicle)
```
KatBrandt, line 31: I like that you made a bunch of helper methods here to abstract all the code. Nice work.

lib/facility.rb
```diff
@@ -1,14 +1,72 @@
...
+  end
+
+  def add_collected_fees(vehicle)
+    if vehicle.antique?
+      @collected_fees += 25
+    elsif vehicle.electric_vehicle?
+      @collected_fees += 200
+    else
+      @collected_fees += 100
+    end
```
KatBrandt, line 42: One way to get ride of if/else statements is to make a "lookup table" using a hash. If you have the plate type as the key and the associated value as the fee, encapsulate that in a method, then you can use that to get rid of if statements like this.

lib/facility.rb
```diff
@@ -1,14 +1,72 @@
...
+  end
+
+  def has_service?(service)
+    services.include?(service)
+  end
```
KatBrandt, line 47: Nice abstraction!

lib/registrant.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def license?
+    license_data[:license]
+  end
+
+  def earn_permit
+    @permit = true
+  end
+
+  def take_written
```
KatBrandt, line 27: Remember that methods that return booleans should end in a `?`

spec/facility_spec.rb
```diff
@@ -1,26 +1,268 @@
...
+  describe '#has_service?' do
+    it 'can check if it has a service' do
+      expect(facility_1.services).to be_empty
+
+      facility_1.add_service('Written Test')
+      facility_1.add_service('New Drivers License')
+
+      expect(facility_1.services).to eq(['Written Test', 'New Drivers License'])
+
+      expect(facility_1.has_service?('Written Test')).to be true
```
KatBrandt, line 46: This it block is actually testing 3 methods. add_service, services, and has_service? We want to try to practice single responsibility for our tests as well. Each test should get it's own it block.

spec/facility_spec.rb
```diff
@@ -1,26 +1,268 @@
...
+
+    it 'cannot administer written tests if it has the service but the registrant does not meet the requirements' do
+      expect(registrant_2.license_data[:written]).to be false
+      expect(registrant_2.permit?).to be false
+      expect(registrant_2.age).to eq(16)
+
+      facility_1.add_service('Written Test')
+
+      expect(facility_1.administer_written_test(registrant_2)).to be false
+      expect(registrant_2.license_data[:written]).to be false
```
KatBrandt, line 77: Yeah, good job testing all the different combinations of conditions.

spec/facility_spec.rb
```diff
@@ -1,26 +1,268 @@
...
+
+      expect(facility_2.registered_vehicles).to be_empty
+
+      facility_2.register_vehicle(bolt)
+
+      expect(facility_2.registered_vehicles).to eq([bolt])
+    end
+  end
+
+  describe '#add_collected_fees' do
```
KatBrandt, line 233: Good job testing these helper methods in isolation.

lib/facility_factory.rb
```diff
@@ -0,0 +1,74 @@
...
+  def create_facility(data_service)
+    data_service.each do |facility_info|
+      formatted_facility_info = {
+        name: change_to_name_key(facility_info),
+        address: change_to_address_key(facility_info),
+        phone: change_to_phone_key(facility_info),
+        hours: change_to_hours_key(facility_info),
+        holidays_closed: change_to_holidays_key(facility_info)
+        }
+        @facilities << Facility.new(formatted_facility_info)
```
KatBrandt, line 17: I like that you stored this data in an instance variable, that way we don't have to run the method everytime we want this data.

lib/facility_factory.rb
```diff
@@ -0,0 +1,74 @@
...
+      Tuesday: #{facility_info[:tuesday_beginning_hours]} - #{facility_info[:tuesday_ending_hours]}
+      Wednesday: #{facility_info[:wednesday_beginning_hours]} - #{facility_info[:wednesday_ending_hours]}
+      Thursday: #{facility_info[:thursday_beginning_hours]} - #{facility_info[:thursday_ending_hours]}
+      Friday: #{facility_info[:friday_beginning_hours]} - #{facility_info[:friday_ending_hours]}".delete("\n")
+    elsif facility_info.has_key?(:daysopen)
+      facility_info[:daysopen]
+    else
+      'Hours unavailable.'
+    end
+  end
```
KatBrandt, line 65: Nice job with all this data. I do agree that Iteration 4 is more of the same. Remember, you are never required to do Iteration 4. Next time if when you finish Iteration 3, if you want additional work, feel free to reach out and I can suggest another project. :)

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+
+      factory.create_facility(missouri_facilities)
+
+      expect(factory.facilities[0]).to be_an_instance_of(Facility)
+      expect(factory.facilities[0].address).not_to be(nil)
+      expect(factory.facilities[0].name).not_to be(nil)
+      expect(factory.facilities[0].phone).not_to be(nil)
+      expect(factory.facilities[0].hours).not_to be(nil)
+      expect(factory.facilities[0].holidays_closed).not_to be(nil)
+    end
```
KatBrandt, line 62: Yup, can't test too much else for now. You mentioned putting in fake data to test these methods and yes, that's exactly what you can do. You could write fake api data that would actually test the function of your methods. 
We'll discuss this more in week 5.

lib/registrant.rb
```diff
@@ -0,0 +1,53 @@
+class Registrant
+  attr_reader :age, :name, :permit, :license_data
```
KatBrandt, line 2: Careful here. We did not want to expose a method called `.permit`, we have the method `.permit?` to read us the value of the attribute, but we don't want to expose methods we don't actually need.

lib/registrant.rb
KatBrandt: And maybe this method name would be more descriptive as `can_take_written?` since it's doing a check for that.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,105 @@
...
+        county: vehicle_info[:county]
+        }
+        @vehicles << Vehicle.new(formatted_vehicle_info)
+    end
+  end
+
+  def change_to_engine_key(vehicle_info)
+    if vehicle_info.has_key?(:electric_vehicle_type)
+      :ev
+    elsif vehicle_info[:fuel_type] == "GAS"
```
KatBrandt, line 25: Line 23 uses the has_key? method but here you use the bracket notation. Either one is fine, but use the same one for consistency.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,105 @@
...
+    end
+  end
+
+  def count_vehicle_makes
+    makes = Hash.new(0)
+    @vehicles.each do |vehicle|
+      makes[vehicle.make] += 1
+    end
+    makes
+  end
```
KatBrandt, line 56: Nice job here. Look into the reduce enumerable, see if you can rewrite this using that one! It's fun.

KatBrandt (commented):
Overall, really great job here!!!!

PR #439, opened 2024-02-13
Rodrigo Chavez

RodrigoACG:


lib/facility.rb
```diff
@@ -1,14 +1,64 @@
...
+              :services,
+              :registered_vehicles,
+              :collected_fees
 
-  def initialize(name, address, phone)
-    @name = name
-    @address = address
-    @phone = phone
+  def initialize(info)
+    @name = info[:name] || info[:dmv_office] || info[:office_name]
```
KatBrandt, line 10: Yup, this works! You can also abstract this in a method called something like `find_name` that has this logic in it, then line 10 could be something like `@name = find_name(info)`.

lib/facility.rb
```diff
@@ -1,14 +1,64 @@
...
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    if !@services.include?('Vehicle Registration')
+      nil
+    else
+      vehicle.registration_date = Date.today.year
+      @registered_vehicles << vehicle
+      if vehicle.antique?
```
KatBrandt, line 28: Your logic is good here. Let's try to avoid nested if statements in the future. We can achieve this by using boolean logic operatos so your if statement on line 23 could be `if @services.include?("Vehicle Registration") && `vehicle.antique?`

lib/facility.rb
```diff
@@ -1,14 +1,64 @@
...
+        @collected_fees += 200
+      else 
+        vehicle.plate_type = :regular
+        @collected_fees += 100
+      end
+    end
+  end  
+
+  def administer_written_test(registrant)
+    if @services.include?('Written Test') && registrant.age >= 16 && registrant.permit? == true
```
KatBrandt, line 42: Yeah, like this! Nice job here

lib/facility.rb
```diff
@@ -1,14 +1,64 @@
...
+      registrant.license_data[:written] = true
+    else
+      false
+    end
+  end
+
+  def administer_road_test(registrant)
+    if @services.include?('Road Test') && registrant.license_data[:written] = true
```
KatBrandt, line 50: Careful here. You have an assignment operator in this if condition which will cause that second part to always be true. Your test should have caught this.

lib/facility.rb
```diff
@@ -1,14 +1,64 @@
...
+      registrant.license_data[:license] = true 
+    else
+     false
+    end
+  end
+
+  def renew_drivers_license(registrant)
+    if @services.include?('Renew License') && registrant.license_data[:written] = true && registrant.license_data[:license] = true && registrant.age >= 16
```
KatBrandt, line 58: Here again. You're wanting the == comparison operator, not the = assignment operator

spec/facility_spec.rb
```diff
@@ -23,4 +30,131 @@
...
+  
+  describe '#Default attributes' do
+    it 'can add vehicle registration' do
+      expect(@facility.services).to eq([])
+      @facility.add_service('Vehicle Registration')
+      expect(@facility.services).to eq(['Vehicle Registration'])
+     expect(@facility.registered_vehicles).to eq([])
+     expect(@cruz.registration_date).to eq(nil)
+     expect(@facility.collected_fees).to eq(0)
+   end
```
KatBrandt, line 42: Watch your indentation here, some lines have different indentation.

lib/facility.rb
```diff
@@ -1,14 +1,64 @@
...
+      if vehicle.antique?
+        vehicle.plate_type = :antique
+        @collected_fees += 25
+      elsif vehicle.electric_vehicle?
+        vehicle.plate_type = :ev
+        @collected_fees += 200
+      else 
+        vehicle.plate_type = :regular
+        @collected_fees += 100
+      end
```
KatBrandt, line 37: The logic here is good, however this method is doing a lot. We could also refactor this and put all the fee collection stuff into it's own method.

spec/facility_spec.rb
```diff
@@ -23,4 +30,131 @@
...
+      expect(@facility.registered_vehicles).to eq([@cruz, @camaro, @bolt])
+      expect(@bolt.registration_date).to eq(Date.today.year)
+      expect(@bolt.plate_type).to eq(:ev)
+      expect(@facility.collected_fees).to eq(325)
+       
+    end
+  end
+
+  describe'#check facility and facility_2' do 
+    it 'sees what is in the first facility' do
```
KatBrandt, line 84: This message could be more specific and technical. What exactly are we looking at here? We want each it block to test a single method, or in this case, looks like we're testing that a facitiliy can register vehicles.

spec/facility_spec.rb
```diff
@@ -23,4 +30,131 @@
...
+    it 'is andministaring the road test' do 
+      expect(@facility.administer_road_test(@registrant_3)).to eq(false)
+      @registrant_3.earn_permit
+      expect(@facility.administer_road_test(@registrant_3)).to eq(false)
+      expect(@registrant_3.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+      
+      expect(@facility.administer_road_test(@registrant_1)).to eq(false)
+      @facility.add_service('Written Test')
+      @facility.add_service('Road Test')
+      expect(@facility.administer_road_test(@registrant_1)).to eq(true)
```
KatBrandt, line 139: Actually, this should be false since the registrant_3 does not yet have their written test passed. These tests were a bit complicated because there were so many specific conditions that needed to be met in different orders. YOu were on the right track here, just need to have a process/system to keep the conditions straight. I like to write them all out on paper or draw pictures. :)

lib/facility_factory.rb
```diff
@@ -0,0 +1,8 @@
+class FacilityFactory
+
+  def create_facilities(facility_data)
+    facility = facility_data.map do |facility|
+      Facility.new(facility)
+    end
+  end
```
KatBrandt, line 7: This works. You could also put the logic for checking all the keys within the data into this class, you put it into the Facility class's initialize method. Either is fine.

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    it 'shows where the dmv offices are in New york' do 
+      expect(@factory).to be_a(FacilityFactory)
+      facilities = @factory.create_facilities(@new_york_facilities)
+      expect(facilities).to be_a(Array)
+    end
+
+    it 'shows where the dmv offices are in Missouri' do 
+      expect(@factory).to be_a(FacilityFactory)
+      facilities = @factory.create_facilities(@missouri_facilities)
+      expect(facilities).to be_a(Array)
```
KatBrandt, line 27: In additional to checking that the return value is an Array, we can also test that each element in the array is indeed a Facility object. We can also check that the values were assigned properly and that none of the attributes are coming back as nil.

lib/registrant.rb
```diff
@@ -0,0 +1,20 @@
+class Registrant
+  attr_reader :name,
+              :age,
+              :license_data
```
KatBrandt, line 4: Good job not putting :permit in your attr_readers! We don't need it and don't want it.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      expect(@registrant_1.permit?).to eq(true)
+      expect(@registrant_2.permit?).to eq(false)
+    end
+  end
+  
+  describe '#earn permit' do
+    it 'earns a permit' do
+      expect(@registrant_2.permit?).to eq(false)
+      @registrant_2.earn_permit
+      expect(@registrant_2.permit?).to eq(true)
```
KatBrandt, line 41: Good job testing both the true and false return values.

KatBrandt (commented):
Overall you did a good job with this project.

PR #440, opened 2024-02-13
Steddman Bell

Steddy1Love:
Final submission


lib/facility.rb
```diff
@@ -1,14 +1,51 @@
...
+      vehicle.assign_plate 
+      @collected_fees += @fees[vehicle.plate_type]
+      @registered_vehicles << vehicle
+    else
+      "Denied"
+    end
+  end
+
+  def administer_written_test(registrant)
+    if @services.include?('Written Test') && (registrant.permit && registrant.age >=16)
```
KatBrandt, line 35: Nice job checking all these conditions.

spec/facility_spec.rb
```diff
@@ -2,25 +2,50 @@
...
+      expect(@facility_1.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
+    end
+  end
+
+  describe '#register vehicle' do
+    it 'contains registery of vehicles' do
+      expect(@facility_1.registered_vehicles).to eq([])
+    end
+
+    it 'collects fees' do
```
KatBrandt, line 36: This it block's description makes me expect that it will test collecting fees, but it only tests that the default fees are 0. Consider changing the 'it' block to something like `it "starts at 0 fees"`

lib/facility.rb
```diff
@@ -1,14 +1,51 @@
...
+    end
+  end
+
+  def administer_road_test(registrant)
+    if @services.include?('Road Test') && registrant.license_data[:written]
+      registrant.license_data[:license] = true
+    end
+  end
+
+  def renew_drivers_license(registrant)
```
KatBrandt, line 46: Where are all you tests for these methods? I didn't see them in the facility_spec.rb file. Remember you want to write tests for every method you write. These ones have a lot of conditions that need to be met so we need to write tests that test every condition combination.

lib/registrant.rb
```diff
@@ -0,0 +1,17 @@
+class Registrant
+    attr_reader :name, 
+                :age, 
+                :permit,
```
KatBrandt, line 4: Careful here, we don't want a method called `.permit`, we wanted a method called `.permit?` since that will return a boolean, so `:permit` should not be in the attr_reader but we should have a method defined below called `.permit?`

spec/registrant_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      expect(@registrant_1.license_data[:license]).to eq(true)
+    end
+
+    it 'checks for license before renewing license' do
+      @facility_1.add_service('Written Test')
+      @facility_1.administer_written_test(@registrant_1)
+      @facility_1.add_service('Road Test')
+      @facility_1.administer_road_test(@registrant_1)
+      @facility_1.addservice('Renew License')
+      @facility_1.renew_drivers_license(@registrant_1)
```
KatBrandt, line 56: These tests should actually be in the facility_spec file. The tests for the specific methods should each live in their corresponding spec file.

KatBrandt (commented):
Overall, great job! We'll want to work on making our tests more robust and figuring what test to put where.

PR #441, opened 2024-02-14
Neil Hendren

NeilTheSeal:
This pull request includes the completion of iterations 1-3, as well as the second option from iteration 4.


spec/facility_spec.rb
```diff
@@ -1,26 +1,211 @@
...
+    it "can renew licenses to those with a license" do
+      @facility1.add_service("Written Test")
+      @facility1.add_service("Road Test")
+      @facility1.add_service("Renew License")
+      @facility1.administer_written_test(@registrant)
+      @facility1.administer_road_test(@registrant)
+      can_renew = @facility1.renew_drivers_license(@registrant)
+      expect(can_renew).to eq(true)
+      expect(@registrant.license_data[:renewed]).to eq(true)
+    end
```
KatBrandt, line 183: Great job testing all of these conditions! I know it's a lot so we had a lot to test. 

Again, go ahead and use white spaces to help organize between blocks of code and even to organize within an it block. I like to put a white spaace between my setup, my execution, and my assertion so I can visually see those sections easily.

spec/facility_spec.rb
```diff
@@ -1,26 +1,211 @@
...
+      age: 18,
+      permit: true
+    })
+  end
+  before(:all) do # rubocop:disable Metrics/BlockLength
+    @co_office = DmvDataService.new.co_dmv_office_locations[0]
+    @ny_office = DmvDataService.new.ny_dmv_office_locations[0]
+    @mo_office = DmvDataService.new.mo_dmv_office_locations[0]
+
+    @co_facility = Facility.new({
```
KatBrandt, line 49: This code should actually be living in a method within the code, rather than the test. Maybe in some kind of FacilityFactory that is taking in the raw API data and converting that raw data into Facility objects. Then you'd be able to write tests to make sure the method itself is working. Here, it's almost like you're testing your test.

lib/registrant.rb
```diff
@@ -0,0 +1,35 @@
...
+#
+# registrant = Registrant.new({name: "Bill", age: 18})
+#
+# Methods:
+#
+# registrant.earn_permit
+# registrant.permit?
+class Registrant
+  attr_reader :name, :age, :permit
+  attr_accessor :license_data
```
KatBrandt, line 16: Moving forward, see if you can avoid using attr_accessors :) How else would you create the logic in this class to alter the attribute? 
Also, careful with the `:permit` in the attr_reader, we didn't want to expose a method called `.permit` so we don't want it in the attr_reader.

lib/registrant.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def initialize(info)
```
KatBrandt, line 18: This definitely works, however we want to pay attention to the specs and requirements of a project. The interaction pattern asked for a Registrant to be initialized with 2-3 arguments, positional arguments, rather than a hash.

lib/registrant.rb
```diff
@@ -0,0 +1,35 @@
...
+    @name = info[:name]
+    @age = info[:age]
+    @permit = info[:permit] || false
+    @license_data = { written: false, license: false, renewed: false }
+    return if info[:license_data].nil?
```
KatBrandt, line 23: hmm, I'm not sure why you have this guard clause. Line 22 sets the license_data for us so we don't need to look into the arguments for this data.

lib/registrant.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    info[:license_data].each_pair { |key, value| @license_data[key] = value }
+  end
+
+  def earn_permit
+    @permit = @age >= 16
```
KatBrandt, line 29: Interesting solution, nice.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    })
+    @registrant2 = Registrant.new({
+      name: "Penny",
+      age: 15
+    })
+    @registrant3 = Registrant.new({
+      name: "Bobby",
+      age: 105,
+      permit: false,
+      license_data: {
```
KatBrandt, line 21: We wanted the license_data to just be a default value whenever the Registrant is initialized. So we don't even want the option to send this in as an argument. If we need to change the values, we can write a method to accomplish that.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,105 @@
...
+
+  def consolidate_all_vehicle_data
+    retrieve_vehicle_makes_and_models
+    retrieve_vehicle_counties
+  end
+
+  def write_vehicle_data_to_json_file
+    File.open("vehicle_data.json", "w") do |file|
+      file.write(JSON.pretty_generate(vehicle_data))
+    end
```
KatBrandt, line 103: Well that's fun. Nice working writing JSON to another file. Very cool.

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,159 @@
...
+        make_data.each_value do |model|
+          makes[make] += model[:total]
+        end
+      end
+      total_vehicles = 0
+      makes.each_value do |vehicle_count|
+        total_vehicles += vehicle_count
+      end
+      expect(total_vehicles).to eq(@vehicle_factory.vehicles.count)
+    end
```
KatBrandt, line 66: For this test lines 51-65, there's too much dynamic code which can causes testing issues with results being off because the code in the test itself is off. Then we'd need to test the test. We'll learn more ways next week of how we can get around this issue when we're using data we don't control, but typically in our tests, we want to be more explicit of what the return values should be.

vehicle_data.json
```diff
@@ -0,0 +1,369 @@
+{
```
KatBrandt, line 1: Yeah, this output is cool. Usually our format for API data are nested arrays, so for example, all the makes would each be elements in an array where there is a key of `name` or `make`, then there's a nested hash structure inside of that. We'll learn more about that in mod 3. Great work!

KatBrandt (commented):
Overall great job here. It's clear you have a solid grasp of how to code and how everything is working. I'd encourage you to start thinking about refactoring, adhering to SRP, and to use white spaces in your code for code organization.

PR #442, opened 2024-02-14
Aurora Sanderson

AuroraTuring:


lib/analytics.rb
```diff
@@ -0,0 +1,20 @@
...
+  def most_popular_car
+    popular_vehicles = []
+    factory = VehicleFactory.new
+    registration = dmv_data_service.wa_ev_registrations
+    all_vehicles = factory.create_vehicles(registration)
+    specific_vehicles = all_vehicles.map do |vehicle|
+      vehicle.make + " " + vehicle.model
+    end
+    specific_vehicles.group_by(&:itself).values.max_by(&:size).first
+  end
```
KatBrandt, line 17: Great job making a separate class for this. This most_popular_car method is great overall. I think line 16 could probably be broken apart into different pieces so we can more clearly see what is going on. I've never seen the `itself` keyword before, very interesting!

lib/dmv_factory.rb
```diff
@@ -0,0 +1,41 @@
...
+    buildings.map do |building|
+      facility_info = {}
+      facility_info[:id] = building[:dmv_id]
+      facility_info[:address] = building[:address_li]
+      facility_info[:zip] = building[:zip]
+      facility_info[:phone] = building[:phone]
+      dmv_buildings << facility_info
+      # DmvFactory.new(facility_info)
+    end
+    dmv_buildings
```
KatBrandt, line 14: Hmm... what is the return value of this method now? It looks like we just have an array of hashes, but we wanted to return an array of Facility objects. We need one more step here, on line 11, we need this to say `dmv_buildings << Facility.new(facility_info)` so that we're creating Facility objects. Otherwise this looks good. 
Also, don't need to use .map if we're using the accumulator. Or use no accumulator and .map

lib/facility.rb
```diff
@@ -1,14 +1,59 @@
...
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    if !vehicle.registration_date
+        vehicle.registration_date = Date.today
+        collect_fees(vehicle)
```
KatBrandt, line 20: Nice job putting the collect_fees logic into it's own method.

lib/facility.rb
```diff
@@ -1,14 +1,59 @@
...
+      @collected_fees += 25
+    elsif vehicle.electric_vehicle?
+      @collected_fees += 200
+    else
+      @collected_fees += 100
+    end
+  end
+
+  def administer_written_test(registrant)
+    if @services.include?("Written Test") && registrant.age >= 16 && registrant.permit?
```
KatBrandt, line 36: Good job with checking all of these conditions.

spec/facility_spec.rb
```diff
@@ -2,25 +2,153 @@
...
+      expect(@facility_1.register_vehicle(@camaro)).to eq([@cruz, @camaro])
+      expect(@camaro.registration_date).to eq(Date.today)
+      expect(@camaro.plate_type).to eq(:antique)
+
+      expect(@facility_1.register_vehicle(@bolt)).to eq([@cruz, @camaro, @bolt])
+      expect(@bolt.registration_date).to eq(Date.today)
+      expect(@bolt.plate_type).to eq(:ev)
+
+      expect(@facility_1.registered_vehicles).to eq([@cruz, @camaro, @bolt])
+      expect(@facility_1.collected_fees).to eq(325)
```
KatBrandt, line 72: Great job with this test, showing the accumulation of fees and registering vehicles.

spec/facility_spec.rb
```diff
@@ -2,25 +2,153 @@
...
     end
   end
+
+  describe '#register_vehcle' do
+    it 'stores registered vehicles' do
+      expect(@cruz.registration_date).to eq(nil)
+      expect(@facility_1.registered_vehicles).to eq([])
+    end
+
+    it 'collects fees' do
```
KatBrandt, line 45: Watch the description of your it blocks, this one implies that we'll be testing it collecting fees in this test, but really this it block is just testing the default value of the collected_fees attribute, so maybe this it block would be `it "has 0 collected_fees by default"`

spec/facility_spec.rb
```diff
@@ -2,25 +2,153 @@
...
+
+      expect(@facility_2.registered_vehicles).to eq([])
+      expect(@facility_2.services).to eq([])
+      expect(@facility_2.register_vehicle(@bolt)).to eq(nil)
+      expect(@facility_2.registered_vehicles).to eq([])
+      expect(@facility_2.collected_fees).to eq(0)
+    end
+  end
+
+  describe '#administer_written_test and #administer_road_test and #renew_drivers_license' do
```
KatBrandt, line 87: I agree, this test is way too big. There are three conditions you are testing in the it block on line 88. You could break those up into mutliple it blocks. One for each condition. Yes that would require more setup, but it would be much easier to read the context of each situation with this broken down into different it blocks.

spec/facility_spec.rb
```diff
@@ -2,25 +2,153 @@
...
+      expect(@registrant_2.age).to eq(16)
+      expect(@registrant_2.permit?).to be false
+      expect(@facility_1.administer_written_test(@registrant_2)).to be false
+
+      @registrant_2.earn_permit
+
+      expect(@facility_1.administer_written_test(@registrant_2)).to be true
+      expect(@registrant_2.license_data).to eq({:written=>true, :license=>false, :renewed=>false})
+
+      # ~~~~~~~~~~~~~~~~~~~~~~
```
KatBrandt, line 108: this is where another it block could go.

lib/registrant.rb
```diff
@@ -0,0 +1,19 @@
+class Registrant
+  attr_reader :name, :age, :permit, :license_data
```
KatBrandt, line 2: Careful here, we didn't want a method called `.permit` so we don't want it in the attr_reader.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,15 @@
...
+    registrations.map do |registration|
+      vehicle_info = {engine: :ev, plate_type: :ev}
+      vehicle_info[:vin] = registration[:vin_1_10]
+      vehicle_info[:year] = registration[:model_year].to_i
+      vehicle_info[:make] = registration[:make]
+      vehicle_info[:model] = registration[:model]
+      vehicle_info[:registration_date] = registration[:transaction_date]
+
+      Vehicle.new(vehicle_info)
+    end
```
KatBrandt, line 13: This one looks good! You did a map with no accumulators, perfect, and you changed all the data into Vehicle objects. This is what those Facility Factory methods should look like.

KatBrandt (commented):
Overall, great job here!

spec/facility_spec.rb
AuroraTuring: I agree! Repeating the setup is what I was trying to avoid, but it ended up making things very long and complicated.

spec/facility_spec.rb
AuroraTuring: My thoughts exactly! :)

PR #445, opened 2024-03-29
Feat/registrant class

noahdurbin:
added registrant class with tests using TDD
- initiallizes with name, age, permit, and license data
- requires name and age arguments
- can earn a permit with the earn permit method


lib/dmv.rb
```diff
@@ -1,16 +1,17 @@
 class Dmv
+    attr_reader :facilities
 
-  def initialize
-    @facilities = []
-  end
+    def initialize
```
juliet-e, line 4: It looks like your indentation is a bit inconsistent. When you're in VSCode, the bottom right of the screen should show you how many spaces it's using to indent. Ideally, we want 2 spaces. But the most important thing is that it's consistent, and this seems a bit inconsistent (some indents are very large)

PR #454, opened 2024-04-02
Noah Durbin

noahdurbin:


lib/facility.rb
```diff
@@ -1,14 +1,59 @@
...
+        @services = []
+        @collected_fees = 0
+        @registered_vehicles = []
+    end
+
+    def add_service(service)
+        @services << service
+    end
+
+    def register_vehicle(vehicle)
```
juliet-e, line 23: Excellent work structuring these methods to be clean and readable. Despite their complexity, you're making great use of helper methods to  break things up. Looks awesome!

lib/facility_factory.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    def create_facilities(facilities)
+        format_data(facilities)
+        facilities.map do |facility|
+            Facility.new(facility)
+        end
+    end
+
+    def format_data(facilities)
+        if facilities.first.has_key?(:dmv_office)
```
juliet-e, line 14: Creating a different helper method for each state is an excellent way to break up this method. Well done!

lib/registrant.rb
```diff
@@ -0,0 +1,22 @@
+class Registrant
+
+    attr_reader :name,
+                :age,
+                :permit,
+                :license_data
```
juliet-e, line 6: Minor note on this attribute -- you are assigning allowing classes outside this class (specifically the Facility class) assign values to keys in the license_data directly (`license_data[:written] = true` for example)

For the purposes of this project, allowing this manipulation or using an `attr_accessor` is just fine. Generally, though, we want to stay away from allowing outside classes to mutate our data. Another alternative here is to create `setter_methods`. For example, if a registrant takes the written exam, you could create a method here called `pass_written_exam` and inside that method, you'd set `licence_data[:written] = true`. In this way, we can control the setting of this value to make sure it's not something random (like it won't end up being `license_data[:written] = yes` or something inconsistent). We could also add controls to this method, for example, if we got a score argument, we could only set the `true` value `if score > 80`. Using a setter method gives us more control over setting this value rather than letting outside code change it to whatever it wants. 

Again, this simple of a case is no problem. Just wanted to share this context for your learning beyond these projects! This discussion revolves around the theme of encapsulation (an OOP principle)

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,139 @@
+require 'spec_helper'
+
+describe FacilityFactory do
+    before do
+        @facility_factory = FacilityFactory.new
+        @co_sample =
```
juliet-e, line 6: Using test data like this is a great way to control the data we're getting (rather than calling the API and just using its response) so that we can make more specific assertions. However, this test file then becomes VERY long. Instead of defining the data here, you could use `File.read()` and put this data into another file. We'll learn more about reading from other files in our group project so you'll have a chance to practice this.

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,139 @@
...
+        ny_facility_1 = ny_facilities.first
+
+        expect(ny_facilities).to be_an(Array)
+        expect(ny_facilities).to all be_a(Facility)
+        expect(ny_facility_1.name).to be_a(String)
+        expect(ny_facility_1.phone).to be_a(String)
+        expect(ny_facility_1.address).to be_a(String)
+    end
+
+    it "formats NY data correctly" do
```
juliet-e, line 113: This is super robust testing! Awesome! In my opinion, it's also okay to just do one of these -- test for the exact values (since you have them) OR if you were just calling the API and weren't sure of the exact values, you could test for the correct data types in each of these fields. You don't necessarily need to do both.

spec/facility_spec.rb
```diff
@@ -1,26 +1,156 @@
...
+          expect(registrant_1.license_data[:renewed]).to eq(true)
+      end
+
+      it 'will not renew if there is no license' do
+          registrant_2 = Registrant.new('Penny', 16)
+
+          @facility_1.renew_license(registrant_2)
+
+          expect(registrant_2.license_data[:renewed]).to eq(false)
+      end
```
juliet-e, line 154: Excellent testing! I appreciate your use of describe blocks to categorize the tests.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    it "starts without any license data" do
+       expect(@registrant_1.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+    end
+
+    it "can earn a permit if have passed the written test and is over 16" do
+        @registrant_2.earn_permit
+        expect(@registrant_2.permit?).to eq(true)
+    end
+
+    #write a test for the helper method permit?
```
juliet-e, line 29: Because you are explicitly testing this in your last test, there's no need to add another testing! You're verifying that it's returning the correct output on line 26 and 17

lib/facility_factory.rb
```diff
@@ -0,0 +1,43 @@
...
+    def co_format(facilities)
+        facilities.map do |facility|
+            facility[:name] = facility.delete :dmv_office
+            facility[:address] = facility.values_at(:address_li, :address__1, :city, :state, :zip).join(" ")
+        end
+    end
+
+    def ny_format(facilities)
+        facilities.map do |facility|
+            facility[:name] = facility.delete :office_name
```
juliet-e, line 32: Whether you map the existing facility hash or create a new hash with details and the key names you'd like, it doesn't really matter. This implementation is great

PR #455, opened 2024-04-02
James McHugh

jdmchugh111:


lib/facility.rb
```diff
@@ -1,14 +1,74 @@
...
   end
+
+  def register_vehicle(vehicle)
+    if @services.include?("Vehicle Registration")
+      vehicle.registration_date = Date.today
+      reg_fees = {:antique => 25, :ev => 200, :regular => 100} 
+      @collected_fees += reg_fees[vehicle.plate_type]
+      @registered_vehicles << vehicle
+    end
+  end
```
KatBrandt, line 36: Interesting choice here to make a variable on line 32 that holds the amounts of the fees!!! This is actually a common practice, however, instead of storing it in a local variable, you would create a method called `reg_fees` and then just have that hash living in the method, so you could call the method from any method to read the amount of the fee.

lib/facility.rb
```diff
@@ -1,14 +1,74 @@
...
+    if @services.include?("Written Test")
+      if registrant.age >= 16
+        if registrant.permit?
+        registrant.license_data[:written] = true
+        end
+      end
+    else
+      false
+    end
+  end
```
KatBrandt, line 48: This works, for sure, and we want to try to avoid nested if statements when possible. How could you rewrite this, using Boolean operators `&&` and `||` to un-nest this if statement?

spec/facility_spec.rb
```diff
@@ -1,26 +1,212 @@
...
-      expect(@facility.services).to eq([])
-      @facility.add_service('New Drivers License')
-      @facility.add_service('Renew Drivers License')
-      @facility.add_service('Vehicle Registration')
-      expect(@facility.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
+      expect(@facility_1.services).to eq([])
+      @facility_1.add_service('New Drivers License')
+      @facility_1.add_service('Renew Drivers License')
+      @facility_1.add_service('Vehicle Registration')
+      expect(@facility_1.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
```
KatBrandt, line 31: Yes, I like this test because it shows the starting value of the services array, then you run some command methods, and then we see the ending value of the services array. Also, nice use of the nested describe blocks for organization.

KatBrandt: Actually, in looking through the rest of your tests, you have all your it blocks nested under the describe "#add_service" block. I would want each describe block to only hold it blocks that have tests that are specific to that method. So you might have another describe block for "#register_vehicle" and then all the it blocks within that describe would test all the separate side effects of that register_vehicle method.

spec/facility_spec.rb
```diff
@@ -1,26 +1,212 @@
...
+
+    expect(@registrant_1.license_data).to eq({:written=>true, :license=>false, :renewed=>false})
+  end
+
+  it "won't allow registrant to take written test if they don't have a permit" do
+    @facility_1.add_service("Written Test")
+    @facility_1.administer_written_test(@registrant_2)
+
+    expect(@registrant_2.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+  end
```
KatBrandt, line 142: Yeah, good job testing all the conditions for these methods. Since there are multiple conditions, we want to ensure we're testing all combinations of these conditions to ensure our methods function as expected.

lib/facility_generator.rb
```diff
@@ -0,0 +1,29 @@
...
+            else
+                Facility.new({name: branch[:name], 
+                            address: "#{branch[:address1]} #{branch[:city]} #{branch[:state]} #{branch[:zipcode]}", 
+                            phone: branch[:phone],
+                            daily_hours: branch[:daysopen],
+                            holidays_closed: branch[:holidays_closed]})
+            end
+        end
+    end
+end
```
KatBrandt, line 29: Yes this works... and it's a huge method! I would like to see you start refactoring your methods to be smaller. We want to follow the Single Responsibility Principle where each method should only be doing one thing. So maybe you have a method called `set_mo_facility` that would be on line 7 and all the code from lines 7-10 would live in that helper method, same with each section of your if/else block. This makes it easier to read and easier to debug.

spec/facility_generator_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+        expect(generator.create_facilities(co_dmv_office_locations)[0].address).not_to eq(nil)
+        expect(generator.create_facilities(co_dmv_office_locations)[0].phone).not_to eq(nil)
+        expect(generator.create_facilities(ny_dmv_office_locations)[0].name).not_to eq(nil)
+        expect(generator.create_facilities(ny_dmv_office_locations)[0].address).not_to eq(nil)
+        expect(generator.create_facilities(ny_dmv_office_locations)[0].phone).not_to eq(nil)
+        expect(generator.create_facilities(mo_dmv_office_locations)[0].name).not_to eq(nil)
+        expect(generator.create_facilities(mo_dmv_office_locations)[0].address).not_to eq(nil)
+        expect(generator.create_facilities(mo_dmv_office_locations)[0].phone).not_to eq(nil)
+    end
+end
```
KatBrandt, line 32: Yeah, nice job with testing this. This is really all you _can_ test with external data that we don't control.

lib/registrant.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    def initialize(name, age, permit = false)
+        @name = name
+        @age = age
+        @permit = permit
+        @license_data = {:written=>false, :license=>false, :renewed=>false}
+    end
+
+    def permit?
+        @permit == true
```
KatBrandt, line 14: Since `@permit` is storing a boolean value, you do not need to do a comparison operator to another boolean. The contents of the `permit?` method could just be `@permit`

lib/vehicle.rb
```diff
@@ -22,4 +30,16 @@ def antique?
   def electric_vehicle?
     @engine == :ev
   end
+
+  def plate_type
+    if registration_date != nil
+      if antique?
```
KatBrandt, line 36: Again, moving forward, try to avoid nested if statements as they are process heavy. We can usually refactor this using our boolean operators.

lib/vehicle.rb
```diff
@@ -22,4 +30,16 @@ def antique?
...
+      @plate_type = :antique
+      elsif electric_vehicle?
+      @plate_type = :ev
```
KatBrandt, line 39: Remember to indent the lines inside the elsif blocks.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,25 @@
...
+                                year: vehicle[:model_year], 
+                                make: vehicle[:make], 
+                                model: nil, 
+                                engine: vehicle[:fuel_type],
+                                county: vehicle[:county]})
+                end
+            end
+        end
+    end
+end
```
KatBrandt, line 25: Same feedback here as in the facility_generator class. Use helper methods to reduce your method length and increase readability. 
Good job reading in this external data!

ev_registrations_runner.rb
```diff
@@ -0,0 +1,51 @@
...
+    county_array = []
+    vehicles.each do |vehicle|
+        county = vehicle.county
+        county_array << county
+    end
+    tallied = county_array.tally
+    sorted = tallied.sort_by {|_key, value| value}
+    puts "The county with the most registered vehicles is #{sorted.last[0]} county with #{sorted.last[1]} vehicles registered."
+end
+
```
KatBrandt, line 50: Nice work with these analytics!! It's clear you understand how all this code is functioning, how to use enumerables, etc. I encourage you to start taking an additional step to refactor your code after it's working by moving sections of code to helper methods. :) If you're not sure how to do that exactly, I recommend buddying up with a mentor or housemate and asking for someone to show you how you can best refactor these methods. Great work!

PR #457, opened 2024-04-02
Final submission

clydeautin:
Submitting my DMV project. Iterations 1-3 completed. Iteration 4 started but incomplete


lib/facility.rb
```diff
@@ -1,14 +1,77 @@
...
+        cost = 100
+        plate_type = :regular
+      end
+
+    vehicle.set_plate_type(plate_type)
+    vehicle.set_registration_date(Date.today)
+    @collected_fees += cost
+    @registered_vehicles << vehicle
+
+    end
```
KatBrandt, line 45: This all works, great job working through the logic of everything this method needed to do. A few notes:
1. Ideally we avoid nested if statements when possible. How could you refactor line 27 and 29 using Boolean Operators `&&` and/or `||` to remove the nested if?
2. This method is doing a lot. Moving forward, let's try to figure out how to get methods to adhere to the Single Responsibility Principle which means every method should do one thing. You could break this method up be creating helper methods that would store pieces of this logic and then call those methods from inside the register_vehicle class. If you're not sure what this could look like, I encourage you to reach out to a mentor or housemate and ask to pair with them to see how you might refactor this.

lib/facility.rb
```diff
@@ -1,14 +1,77 @@
...
+        plate_type = :antique
+      elsif vehicle.electric_vehicle?
+        cost = 200
+        plate_type = :ev
+      else
+        cost = 100
+        plate_type = :regular
+      end
+
+    vehicle.set_plate_type(plate_type)
```
KatBrandt, line 40: Indentation is off on lines 40-43, needs to be tabbed over one.

lib/facility.rb
```diff
@@ -1,14 +1,77 @@
...
+    end
+  end 
+
+  
+  def vehicle_registered?(vehicle)
+    @registered_vehicles.include?(vehicle)
+  end
+
+  def administer_written_test(registrant)
+    if @services.include?('Written Test') && registrant.age >= 16 && registrant.permit == true
```
KatBrandt, line 54: Yes! This looks great and this is exactly how you would avoid nested if statements. Nice work with this! The only part I would fix, is I would call `registrant.permit?` at the end, since that returns a boolean, we want to use our method that ends in a `?`

spec/facility_spec.rb
```diff
@@ -23,4 +29,173 @@
       expect(@facility.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
     end
   end
+
+  describe '#register' do
```
KatBrandt, line 33: I like that you used the describe block here to organize your code. All the it blocks nested inside this describe are related to the side effects the register_vehicle method have. However, I don't see the it blocks for the collected fees par. Make sure to include ALL the side effects in this section

spec/facility_spec.rb
```diff
@@ -23,4 +29,173 @@
...
+
+  describe '#add_service' do
+    it 'can add a service' do
+      @facility.add_service('Written Test')
+
+      expect(@facility.services).to eq(['Written Test'])
+    end
+  end
+
+  describe '#cost' do
```
KatBrandt, line 77: I'm glad you still tested the fees here, but again, might make more sense to put them in the register_vehicle describe block.

spec/facility_spec.rb
```diff
@@ -23,4 +29,173 @@
...
+        end
+
+    it 'will not assign a written test to anyone under the age of 16' do
+      registrant_3 = Registrant.new('Tucker', 15, true )
+      
+      @facility.administer_written_test(registrant_3)
+  
+      expect(registrant_3.license_data[:written]).to eq(false)
+      end
+  end
```
KatBrandt, line 132: Great job here testing all the combinations of conditions to test all the ways this method could function.

lib/facility_factory.rb
```diff
@@ -0,0 +1,41 @@
+class FacilityFactory
+
+    def initialize
```
KatBrandt, line 3: If your initialize isn't doing anything, you actually don't even need to include it! :)

lib/facility_factory.rb
```diff
@@ -0,0 +1,41 @@
...
+            facility_details = {
+                name: dmv[:number],
+                address: dmv[:address1],
+                phone: dmv[:phone],
+                services: []
+            }
+            Facility.new(facility_details)
+        end
+    end
+end
```
KatBrandt, line 41: Yup, this works! Nice job reading in the api data and making Ruby objects

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+        expect(factory.create_ny_facility(ny_dmvs)).to be_an_instance_of(Array)
+    end
+
+    it 'can take a Missouri DMV info from an API and format it for our needs' do
+        factory = FacilityFactory.new
+        mo_dmvs = DmvDataService.new.mo_dmv_office_locations
+
+        expect(factory.create_mo_facility(mo_dmvs)).to be_an_instance_of(Array)
+    end
+end
```
KatBrandt, line 36: Yeah, there isn't muich we can test when using external data, however, you could also test that the return values not only are Arrays, but that the elements of those arrays are Facility objects.

lib/fun_facts.rb
```diff
@@ -0,0 +1,18 @@
+class FunFacts
+
+    def initialize
+    end
+
+    def most_popular_make(api)
```
KatBrandt, line 6: You're on the right track here! Instead of this reading in api data though, you'd want to send it the array of Vehicle objects you created from the api data in Iteration 2, that way you're dealing with Ruby objects, not nested hashes.

lib/registrant.rb
```diff
@@ -0,0 +1,23 @@
+class Registrant
+
+    attr_reader :name,
+                :age,
+                :permit,
```
KatBrandt, line 5: Careful here, we actually do not want to include permit in the attr_readers. We have a method called `permit?` which acts as an attr_reader, but follows our convention of having a `?` at the end since the method returns a boolean, so we do not want this extra method here.

lib/vehicle.rb
```diff
@@ -22,4 +26,12 @@ def antique?
   def electric_vehicle?
     @engine == :ev
   end
+
+  def set_registration_date(date)
+    @registration_date = date
+  end
+
+  def set_plate_type(plate_type)
```
KatBrandt, line 34: I like that you used these methods instead of using attr_accessors. Nice job.

spec/vehicle_spec.rb
```diff
@@ -32,5 +37,32 @@
...
+  end
+
+  describe '#plate_type' do
+  it 'can set a plate_type' do
+      @facility = Facility.new({name: 'DMV Tremont Branch', address: '2855 Tremont Place Suite 118 Denver CO 80205', phone: '(720) 865-4600'})
+      @facility.add_service('Vehicle Registration')
+      @facility.register_vehicle(@cruz)
+
+      expect(@cruz.set_plate_type(:regular)).to eq(:regular)
+    end
```
KatBrandt, line 66: Great job testing your additional helper methods

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,18 @@
...
+            vin: car[:vin_1_10],
+            year: car[:model_year],
+            make: car[:make],
+            model: car[:model],
+            engine: :ev
+            }
+            Vehicle.new(vehicle_details)
+        end
+    end
+end
```
KatBrandt, line 18: Yup! Nice work with this.

PR #458, opened 2024-04-02
DMV Project Turn-in

GBowman1:
- DMV project complete
- submit pull request for review


lib/facility.rb
```diff
@@ -1,14 +1,85 @@
...
+      @collected_fees = 25
+    elsif vehicle.electric_vehicle?
+      @collected_fees = 200
+    else
+      @collected_fees = 100
+    end
+  end
+
+  def administer_written_test(registrant)
+    if @services.include?('Written Test')
```
juliet-e, line 54: Be careful about the interaction pattern and project specs here: we only want to administer the written test if the registrant has a permit and is old enough! Your if statement would need to be a bit more complex
`    if @services.include?('Written Test') && registrant.age >= 16 && registrant.permit?`

lib/facility.rb
```diff
@@ -1,14 +1,85 @@
...
+      registrant.license_data[:written] = true
+    else
+      "Facility does not offer this service"
+    end
+  end
+
+  def administer_road_test(registrant)
+    if @services.include?('Driving Test') 
+      if registrant.license_data[:written] == true
```
juliet-e, line 63: This level of nested if statements is okay because the conditions are simple, and this allows you to give specific error messages. Just be wary of doing any *further* nesting or making the conditions more complex as that would probably get unreadable. But this looks great!

lib/facility_generator.rb
```diff
@@ -0,0 +1,29 @@
...
+                Facility.new(data)
+            else data[:state] == 'MO'
+                data[:address] = "#{data[:address1]}, #{data[:city]}, #{data[:state]} #{data[:zipcode]}"
+                data[:hours] = data[:daysopen]
+                Facility.new(data)
+            end
+            # is key mapping more efficient than this?
+            # pros : looks cleaner, easier to read
+            # cons : still have to write out new api keys
+        end
```
juliet-e, line 27: Making a new hash with these keys is not necessarily more efficient than this, but might be a bit more readable. One major way to break this up and make it more readable would be to introduce helper methods based on the state:
```
if data[:state] == "MO"
     generate_mo_facility(data)
elseif data[:state] == "NY"
     generate_ny_facility(data)
...
```
and then your helper method could look like this:
```
def generate_mo_facility(data)
    facility_data = {
        name: data[:dmv_office]
        address: "#{data[:address_li]}, #{data[:address_1]}, #{data[:city]}, #{data[:state]} #{data[:zip]}"
        ....
    }
    Facility.new(facility_data)
}
```

lib/registrant.rb
```diff
@@ -0,0 +1,20 @@
+class Registrant
+    attr_reader :name, :age, :permit
+    attr_accessor :license_data
```
juliet-e, line 3: For the purposes of this project, using an `attr_accessor` is just fine. Generally, though, we want to stay away from using these because they allow outside classes to mutate our data. Another alternative to using `attr_accessor` here is to create `setter_methods`. For example, if a registrant takes the written exam, you could create a method here called `pass_written_exam` and inside that method, you'd set `licence_data[:written] = true`. In this way, we can control the setting of this value to make sure it's not something random (like it won't end up being `license_data[:written] = yes` or something inconsistent). We could also add controls to this method, for example, if we got a score argument, we could only set the `true` value `if score > 80`. Using a setter method gives us more control over setting this value rather than letting outside code change it to whatever it wants. 

Again, this simple of a case is no problem. Just wanted to share this context because you may hear discussion of why attr_accessors are not always the best choice

lib/vehicle.rb
```diff
@@ -5,14 +5,20 @@ class Vehicle
...
 
   def initialize(vehicle_details)
     @vin = vehicle_details[:vin]
     @year = vehicle_details[:year]
     @make = vehicle_details[:make]
     @model = vehicle_details[:model]
     @engine = vehicle_details[:engine]
+    @county = vehicle_details[:county]
+    @plate_type = nil
+    registration_date = nil
```
juliet-e, line 21: It's typically more conventional to set this to an instance variable here (@registration_date) rather than in your last method

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,39 @@
...
+            data[:vin] = data[:vin_1_10]
+            data.delete(:vin_1_10)
+            data[:year] = data[:model_year]
+            data.delete(:model_year)
+            data[:engine] = :ev
+            Vehicle.new(data)
+        end
+    end
+
+    def most_popular_makemodel(vehicles)
```
juliet-e, line 13: Nice work getting to these stats! This is more of a nitpick, but it could be worth putting them in another class (VehicleStatsGenerator) or something because typically our Factory classes are *just* for making new objects.

spec/facility_spec.rb
```diff
@@ -23,4 +23,165 @@
...
+
+      expect(cruz.plate_type).to eq('regular')
+    end
+  end
+
+  describe 'update_registered_vehicles' do
+    it 'can update array of vehicles' do
+      cruz = Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice} )
+
+      @facility.update_registered_vehicles(cruz)
```
juliet-e, line 79: This test could be more robust if it tries to register a vehicle and verifies that the vehicle has been added to the registered_vehicles array (rather than just calling the helper method that just shovels the vehicle into the array)

spec/facility_spec.rb
```diff
@@ -23,4 +23,165 @@
...
+
+  describe 'administer_written_test' do
+    it 'administer test and update license_data' do
+      registrant_1 = Registrant.new('Bruce', 18, true )
+      @facility.add_service('Written Test')
+
+      @facility.administer_written_test(registrant_1)
+
+      expect(registrant_1.license_data[:written]).to eq(true)
+    end
```
juliet-e, line 93: This is where it would be helpful to think about some sad path/edge cases (we have some with Penny in the interaction pattern) for when a registrant is too young or doesn't have a permit. Making sure to add those tests will help you make sure you're not missing any requirements!

PR #460, opened 2024-04-02
Chee Lee Second Solo Project Submission

cheeleertr:
Chee Lee Second Solo Project Submission


KatBrandt (commented):
Overall you're doing great work! It seems like you're understanding how to interact with objects between classes. Great work.

PR #461, opened 2024-04-02
Zach Bergman - Project Submission

zach-bergman:
The DMV Project - Second Solo Project

### DIFF
--- a/lib/facility_factory.rb
+++ b/lib/facility_factory.rb
@@ -19,6 +19,14 @@ class FacilityFactory
                     phone: facility[:public_phone_number]
                 }
                 Facility.new(facility_details)
+            else facility[:state] == 'MO'
+                facility_details = {
+                    name: facility[:name],
+                    address: format_address(facility[:address1], facility[:address_2], 
+                    facility[:city], facility[:state], facility[:zipcode]),
+                    phone: facility[:phone]
+                }
+                Facility.new(facility_details)
             end
         end
     end
--- a/lib/facility_factory.rb
+++ b/lib/facility_factory.rb
@@ -27,7 +27,8 @@ class FacilityFactory
                     name: facility[:name],
                     address: format_address(facility[:address1], facility[:address_2], 
                     facility[:city], facility[:state], facility[:zipcode]),
-                    phone: facility[:phone]
+                    phone: facility[:phone],
+                    hours: format_mo_hours(facility[:daysopen], facility[:daysclosed], facility[:holidaysclosed])
                 }
                 Facility.new(facility_details)
             end
--- a/lib/facility_factory.rb
+++ b/lib/facility_factory.rb
@@ -5,7 +5,7 @@ class FacilityFactory
             if facility[:state] == 'CO'
                 facility_details = {
                     name: facility[:dmv_office],
-                    address: format_co_address(facility[:address_li], facility[:address__1], 
+                    address: format_address(facility[:address_li], facility[:address__1], 
                     facility[:city], facility[:state], facility[:zip]),
                     phone: facility[:phone]
                 }
@@ -13,8 +13,9 @@ class FacilityFactory
             elsif facility[:state] == 'NY'
                 facility_details = {
                     name: facility[:office_name],
-                    address: format_ny_address(facility[:street_address_line_1], 
-                    facility[:city], facility[:state], facility[:zip_code]),
+                    address: format_address(facility[:street_address_line_1], 
+                    facility[:street_address_line_2], facility[:city], facility[:state], 
+                    facility[:zip_code]),
                     phone: facility[:public_phone_number]
                 }
                 Facility.new(facility_details)
@@ -22,11 +23,7 @@ class FacilityFactory
         end
     end
 
-    def format_co_address(address_1, address_2, city, state, zip)
+    def format_address(address_1, address_2, city, state, zip)
         "#{address_1} #{address_2} #{city} #{state} #{zip}"
-    end
-
-    def format_ny_address(address, city, state, zip)
-        "#{address} #{city} #{state} #{zip}"
     end
 end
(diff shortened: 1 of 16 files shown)


lib/facility_factory.rb
```diff
@@ -0,0 +1,49 @@
...
+                    name: facility[:name],
+                    address: format_address(facility[:address1], facility[:address_2], 
+                    facility[:city], facility[:state], facility[:zipcode]),
+                    phone: facility[:phone],
+                    hours: format_mo_hours(facility[:daysopen], facility[:daysclosed], facility[:holidaysclosed])
+                }
+                Facility.new(facility_details)
+            end
+        end
+    end
```
juliet-e, line 36: This works, and you're accessing nested data well, but it is a very big method. Can you make some helper methods for each state so that this if statement just calls different methods based on the value of `[:state]`?

lib/most_popular_finder.rb
```diff
@@ -0,0 +1,31 @@
...
+    def initialize
+        @factory = VehicleFactory.new
+    end
+
+    def most_popular_vehicle_registered(vehicle_registrations)
+        vehicles = @factory.create_vehicles(vehicle_registrations)
+
+        make_model = vehicles.map {|vehicle| "#{vehicle.make} #{vehicle.model}"}
+
+        make_model.group_by {|string| string}.max_by {|k,v| v.size}.first
```
juliet-e, line 13: Awesome use of group_by to count occurrences!

lib/most_popular_finder.rb
```diff
@@ -0,0 +1,31 @@
+class MostPopularFinder
```
juliet-e, line 1: Extracting this into a separate class is a great call. We call this a "utility" class because it's not necessarily representing a domain object like a vehicle or facility, but it's performing some utility for us. This name is fine IMO.

lib/vehicle.rb
```diff
@@ -22,4 +28,18 @@ def antique?
   def electric_vehicle?
     @engine == :ev
   end
+
+  def set_plate_type
```
juliet-e, line 32: Nice work using a setter method here rather than giving other classes permission to mutate your :plate_type attribute. This is a good example of encapsulation in object oriented programming

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,33 @@
...
+                    vin: vehicle[:vin],
+                    year: vehicle[:model_year].to_i,
+                    make: vehicle[:make],
+                    model: vehicle[:body_type],
+                    engine: vehicle[:fuel_type],
+                    registration_date: vehicle[:reg_valid_date],
+                    plate_type: vehicle[:record_type],
+                    county: vehicle[:county]
+                }
+                Vehicle.new(vehicle_details)
```
juliet-e, line 29: Similar to your other Factory, this looks great but could be broken up into 2 methods (one for NY one for WA)

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,191 @@
...
+                expect(first_facility.address).to be_a(String)
+                expect(first_facility.address).to eq('409 E AUSTIN BLVD  NEVADA MO 64772')
+            end
+
+            it 'creates proper phone number for the MO DMV Facility instances' do
+                first_facility = @factory.create_facilities(@missouri_facilities).first
+
+                expect(first_facility.phone).not_to eq(nil)
+                expect(first_facility.phone).to be_a(String)
+                expect(first_facility.phone).to eq('(417) 448-5705')
```
juliet-e, line 149: These tests are great -- very robust and organized. Because we're calling an API though, there is the possible risk that this data could change. For the purposes of this project, this is not a problem, but typically when we are handling external data, we won't want to make assertions about the actual values of the first facility in case those changes. It is sufficient to just test that the attribute is the correct datatype

spec/facility_spec.rb
```diff
@@ -2,25 +2,265 @@
...
+      expect(@bolt.registration_date).to eq(nil)
+      @facility_1.register_vehicle(@bolt)
+      expect(@bolt.registration_date).to be_an_instance_of(Date)
+
+      expect(@cruz.registration_date).to eq(nil)
+      @facility_1.register_vehicle(@cruz)
+      expect(@cruz.registration_date).to be_an_instance_of(Date)
+    end
+
+    describe 'SadPath - Only facilities offering Vehicle Registration can perform service' do
```
juliet-e, line 100: Great attention to edge cases in your testing!

spec/most_popular_finder_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+        it 'returns count of most popular model year registered' do
+            expect(@most_popular_finder.most_popular_model_year_registered(@wa_ev_registrations)).to be_a String
+            expect(@most_popular_finder.most_popular_model_year_registered(@wa_ev_registrations)).to eq('2023 with 131 registrations')
+        end
+    end
+
+    describe '#most_popular_county_registered' do
+        it 'returns most popular county for WA EV registrations' do
+            expect(@most_popular_finder.most_popular_county_registered(@wa_ev_registrations)).to be_a String
+            expect(@most_popular_finder.most_popular_county_registered(@wa_ev_registrations)).to eq('King county with 774 registrations')
```
juliet-e, line 40: Great testing here. This one is tricky because we are handling external data, but we want to make sure the return values from those stats methods are correct. So in this case, just checking the datatype is not really sufficient. Using the actual value you expect returned here is the best way to go. But, in the future, you'll learn about how you can use some test data (i.e. a file full of an example response from the wa_ev_registrations data) and read that in rather than grabbing the external data. That way, you have total control over the data and know it won't change

lib/facility_factory.rb
```diff
@@ -0,0 +1,49 @@
+class FacilityFactory
+
+    def create_facilities(facility_data)
+        facility_data.map do |facility|
+            if facility[:state] == 'CO'
+                facility_details = {
+                    name: facility[:dmv_office],
+                    address: format_address(facility[:address_li], facility[:address__1], 
+                    facility[:city], facility[:state], facility[:zip]),
```
juliet-e, line 9: Just missing one thing -- services! We want to read in the services from the external data so that the facility can perform the appropriate tasks

PR #462, opened 2024-04-02
DMV Turn-In

AustinKCodes:


lib/dmv.rb
```diff
@@ -9,7 +11,7 @@ def add_facility(facility)
   end
 
   def facilities_offering_service(service)
-    @facilities.find do |facility|
+    @facilities.select do |facility|
```
KatBrandt, line 14: Yup, good use of select.

lib/facility.rb
```diff
@@ -1,14 +1,49 @@
+require "./lib/vehicle"
 class Facility
-  attr_reader :name, :address, :phone, :services
+  attr_reader :name, :address, :phone, :services, :registered_vehicles, :collected_fees
```
KatBrandt, line 3: Can list these vertically once you get past 3 attr_readers

lib/facility.rb
```diff
@@ -1,14 +1,49 @@
...
+  def add_service(service)
     @services << service
   end
-end
+
+  def register_vehicle(vehicle)
+    @registered_vehicles << vehicle
+    @collected_fees += registration_fee(vehicle)
+    @registration_date = Date.today
+
```
KatBrandt, line 22: This method is on the right track. I like that you created the `registration_fee` helper method to move out some of the logic.
One thing we needed to do before running lines 19-21 though, is make sure that the facility.services included the service of "Vehicle Registration", so before line 19 we'd want something like `if @services.include?("Vehicle Registration"` then run the logic.
We also want to think about where we want the registration date to be set. It's the vehicle object that has the attribute of registration date, so line 21 we could write like `vehicle.registration_date = Date.today`

lib/facility.rb
```diff
@@ -1,14 +1,49 @@
...
+  def set_plate
+      if vehicle.antique?
+        vehicle.set_plate(:antique)
+      elsif vehicle.electric_vehicle?
+        vehicle.set_plate(:ev)
+      else
+        vehicle.set_plate(:regular)
+      end
+    end
```
KatBrandt, line 34: This helper method looks really great! I like that you made a helper method in the Vehicle class called `set_plate`

lib/facility.rb
```diff
@@ -1,14 +1,49 @@
...
+      @collected_fees += 25
+      # vehicle.set_plate(:antique)
+    elsif vehicle.electric_vehicle?
+      @collected_fees += 200
+      # vehicle.set_plate(:ev)
+    else 
+      @collected_fees += 100
+      # vehicle.set_plate(:regular)
+    end
+  end
```
KatBrandt, line 47: This method looks great! Since you created that `set_plate` helper method, you don't need any of that commented out code in this method.

lib/registrant.rb
```diff
@@ -0,0 +1,25 @@
+class Registrant
+    attr_reader :name, :age, :permit, :license_data
+    def initialize(name, age, permit = nil)
+        @name = name
+        @age = age
+        @permit = permit unless permit.nil? do
```
KatBrandt, line 6: To set the default value of the permit to false, we'd want line 3 to read `def initialize(name, age, permit = false)` then line 6 would just be `@permit = permit` since we already set it to false on line 3.

lib/registrant.rb
```diff
@@ -0,0 +1,25 @@
...
+        @license_data = {:written => false, :license => false, :renewed => false}
+    end
+
+    def permit?
+        if permit == nil
+        return false
+        else 
+            permit == true
+        return true
+        end
```
KatBrandt, line 17: This does work logic-wise. 
However, if we already have @permit = false in the initialize method, then the inside of this method could just be `@permit` so it acts like an attr_reader.

spec/vehicle_registration_spec.rb
```diff
@@ -0,0 +1,17 @@
+require './lib/facility'
+require './lib/vehicle'
+require "./lib/vehicle_registration"
+
+RSpec.describe VehicleRegistration do
```
KatBrandt, line 5: Remember we want the test file to match the class file, so this file should be called `vehicle_spec.rb`

PR #463, opened 2024-04-02
Austin Carr-Jones

austincarrjones:


lib/facility.rb
```diff
@@ -1,14 +1,62 @@
...
+    @collected_fees = 0
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    @registered_vehicles << vehicle
```
juliet-e, line 24: We only want to register a vehicle if it's an included service, so you'll want to wrap this logic into an if statement checking that it's available. 

Nice use of a helper method in the Vehicle class!
You might want to consider a helper method for collecting the fees as well. Because this registration method can get pretty complex, you could create a helper method in this class that collects the fees based on the vehicle type. 

Finally -- be mindful of return values in the interaction pattern - we want this method to either return the @registered_vehicles attribute (if successful) or return false if vehicle registration is not a provided service

lib/facility_factory.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def create_facilities(office_locations)
+    office_locations.map do |office_location|
+      office_location[:name] = office_location[:dmv_office]
+      office_location[:address] = office_location[:address_li]
+      Facility.new(office_location)
+    end
+  end
+
+  # def concatenate_address
```
juliet-e, line 11: Something tricky about this factory -- the facility data that were reading in has a list of services the facility provides! So, instead of starting `services` as an empty array, you could refactor your Facility initialize to pull in the existing services from the data

lib/registrant.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def license_data?
+    @license_data
+  end
+
+  def earn_permit
+    @permit = true
+  end
+
+  def written_test_complete
```
juliet-e, line 31: I really love this and the following methods that set the value of license_data keys. This is a great example of encapsulation. We *could* use an `attr_accessor` for the license_data attribute and let other classes set the values for license_data (and if you did this it would be fine), but we often talk about how attr_accessors are not always the best choice because they give the code outside of this class a lot of control to set the values. What if they set `license_data[:banana] = true`? What's stopping that from happening! Instead, you're keeping license_data to read-only permission, and controlling the value of these hash keys with these methods, which we call `setter` methods (because they set the value).

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,12 @@
+class VehicleFactory
+
+  def create_vehicles(registrations)
+    registrations.map do |registration|
```
juliet-e, line 4: Nice handling of external data!

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,64 @@
+require 'spec_helper'
+
+RSpec.configure do |config|
+config.formatter = :documentation
+end
+
+RSpec.describe FacilityFactory do
```
juliet-e, line 7: Great adherence to TDD! :)

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      :zip=>"80205",
+      :phone=>"(720) 865-4600",
+      :hours=>"Mon, Tue, Thur, Fri  8:00 a.m.- 4:30 p.m. / Wed 8:30 a.m.-4:30 p.m.",
+      :services_p=>"vehicle titles, registration, renewals;  VIN inspections",
+      :parking_no=>"parking available in the lot at the back of the bldg (Glenarm Street)",
+      :photo=>"images/Tremont.jpg",
+      :address_id=>"175164",
+      :":@computed_region_nku6_53ud"=>"1444"
+      }]
+
```
juliet-e, line 34: There are a couple different ways to handle these tests. 
1) what you're doing here is defining the test data in this office_locations array. You are defining the test data and the benefit of this is that you have total control over the data and you know it won't change. You only have one facility object though here, so you'd maybe want to add another to ensure that your code can create more than one facility. That can get long so sometimes we pull this test data into other files to read from (more on that later)

2) Another option is to call the DmvDataService.new.co_dmv_office_locations and use that instead of office_locations as the result that you're going to call assertions on. The benefit is that you have a very realistic data set (i.e. many facilities listed in an array), but the downside is that this data could change so you wouldn't always want to make assertions on the actual values. Instead, you might want to verify the structure (that certain keys exist) and check the data types of the values rather than the actual values. If you prefer being able to test the actual values, then going with option 1 would be best. But, generally testing that office attributes are not nil and the right datatype can be sufficient. 

Which option you choose will definitely depend on lots of factors, but thought I would share this context since it seems you were on the track to using one and then switched to the other!

spec/facility_spec.rb
```diff
@@ -1,26 +1,238 @@
...
+      @facility_1.add_service('Road Test')
+      @facility_1.add_service('Renew License')
+      @registrant_2.earn_permit
+      @facility_1.administer_written_test(@registrant_2)
+      @facility_1.administer_road_test(@registrant_2)
+
+      expect(@facility_1.renew_drivers_license(@registrant_2)).to be true
+      expect(@registrant_2.license_data).to eq({:written=>true, :license=>true, :renewed=>true})
+    end
+  end
```
juliet-e, line 236: Great work writing robust and well organized specs!

spec/registrant_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+
+RSpec.describe Registrant do
+
+  describe '#initialize' do
+    it 'can initialize' do
+      registrant_1 = Registrant.new('Bruce', 18, true)
+      registrant_2 = Registrant.new('Penny', 16 )
+      registrant_3 = Registrant.new('Tucker', 15 )
+
+      expect(registrant_1).to be_instance_of(Registrant)
```
juliet-e, line 15: just testing this for one Registrant object is sufficient IMO -- just populating the attributes in the same way (i.e. there's no sad path or edge case to consider) doesn't need to be tested with multiple objects

austincarrjones: Ahh I see, thanks!

PR #464, opened 2024-04-02
Cory bretsch/the dmv

CoryBretsch:
project 2 submission


lib/facility.rb
```diff
@@ -1,14 +1,107 @@
...
+  def add_service(service)
     @services << service
   end
+
+
+  def register_vehicle(automobile)
+    if @services.include?('Vehicle Registration')
+    automobile.registration_date = Date.today
+    @registered_vehicles << automobile
+      if automobile.antique? == true
```
juliet-e, line 28: setting the plate type and collecting fees here can add a good amount of complexity, making this method pretty big. Consider creating one or two helper methods that you can call here to handle the plate type and fee collection. For example, your method could become:
```
 def register_vehicle(automobile)
        automobile.registration_date = Date.today
        @registered_vehicles << automobile
        collect_fees(automobile)
        automobile.set_plate_type
        return @registered_vehicles
    else
        return nil
   end
end
```

Then, `collect_fees` could be another method in this class that determines the fees to add to @collected_fees based on the car type. `set_plate_type` could be a method in the Vehicle class to change the plate_type since thats more of a vehicle responsibility than something the facility needs to worry about.

lib/facility.rb
```diff
@@ -1,14 +1,107 @@
...
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+
+  def register_vehicle(automobile)
+    if @services.include?('Vehicle Registration')
+    automobile.registration_date = Date.today
```
juliet-e, line 26: Remember to indent after an if statement!

lib/facility.rb
```diff
@@ -1,14 +1,107 @@
...
+      if registrant.license_data[:license] == true
+        registrant.license_data[:renewed] = true
+        return true
+      else
+        return false
+      end
+    else
+      return false
+    end
+  end
```
juliet-e, line 80: Great structure with the rest of these methods!

lib/registrant.rb
```diff
@@ -0,0 +1,30 @@
...
+        @name = name
+        @age = age
+        @permit = permit
+        @license_data = {written: false, 
+                        license: false, 
+                        renewed: false}
+    end
+
+    def permit?
+        @permit = @permit
```
juliet-e, line 17: This method is a boolean method, so we just need it to return true or false depending on the value of @permit. @permit is a boolean attribute, so that means that this `permit?` method can just contain `@permit` since that will evaluate to either true or false! We don't need to assign it to anything in this method

lib/registrant.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    def earn_permit
+        if @permit == false
+            return @permit = true
+        else
+            @permit
+        end
+
+    end
+end
```
juliet-e, line 28: Your indentation looks really large in this file. When you're working in VSCode, the bottom right of the screen shows how many spaces are being added for an indent. Typically, you want 2 spaces

spec/facility_spec.rb
```diff
@@ -23,4 +37,97 @@
...
+      expect(@bolt.plate_type).to eq(:ev)
+      expect(@facility_1.registered_vehicles).to eq([@cruz, @camaro, @bolt])
+      expect(@facility_1.collected_fees).to eq(325)
+      expect(@facility_2.registered_vehicles).to eq([])
+      expect(@facility_2.services).to eq([])
+
+      @facility_2.register_vehicle(@bolt)
+
+      expect(@facility_2.registered_vehicles).to eq([])
+      expect(@facility_2.collected_fees).to eq(0)
```
juliet-e, line 75: This test is a bit long, so I think you could break out the last piece (where you're registering a vehicle at a facility that doesn't have that service) into a separate test (i.e. `it "will not register a vehicle if it's not a provided service"`)

spec/facility_spec.rb
```diff
@@ -23,4 +37,97 @@
...
+      @facility_1.add_service('Written Test')
+
+      expect(@facility_1.services).to eq(['Written Test'])
+      expect(@facility_1.administer_written_test(@registrant_1)).to be(true)
+      expect(@registrant_1.license_data).to eq({:written=>true, :license=>false, :renewed=>false})
+      expect(@registrant_2.age).to eq(16)
+      expect(@registrant_2.permit?).to be(false)
+      expect(@facility_1.administer_written_test(@registrant_2)).to be(false)
+
+      @registrant_2.earn_permit
```
juliet-e, line 98: These chunks could also be broken into separate tests -- one for taking the written test if it's possible, one for NOT taking the written test if you're too young, one for NOT taking the test if you don't have a permit, etc. Then another one for earning a permit based on requirements. Another one for taking the road test. So for each service/functionality, you write a test for the happy path, and then as needed, write some tests for the sad paths (when the registrant or facility is not capable of doing that service)

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    describe '#initialize' do
+        it 'can initialize' do
+        expect(@factory).to be_an_instance_of(VehicleFactory)
+        end
+    end
+
+    describe '#create_vehicles' do
+        it "can create an array of vehicles objects" do
+            wa_ev_registrations = DmvDataService.new.wa_ev_registrations
```
juliet-e, line 20: You can remove this `wa_ev_registrations` and just use the test data you're defining on the next line if you'd like! This is a great example of using fixed test data (even if it originally came from an external source) in order to have control over what the data coming in will be.

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+                            :_2019_hb_2042_sale_price_value_requirement=>"The sale price is too high",
+                            :electric_vehicle_fee_paid=>"Not Applicable",
+                            :transportation_electrification_fee_paid=>"Not Applicable",
+                            :hybrid_vehicle_electrification_fee_paid=>"Not Applicable",
+                            :census_tract_2020=>"53033024001",
+                            :legislative_district=>"48",
+                            :electric_utility=>"PUGET SOUND ENERGY INC||CITY OF TACOMA - (WA)"}]
+            cars = @factory.create_vehicles(registration)
+
+            expect(cars[0]).to be_an_instance_of(Vehicle)
```
juliet-e, line 52: Nice assertions here!

PR #465, opened 2024-04-02
Lito Croy

litobot:
Better luck next time...


lib/facility.rb
```diff
@@ -1,14 +1,97 @@
...
+        @collected_fees += 25 # Add $25 to @collected_fees
+        vehicle.set_plate(:antique)
+      elsif vehicle.electric_vehicle? # Elsif vehicle = EV
+        @collected_fees += 200    # Add $200 to @collected_fees
+        vehicle.set_plate(:ev)
+      else
+        @collected_fees += 100 # Else add $100 to @collected_fees
+        vehicle.set_plate(:regular)
+      end
+    end
```
KatBrandt, line 36: This method looks good! I like that you wrote a method in the Vehicle class called `set_date` to handle setting the date. 
The other thing we could do to refactor this, is take the if block from lines 26-35 and put that in it's own "helper method", maybe it could be called `set_fees_and_plate` and then the `register_vehicle` method would just call that method on line 26, but all the code/logic would be inside that helper method. That's how we would go about refactoring this.

lib/facility.rb
```diff
@@ -1,14 +1,97 @@
...
+  end
+
+
+
+# Administer a written test
+# A written test can only be administered to registrants with a permit and who are at least 16 years of age
+
+  def administer_written_test(registrant)
+    if services.include?("Written Test")
+      if registrant.permit? && registrant.age >= 16
```
KatBrandt, line 46: Yes, good job with thinking through all the logic here. Good use of your boolean operator: `&&`. You can also get rid of the nested if statement by making line 45 read: `if services.include?("Written Test") && registrant.permit? && registrant.age >=16`

lib/facility.rb
```diff
@@ -1,14 +1,97 @@
...
+        false
+      end
+    else
+      false
+    end
+  end
+
+  def administer_road_test(registrant)
+    if services.include?("Road Test") 
+      registrant.permit? && registrant.age >= 16
```
KatBrandt, line 58: this line is correct, but it's not connected to anything. Would need another `if` in front of it, or a `&&` to tie it to the line above.

lib/registrant.rb
```diff
@@ -0,0 +1,42 @@
...
+              :license_data
+
+  def initialize(name, age, permit = false)
+    @name = name
+    @age = age
+    @permit = permit
+    @license_data = {:written => false, :license => false, :renewed => false}
+  end
+
+  def permit?
```
KatBrandt, line 13: Great job with this one and great job NOT putting `:permit` in the attr_reader which many people did. We don't want it in the attr_reader since we have it here.

lib/vehicle.rb
```diff
@@ -22,4 +28,13 @@ def antique?
   def electric_vehicle?
     @engine == :ev
   end
+
+  def set_plate(plate)    # This is a helper method - not defined in IP
```
KatBrandt, line 32: Yes, I like that you created these additional helper methods!

spec/facility_spec.rb
```diff
@@ -1,26 +1,241 @@
...
-      @facility.add_service('Renew Drivers License')
-      @facility.add_service('Vehicle Registration')
-      expect(@facility.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
+      expect(@facility_1.services).to eq([])
+
+      @facility_1.add_service('New Drivers License')
+      @facility_1.add_service('Renew Drivers License')
+      @facility_1.add_service('Vehicle Registration')
+
+      expect(@facility_1.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
```
KatBrandt, line 32: This is a good test because it shows the starting state of the .services attribute, then we run a command method and add in services, then you check the final result of the services attribute.

spec/facility_spec.rb
```diff
@@ -1,26 +1,241 @@
...
+
+      @facility_1.add_service('Vehicle Registration')
+      @facility_1.register_vehicle(@cruz)
+      
+      expect(@facility_1.registered_vehicles).to eq([@cruz])
+      
+      @facility_1.register_vehicle(@camaro)
+      @facility_1.register_vehicle(@bolt)
+
+      expect(@facility_1.registered_vehicles).to eq([@cruz, @camaro, @bolt])
```
KatBrandt, line 53: Based on the it statement for this test, I would expect this test to stop here after line 53. Then the tests below might be in a new it block that says `it "does not register cars if service not in facility"`

spec/facility_spec.rb
```diff
@@ -1,26 +1,241 @@
...
+      @facility_1.add_service('Vehicle Registration')
+      @facility_1.register_vehicle(@cruz)
+      @facility_1.register_vehicle(@camaro)
+      @facility_1.register_vehicle(@bolt)
+
+      expect(@cruz.registration_date).to eq(Date.today)
+      expect(@camaro.registration_date).to eq(Date.today)
+      expect(@bolt.registration_date).to eq(Date.today)
+
+    end
```
KatBrandt, line 118: The rest of these tests look really good. Good organization with the it blocks.

spec/vehicle_spec.rb
```diff
@@ -33,4 +34,21 @@
       expect(@camaro.electric_vehicle?).to eq(false)
     end
   end
+
+  describe '#set_date' do
```
KatBrandt, line 38: Yes, good job testing your helper method. You'd also want to test your other `set_plate` helper method.

KatBrandt (commented):
Overall this code looks good! You're on the right track here Lito, hang in there!

PR #468, opened 2024-04-03
Jan Lehigh

JCL461437:
Shamefully incomplete. We'll get it next time!


lib/facility.rb
```diff
@@ -1,14 +1,50 @@
...
-    @address = address
-    @phone = phone
+  def initialize(facility_details)
+    @name = facility_details[:name]
+    @address = facility_details[:address]
+    @phone = facility_details[:phone]
     @services = []
+    @registered_vehicles = []
+    @collected_fees = 0
+    @plate_type = :regular
```
KatBrandt, line 20: This plate_type attribute is not an attribute of the Facility, but of the Vehicle. So we don't want this instance variable here, we'd want it in the Vehicle class.

lib/facility.rb
```diff
@@ -1,14 +1,50 @@
...
+        @registered_vehicles << vehicle
+        vehicle.plate_type
+        vehicle.assign_date
+      else
+        @collected_fees += 100
+        @registered_vehicles << vehicle
+        vehicle.plate_type
+        vehicle.assign_date
+      end
+    end
```
KatBrandt, line 45: Yes, so all the logic here is working great. We could refactor this be noticing that every branch of logic is doing `vehicle.plate_type` and `vehicle.assign_date` which means those two lines could just be above line 29, before your second if statement. 
Great job making helper methods of `plate_type` and `assign_date`!

lib/registrant.rb
```diff
@@ -0,0 +1,22 @@
+
+class Registrant 
+    attr_reader :name, :age, :permit, :license_data
```
KatBrandt, line 3: Careful here, we do not want `:permit` in the attr_readers because we don't want a method called `permit`. We have a method called `permit?` that acts as an attr_reader for us.

lib/vehicle.rb
```diff
@@ -5,21 +5,40 @@ class Vehicle
...
     @engine == :ev
   end
+
+  def plate_type
+    if @engine == :ev
+      @plate_type = :ev
+    elsif antique?
+      @plate_type = :antique
+    else 
+      @plate_type = :regular
```
KatBrandt, line 37: Yes, see, you have the @plate_type attribute here, but it's missing from the initialize so it won't ever get set.

spec/facility_spec.rb
```diff
@@ -1,26 +1,89 @@
...
+    end
+
+    it 'determines a registration date' do
+      @facility_1.add_service('Vehicle Registration')
+      @facility_1.register_vehicle(@cruz)
+    
+      expect(@cruz.registration_date).to eq(Date.today.year)
+    end
+
+  end
```
KatBrandt, line 78: Good job with your test organization here. Great job testing all the side effects of the register_vehicle method.

KatBrandt (commented):
This is a great start, Jan! More practice and more reps and you'll get this.

PR #469, opened 2024-04-03
Lance Nelson

LancePants97:


lib/facility.rb
```diff
@@ -1,14 +1,65 @@
...
+     elsif vehicle.electric_vehicle? == true
+      @collected_fees += 200
+      vehicle.plate_type = :ev
+     else
+      @collected_fees += 100
+      vehicle.plate_type = :regular
+     end
+    else
+    end
+  end
```
KatBrandt, line 33: Overall, this method looks good, it's doing all the things we need this method to do. A few notes:
1. The indentation on this method is all off. Line 21 should not be spaced in. Also, the indentation appears to sometimes be a tab and sometimes a space, we want to be consitent and always use a tab.
2. We could refactor this method by taking lines 21-30 and putting those in their own helper method, then just calling the helper method from line 21. We want to start working on having each method doing only one thing.

lib/facility.rb
```diff
@@ -1,14 +1,65 @@
...
+
+  def administer_written_test(registrant)
+    if @services.include?('Written test') &&
+      registrant.age >= 16 && registrant.permit? == true
```
KatBrandt, line 37: Good use of boolean operators in your if statement.

spec/facility_spec.rb
```diff
@@ -2,25 +2,254 @@
...
+
+    it 'can add multiple available services' do
+      expect(@facility_1.services).to eq([])
+      
+      @facility_1.add_service('New Drivers License')
+      @facility_1.add_service('Renew Drivers License')
+      @facility_1.add_service('Vehicle Registration')
+      
+      expect(@facility_1.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
+    end
```
KatBrandt, line 44: This test and the one above it are redundant, I would just have the it block that starts on line 36, and remove lines 28-34.

spec/facility_spec.rb
```diff
@@ -2,25 +2,254 @@
...
+
+    it 'can register an electric vehicle, collect fees, and assign a license plate' do
+      @bolt = Vehicle.new({vin: '987654321abcdefgh', year: 2019, make: 'Chevrolet', model: 'Bolt', engine: :ev} )
+      
+      @facility_1.add_service('Vehicle Registration')
+      @facility_1.register_vehicle(@bolt)
+
+      expect(@facility_1.collected_fees).to eq (200)  
+      expect(@bolt.plate_type).to be (:ev)
+    end
```
KatBrandt, line 76: Yes, I like how you separated out the different vehicle types into different it blocks. We also need to test the case where the facility does not have the "Vehicle Registration" service, and make sure that none of these side effects happen.

spec/facility_spec.rb
```diff
@@ -2,25 +2,254 @@
...
+    it 'can deny a written test even if registrant is of age' do
+      @registrant_3 = Registrant.new('Tucker', 16)
+
+      @facility_1.add_service('Written test')
+      @facility_1.administer_written_test(@registrant_3)
+
+      expect(@registrant_3.license_data[:written]).to be false
+      expect(@registrant_3.license_data[:license]).to be false
+      expect(@registrant_3.license_data[:renewed]).to be false
+    end
```
KatBrandt, line 133: For these it blocks, we want to be even more descriptive of what we're testing in each it block. This method has three conditions that must be met in order for the method to get into the if statement, so we want each it block to reflect that logic so something like `it "checks for service"` and `it "checks registrant's age"`, etc. Then we want to test all 3 conditions including if none of the conditions are met, and test the return values and side effects. You're on the right track here!

lib/registrant.rb
```diff
@@ -0,0 +1,21 @@
+class Registrant
+    attr_accessor :name, :age, :permit, :license_data
```
KatBrandt, line 2: Careful here, we do not want `:permit` in the attr_reader since we never wanted to create a method called `permit`

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,25 @@
...
+            make = vehicle_data[:make]
+            model = vehicle_data[:model]
+            engine = vehicle_data[:electric_vehicle_type]
+            year = vehicle_data[:model_year]
+            
+            vehicle = Vehicle.new({:vin => vin, :year => year, :make => make, :model => model, :engine => engine})
+            @factory << vehicle
+        end          
+        return @factory
+    end
```
KatBrandt, line 23: Yeah, overall, this method looks really good. I like that you stored all the new Vehicle instances in an instance variable.

lib/vehicle.rb
```diff
@@ -5,14 +5,20 @@ class Vehicle
               :year,
               :make,
               :model,
-              :engine
+              :engine,
+              :registration_date
+
+  attr_accessor :plate_type
```
KatBrandt, line 11: The attr_accessor is technically a method, so you'll want to make sure you write tests for this method to make sure you can read and update the attribute.

KatBrandt (commented):
Overall, really great work here!

PR #475, opened 2024-05-21
Cory Bretsch - 2405 - the_dmv Project Submission

CoryBretsch:


lib/registrant.rb
```diff
@@ -0,0 +1,18 @@
...
+  end
+
+  def permit?
+    @permit
+  end
+
+  def earn_permit
+    @permit = true
+  end
+end
```
KatBrandt, line 18: This class looks great, nice and clean

spec/registrant_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  it "has license data hash with default values false" do
+    expect(@registrant_2.license_data).to eq ({written: false, license: false, renewed: false})
+  end
+
+  it "can earn a permit" do
+    @registrant_2.earn_permit
+
+    expect(@registrant_2.permit?).to be true
+  end
+
```
KatBrandt, line 32: Great job with these tests, very well organized.

lib/facility.rb
```diff
@@ -1,14 +1,72 @@
...
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    if @services.include?("Vehicle Registration")
+      assign_registration_date(vehicle)
+      assign_plate_type(vehicle)
+      charge_fee(vehicle)
+      @registered_vehicles << vehicle
+    end
```
KatBrandt, line 28: Nice job using a lot of helper methods to achieve all the functions of this method. Makes it very readable.

lib/facility.rb
```diff
@@ -1,14 +1,72 @@
...
+  end
+
+  def assign_plate_type(vehicle)
+    if vehicle.engine == :ev
+      vehicle.plate_type = :ev
+    elsif vehicle.antique? == false
+      vehicle.plate_type = :regular
+    else
+      vehicle.plate_type = :antique
+    end
```
KatBrandt, line 48: I like that you made this helper method, however, to improve this even more, it would be better to have this method in the Vehicle class. So from the `register_vehicle` method, you'd call `vehicle.assign_plate_type` since this is an attribute of the vehicle class, we want that class to be in charge of setting it's own attribute. Otherwise, this logic looks great.

lib/facility.rb
```diff
@@ -1,14 +1,72 @@
...
+  end
+
+  def assign_registration_date(vehicle)
+    vehicle.registration_date = Date.today
+  end
+
+  def administer_written_test(registrant)
+    if @services.include?("Written Test") && registrant.age >= 16 && registrant.permit? == true
```
KatBrandt, line 56: Nice job with the boolean logical operators. One nitpick, if a method already returns a boolean value, then you don't need to also compare that result with a boolean, for example, the end of this line could just be `registrant.permit?` since that will result in a `true` or `false` value already, no need to compare it to true.

spec/facility_spec.rb
```diff
@@ -2,25 +2,220 @@
...
+      @facility_1.add_service("Vehicle Registration")
+      @facility_1.register_vehicle(@cruz)
+      @facility_1.register_vehicle(@camaro)
+      @facility_1.register_vehicle(@bolt)
+
+      expect(@facility_1.registered_vehicles).to eq [@cruz, @camaro, @bolt]
+      expect(@facility_1.collected_fees).to be 325
+    end
+
+    it "will not register vehicle if service is not available at facility" do
```
KatBrandt, line 82: great job testing for this case

spec/facility_spec.rb
```diff
@@ -2,25 +2,220 @@
...
+      expect(@registrant_3.permit?).to be false
+
+      @facility_1.administer_written_test(@registrant_3)
+      
+      expect(@registrant_3.license_data).to eq ({written: false, license: false, renewed: false})
+
+      @registrant_3.earn_permit
+      @facility_1.administer_written_test(@registrant_3)
+
+      expect(@registrant_3.license_data).to eq ({written: false, license: false, renewed: false})
```
KatBrandt, line 141: Overall, great job testing for all the separate conditions that are going on here. Good test organization.

lib/facility_factory.rb
```diff
@@ -0,0 +1,50 @@
...
+                  city + " " +
+                  state + " " +
+                  zip
+      end
+      @facilities << Facility.new({name: name,
+                                  address: address, 
+                                  phone: phone,
+                                  services: services})
+    end
+  end
```
KatBrandt, line 48: I like that you accounted for all these scenarios where address lines are present or not and you've also made it so this one method can take in any of the state's facilities. I do however, think this method is doing too much and could benefit from being refactored. Pulling out pieces of logic into helper methods will help this be more readable and easier to navigate and debug.

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+                                :fax=>"(314) 733-5319",
+                                :size=>"3",
+                                :email=>"FERGUSONAGENTOFFICE@DOR.MO.GOV",
+                                :latlng=>{:latitude=>"38.737935", :longitude=>"-90.220029"},
+                                :textingphonenumber=>"(314) 730-0606",
+                                :additional_license_office_info=>"Personalized plates can be picked up at the North County License office located at 11700 W Florissant, St. Louis, MO 63033.",
+                                :":@computed_region_ny2h_ckbz"=>"203",
+                                :":@computed_region_c8ar_jsdj"=>"51",
+                                :":@computed_region_ikxf_gfzr"=>"2210"}]
+  end
```
KatBrandt, line 141: A common practice is to put this stubbed data into another file in your specs though, so it's not creating clutter in your specs. Usually a fixtures directory will hold this stubbed info, then you can read in the file from your spec.

KatBrandt (commented):
Overall, this is fantastic Cory! Your code is very clean and well organized (especially the tests). Please review my feedback and let me know if you have any questions. Great work! 
Your git workflow looks mostly good, but I didn't see any branches. We were wanting to see at least 3 branches for this work. Your commit messages look good, it's up to you if you like using the feat & test tags for commit messages. Some people like them, some don't, up to you. Definitely need to see branching though.

PR #478, opened 2024-05-21
Iteration 3 Complete Full Project Submission

DAVISEVAN:
This is my submission for project 2 the_dmv


lib/dmv.rb
```diff
@@ -1,15 +1,15 @@
...
   end
 
   def add_facility(facility)
     @facilities << facility
   end
-
+# changed .find to .select to find all facilities matching service offers
   def facilities_offering_service(service)
-    @facilities.find do |facility|
+    @facilities.select do |facility|
```
juliet-e, line 12: Great choice!

lib/facility.rb
```diff
@@ -1,14 +1,94 @@
...
-    @address = address
-    @phone = phone
-    @services = []
+# changed to accept single argument as hash named facility_info
+# added .fetch to use for default service value
+  def initialize(facility_info)
+    @name = facility_info[:name]
+    @address = facility_info[:address]
+    @phone = facility_info[:phone]
+    @services = facility_info.fetch(:services, [])
```
juliet-e, line 12: Awesome use of the `fetch` method for this context!! Well done

lib/facility.rb
```diff
@@ -1,14 +1,94 @@
...
-
-  def add_services(service)
+# add_services changed to add_service 
+  def add_service(service)
     @services << service
   end
+  # administer_written_test method to update license_data
+  def administer_written_test(registrant)
+    if @services.include?('Written Test') 
+    registrant.license_data[:written] = true
```
juliet-e, line 23: Just be mindful of indenting your if statement properly for readability. Line 23, 24 and 26 should be indented

lib/facility.rb
```diff
@@ -1,14 +1,94 @@
...
+      else
+        vehicle.plate_type = :regular
+        @collected_fees += 100
+      end
+      
+      @registered_vehicles << vehicle
+    end
+  end
+  #Method create_from_data to create facilities from dmv_data_service
+  def self.create_from_data(data)
```
juliet-e, line 68: Nice work using a class method here. For Factory methods, you're correct that the two conventional places where we would see these types of utility methods that take external data and create objects would be specific Factory classes (like you have for Vehicle) OR class methods. Since we haven't gone over class methods yet, we mostly see FacilityFactory classes for this functionality but class methods are a great option! Awesome

lib/facility.rb
```diff
@@ -1,14 +1,94 @@
...
+      registrant.license_data[:renewed] = true
+      return true
+    else
+      return false
+    end
+  end
+  # register_vehicle method to register vehicles
+  # determines fee collection based on plat_type
+  def register_vehicle(vehicle)
+    if @services.include?('Vehicle Registration')
```
juliet-e, line 50: In order to avoid a nested if statement, you can also consider a guard clause, which will return prematurely if "Vehcle Registration" is not one of the facility services:
`return nil unless @services.include?("Vehicle Registration")`

lib/facility.rb
```diff
@@ -1,14 +1,94 @@
...
+      vehicle.registration_date = Date.today
+  
+      if vehicle.engine == :ev
```
juliet-e, line 53: Refactor Opportunity (for the future, we know time is limited!!): Consider, when you have a long method like this, how you can use helper methods to break up the logic. For example, you could have a helper method here that you call `process_antique_registration` and it can handle setting the plate type, setting registration date, collecting the right amount of fees for this type of car, etc.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,15 @@
+require_relative 'vehicle'
+
+class VehicleFactory
+  def create_vehicles(vehicle_data_array)
+    vehicle_data_array.map do |vehicle_data|
```
juliet-e, line 5: Great use of map!

spec/create_facility_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    it 'creates facilities from Missouri data' do
+      mo_data = dmv_data_service.mo_dmv_office_locations
+      facilities = Facility.create_from_data(mo_data)
+
+      expect(facilities).not_to be_empty
+      facilities.each do |facility|
+        expect(facility).to be_an_instance_of(Facility)
+        expect(facility.name).not_to be_nil
+        expect(facility.address).not_to be_nil
```
juliet-e, line 44: Excellent testing! By testing for the presence of values but not making exact equivalence assertions here, you're building resilient tests even if the data from the external source changes! Awesome job

lib/facility.rb
```diff
@@ -1,14 +1,94 @@
...
+  #Method create_from_data to create facilities from dmv_data_service
+  def self.create_from_data(data)
+    # Iterates over each element in the data array
+    data.map do |facility_record|
+      # Extracts relevant information from the facility_record and create a hash
+      facility_info = {
+        # Uses the value from the first available key for the facility name
+        name: facility_record[:dmv_office] || facility_record[:office_name] || facility_record[:name],
+        
+        # Constructs the address by combining multiple fields, ensuring to handle different possible key names
```
juliet-e, line 76: Great flexibility in your implementation. I think the only thing missing here is the services data (and your Facility is flexible to grab the services key from the hash or default to an  [])

spec/facility_spec.rb
```diff
@@ -12,15 +17,77 @@
...
+      end
+    end
+
+    context 'when the service is not available' do
+      it 'does not register the vehicle' do
+        expect(@facility_2.register_vehicle(@bolt)).to be_nil
+        expect(@facility_2.registered_vehicles).to be_empty
+        expect(@facility_2.collected_fees).to eq(0)
+      end
+    end
```
juliet-e, line 88: Excellent structure of different registration scenarios! This makes for a very readable collection of tests!

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    it 'returns an array' do
+      factory = VehicleFactory.new
+      vehicles = factory.create_vehicles([])
+
+      expect(vehicles).to be_an_instance_of(Array)
+    end
+
+    it 'returns an empty array when given an empty data set' do
+      factory = VehicleFactory.new
+      vehicles = factory.create_vehicles([])
```
juliet-e, line 14: Make sure you also add a test for when you send in a dataset with real data. It doesn't have to be the full dataset, but could be an array of 2 sample hashes, and you could just verify that 2 Vehicle objects get created.

PR #479, opened 2024-05-21
Sid Swaminathan

sidswaminathan1989:
DMV project for Sid Swaminathan


Gemfile
```diff
@@ -3,4 +3,4 @@ source 'https://rubygems.org'
 gem 'rspec'
 gem 'faraday'
 gem 'pry'
-
+gem 'pry-reload'
```
juliet-e, line 6: This will be more relevant in Mod2 when you're dealing with databases, but pry on its own has a `reload!` command that you can use without this additional gem!

lib/dmv.rb
```diff
@@ -9,8 +11,31 @@ def add_facility(facility)
   end
 
   def facilities_offering_service(service)
+    
+    @matching_facilities = []
+    
+    # Iterate over each Facility to determine
+    # whether or not the passed "service" is
+    # included in the list of "facility.services"
     @facilities.find do |facility|
```
juliet-e, line 20: you could simplify this solution by using `find_all` instead of `find`

juliet-e: This would remove the need to have an accumulator array

lib/facility.rb
```diff
@@ -1,14 +1,85 @@
...
+    @name = facility_obj[:name]
+    @address = facility_obj[:address]
+    @phone = facility_obj[:phone]
     @services = []
+    @registered_vehicles = []
   end
 
-  def add_services(service)
+  def add_service(service)
+    @collected_fees = 0
```
juliet-e, line 31: I'd suggest defaulting the @collected_fees in the initialize rather than here. Every time you add a service, this instance var would get re-assigned back to 0 which is not expected behavior. If you default it in the initialize, however, it can accurately store all the collected fees for any service!

lib/facility.rb
```diff
@@ -1,14 +1,85 @@
...
+# require './lib/registrant'
+
 class Facility
-  attr_reader :name, :address, :phone, :services
+  
+  # Only Enables Writing (Setting New Values) of Properties, Not Reading
+  # attr_writer :name, :address, :phone, :services
+
+  # Enables Both Reading / Writing Properties
+  # attr_accessor :name, :address, :phone, :services
```
juliet-e, line 12: ^^ correct! :)

lib/facility.rb
```diff
@@ -1,14 +1,85 @@
...
   end
 
-  def add_services(service)
+  def add_service(service)
+    @collected_fees = 0
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    if @services.include?('Vehicle Registration')
```
juliet-e, line 36: Nice conditional structure!

lib/facility.rb
```diff
@@ -1,14 +1,85 @@
...
+      
+      if vehicle.antique?
+        vehicle.plate_type = :antique
```
juliet-e, line 39: Refactor Opportunity (for the future, we know time is limited!!): Consider, when you have a long method like this, how you can use helper methods to break up the logic. For example, you could have a helper method here that you call `process_antique_registration` and it can handle setting the plate type and collecting the right amount of fees for this type of car. 

Using helper methods can help a lot when we have complex or nested conditionals, because those get hard to read. If you call helper  methods though, you're able to extract some of that logic into a separate location and give it a readable name to indicate to another dev what is happening.

lib/facility.rb
```diff
@@ -1,14 +1,85 @@
...
+      elsif vehicle.electric_vehicle?
+        vehicle.plate_type = :ev
+        @collected_fees += 200
+      else
+        vehicle.plate_type = :reqular
+        @collected_fees += 100
+      end
+
+      vehicle.registration_date = Date.today
+      @registered_vehicles << vehicle
```
juliet-e, line 52: Nice work pulling these common tasks out of the if statement!

lib/facility.rb
```diff
@@ -1,14 +1,85 @@
...
+
+      return true
+    end
+
+    return false
+  end
+  
+  def renew_drivers_license(registrant)
+    return false unless @services.include?('Renew License')
+    return false unless registrant.license_data[:written] && registrant.license_data[:license]
```
juliet-e, line 78: Nice use of single line if statements (also called guard clauses). These are a great alternative to nested conditionals

lib/registrant.rb
```diff
@@ -0,0 +1,40 @@
+# We should also be able to change a permit from false to true after a Registrant has earned their permit. 
+# There is no age limit to when someone can earn their permit.
+
+require 'pry'
```
juliet-e, line 4: Because you have these gems in your Gemfile, you shouldn't need to require them in every file as long as you run `bundle install` when setting up the repo

lib/registrant.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+
+    def permit?
+        return @permit
+    end
+
+    def earn_permit
+        @permit = true
+    end
+end
```
juliet-e, line 29: This class looks excellent!

lib/vehicle.rb
```diff
@@ -1,18 +1,22 @@
 require 'date'
 
 class Vehicle
-  attr_reader :vin,
-              :year,
-              :make,
-              :model,
-              :engine
+  attr_accessor :vin,
```
juliet-e, line 4: We typically want to avoid attr_accessors as much as possible to maintain data integrity. Allowing other classes to mutate an object's attributes can lead to issues. For example, with this setup, there's nothing stopping our Facility class from setting the `year` attribute to 1234567890, which wouldn't make sense! For the purposes of this project, this is fine, but in applications on the job, we want to try to stick to read-only as much as possible. 

In this case with your current code, you really only need attr_accessors for :registration_date and :plate_type because those are the attributes you will be mutating in the facility class. 

if you wanted to avoid attr_accessors *completely*, the option would be to create some "setter" methods in this class called `assign_plate_type` and `assign_registration_date` and the Facility class could call those methods to make  changes to those attributes. The benefit to doing this is adding some controls in those methods (i.e. an if statement that makes sure that the plate_type is ONLY going to be set to the value of :ev, :antique, etc. Or a conditional to make sure the year is only set to an integer with 4 digits). We wouldn't expect you to add those controls into such a method as part of this project's requirements but I'm sharing this info so you know why we encourage that type of assignment rather than attr_accessors! Let me know if you have any questions about this

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,18 @@
+class VehicleFactory
+    def create_vehicles(registrations = [])
+      vehicles = []
+      registrations.each do |registration|
```
juliet-e, line 4: Nice work! You can also swap out `.each` for `.map` here and avoid the need for your `vehicles` array as an accumulator. But both solutions are good!

spec/facility_spec.rb
```diff
@@ -23,4 +31,35 @@
       expect(@facility.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
     end
   end
+
+  describe '#register_vehicle' do
+    it 'can register vehicles' do
```
juliet-e, line 36: These tests could be more readable and robust if you broke them into a few different test cases - one for antique cars, one for electric vehicles, one for basic cars. 

This test only covers one case -- the Cruz. So it doesn't capture all the possible scenarios in your if statement in the register_vehicle method, but we *do* want to test for every scenario in a if statement.

spec/facility_spec.rb
```diff
@@ -23,4 +31,35 @@
...
+      expect (@facility_1.collected_fees).should == 0
+      expect (@facility_1.register_vehicle(@cruz)).should == [@cruz]
+      # expect (@cruz.plate_type).should == :regular
+      expect (@cruz.registration_date).should == Date.today
+      expect (@facility_1.collected_fees).should == 100
+      # expect(@facility_1.registered_vehicles).should == [@cruz]
+    end
+  end
+
+  describe '#renew_drivers_license' do
```
juliet-e, line 54: Same case here -- using the interaction pattern, try to build out a few test cases to cover the different scenarios:
- registrant cannot renew if the facility doesn't have that service
- registrant cannot renew if they don't have written and license exam set to true
- registrant CAN renew if written and license is true and facility has exam service

These tests will be shorter than what you have here because you're trying to test all the scenarios in one. For simple methods (i.e one simple if statement) that's okay, but this is a more complex method so breaking it up into multiple `it` blocks would be better for troubleshooting, readability and robustness.

spec/facility_spec.rb
```diff
@@ -23,4 +31,35 @@
...
+    it 'describes if a license can be renewed' do
+      @registrant_1.license_data[:written] = true
+      @registrant_1.license_data[:license] = true
+      expect(@facility.renew_drivers_license(@registrant_1)).to eq(false)
+      @registrant_1.license_data[:renewed] = false
+      expect(@facility.renew_drivers_license(@registrant_1)).to eq(true)
+      expect(@facility.renew_drivers_license(@registrant_3)).to eq(false)
+      expect(@facility.renew_drivers_license(@registrant_2)).to eq(false)
+    end
+  end
```
juliet-e, line 64: It looks like you're missing tests for the administer_road_test method -- don't forget to write these tests first to guide your development!

spec/registrant_spec.rb
```diff
@@ -0,0 +1,16 @@
+
```
juliet-e, line 1: Did you have a test file here that didn't get committed?

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+      ]
+      # Create an instance of VehicleFactory
+      factory = VehicleFactory.new
+      # Call create_vehicles method with custom registration data
+      electric_vehicles = factory.create_vehicles(vehicle_data)
+      # Assert that the create_vehicles method returns an array of Vehicle instances
+      expect(electric_vehicles).to be_an(Array)
+      electric_vehicles.each do |vehicle|
+        expect(vehicle).to be_an_instance_of(Vehicle)
+        expect(vehicle.electric_vehicle?).to eq(true)
```
juliet-e, line 21: Nice assertions here! Well done using shorter test data rather than passing in the full dataset -- that's a great option for having more control of the data.

spec/facility_spec.rb
juliet-e: administer_written test needs a test too!

PR #480, opened 2024-05-21
Lito Croy's Project Submission

litobot:


lib/registrant.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  def permit?
+    @permit
+  end
+
+  def earn_permit
+    @permit = true
+  end
+end
```
KatBrandt, line 20: This class looks great, very clean.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,73 @@
+require 'spec_helper'
+
+RSpec.describe Registrant do
+  before(:each) do
+
+    @registrant_1 = Registrant.new('Bruce', 18, true )
+    # pry(main)> registrant_1 = Registrant.new('Bruce', 18, true )
+    # #=> #<Registrant:0x000000015c10bed8 @age=18, @license_data={:written=>false, :license=>false, :renewed=>false}, @name="Bruce", @permit=true>
```
KatBrandt, line 8: For now, if leaving these comments in the project helps you, go ahead, but if you're ever doing a project that you're showing off in github or showing off in a job interview, you'll want to remove comments like this in the final code.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    # pry(main)> registrant_2.name
+    # #=> "Penny"
+  end
+  
+  it "has an age" do
+    expect(@registrant_1.age).to eq(18)
+    # pry(main)> registrant_1.age
+    # #=> 18
+    
+    expect(@registrant_2.age).to eq(15)
```
KatBrandt, line 36: For these attributes and the check if it's an instance of Registrant, really only need to test one instance unless there's any flow control going on.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    # pry(main)> registrant_2.age
+    # #=> 15
+  end
+
+  it "can have a permit (#permit?)" do
+    expect(@registrant_1.permit?).to be(true)
+    # pry(main)> registrant_1.permit?
+    # #=> true
+
+    expect(@registrant_2.permit?).to be(false)  # I think #permit? defaults to false >> check IP for age requirements, etc.
```
KatBrandt, line 46: yup

lib/vehicle.rb
```diff
@@ -22,4 +26,12 @@ def antique?
...
     @engine == :ev
   end
+
+  def set_date
+    @registration_date = Date.today
+  end
+
+  def set_plate_type(plate_type)
+    @plate_type = plate_type
+  end
```
KatBrandt, line 36: Nice, I like that you put these two methods in the Vehicle class, since it's the vehicle attributes that are being changed. Nice job.

KatBrandt: Don't forget to add tests though for helper methods like this!

lib/facility.rb
```diff
@@ -1,14 +1,95 @@
...
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    if services.include?("Vehicle Registration") 
+      @registered_vehicles << vehicle # I don't understand why I have to add to @registered_vehicles
```
KatBrandt, line 24: We're only adding the vehicle to registered vehicles so the facility has an array of vehicle objects that are registered. You're setting the date on line 26.

lib/facility.rb
```diff
@@ -1,14 +1,95 @@
...
+      if vehicle.antique?
+        @collected_fees += 25
+        vehicle.set_plate_type(:antique)
+      elsif vehicle.electric_vehicle?
+        @collected_fees += 200
+        vehicle.set_plate_type(:ev)
+      else 
+        @collected_fees += 100
+        vehicle.set_plate_type(:regular)
+      end
```
KatBrandt, line 36: This logic looks great! To make this even better, you could have taken lines 27-36 and put them in a helper method in this class called say `collect_fees` and then you'd just call that helper method from line 27.

lib/facility.rb
```diff
@@ -1,14 +1,95 @@
...
+    end
+  end
+  
+  #   Administer a written test
+  # A written test can only be administered to registrants with a permit and who are at least 16 years of age
+  def administer_written_test(registrant)
+    if services.include?("Written Test") && registrant.permit? && registrant.age >= 16
```
KatBrandt, line 43: Nice job with the boolean logical operators here.

spec/facility_spec.rb
```diff
@@ -1,26 +1,226 @@
...
-      expect(@facility.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
+      it "can have a vehicle" do
+        expect(@cruz).to be_an_instance_of(Vehicle)
+        expect(@bolt).to be_an_instance_of(Vehicle)
+        expect(@camaro).to be_an_instance_of(Vehicle)
+      end
+    end
+
+    
+    describe '#add service' do
```
KatBrandt, line 40: So this test and the following 4 tests, all are side effects that happen when we call `register_vehicle`. Another way to organize your tests would be to have this outer describe block be `describe "#register_vehicle"`, then have a nested describe or it block to describe each of these side effects.

spec/facility_spec.rb
```diff
@@ -1,26 +1,226 @@
...
+        @facility_1.add_service('Vehicle Registration')
+        @facility_1.register_vehicle(@cruz)
+        @facility_1.register_vehicle(@bolt)
+        @facility_1.register_vehicle(@camaro)
+        
+
+        expect(@cruz.plate_type).to eq(:regular)
+        expect(@camaro.plate_type).to eq(:antique)
+        expect(@bolt.plate_type).to eq(:ev)
+      end
```
KatBrandt, line 123: These are all great tests up to this point, testing all the different scenarios of registering a vehicle depending on the type of car it is.

spec/facility_spec.rb
```diff
@@ -1,26 +1,226 @@
...
+
+        expect(@facility_1.add_service('Written Test')).to eq(["Written Test"])
+
+        expect(@facility_1.administer_written_test(@registrant_2)).to eq(true)
+        expect(@facility_1.administer_written_test(@registrant_1)).to eq(true)
+
+        expect(@registrant_1.license_data).to eq({:written => true, :license => false, :renewed => false})
+        expect(@registrant_2.license_data).to eq({:written => true, :license => false, :renewed => false})
+        expect(@registrant_3.license_data).to eq({:written => false, :license => false, :renewed => false})
+      end
```
KatBrandt, line 155: This test does need a lot because there are 3 conditions that must be met in order for the method to return true, so we have to test all combinations of these conditions to make sure we get the desired return value. So if your three conditions are false/false/false, what do we get? If the conditions are false/false/true, what do we get? When they're true/false/false, etc....  So breaking up each of those scenarios into different it blocks explaining what each one is testing for can be helpful in reading these tests.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,20 @@
...
+      vehicle_deets = {}
+      vehicle_deets[:vin] = potential_vehicle[:vin_1_10]
+      vehicle_deets[:year] = potential_vehicle[:model_year]
+      vehicle_deets[:make] = potential_vehicle[:make]
+      vehicle_deets[:model] = potential_vehicle[:model]
+      vehicle_deets[:engine] = :ev
+
+      Vehicle.new(vehicle_deets)
+    end
+  end
```
KatBrandt, line 18: Yup, this works and is very clean. Nice job. Note: if there is nothing in the initialize method, you do not need to include it at all.

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    wa_ev_registrations = DmvDataService.new.wa_ev_registrations
+    
+    wa_example = factory.create_vehicles(wa_ev_registrations)
+    
+    expect(wa_example[0].vin).to eq("1G1FZ6S08P")
+    
+    wa_example.each do |vehicle|
+      expect(vehicle).to be_an_instance_of(Vehicle)
+      expect(vehicle).to respond_to(:vin)
+      expect(vehicle).to respond_to(:year)
```
KatBrandt, line 26: We can also test that each of the attributes we set in the create_vehicles method is actually being set. We don't need to test the exact values, but we can test that the attribute is a String and it's not nil.

lib/facility_factory.rb
```diff
@@ -0,0 +1,63 @@
+class FacilityFactory
+  def initialize
+
+  end
+
+  def create_facilities(facility_database)
+    if facility_database.first[:state] == "CO"
+      return build_co_facilities(facility_database)
```
KatBrandt, line 8: `return` keyword not needed, Ruby will do this automatically = implicit return.

lib/facility_factory.rb
```diff
@@ -0,0 +1,63 @@
...
+      facility_deets[:phone] = potential_facility[:phone]
+      facility_deets[:address] = "#{potential_facility[:address1]} " \
+                                  "#{potential_facility[:city]} " \
+                                  "#{potential_facility[:state]} " \
+                                  "#{potential_facility[:zipcode]}" 
+      facility_deets[:services] = [] if potential_facility[:services_p].nil?
+
+      Facility.new(facility_deets)
+    end
+  end
```
KatBrandt, line 61: I really like your use of helper methods in this class. Reading the flow control is super easy because of these helper methods. This class looks good.

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    co_example = facility_factory.create_facilities(co_dmv_office_locations)
+    
+    expect(co_example[0].phone).to eq("(720) 865-4600")
+    
+    co_example.each do |facility|
+      expect(facility).to be_an_instance_of(Facility)
+      expect(facility).to respond_to(:name)
+      expect(facility).to respond_to(:phone)
+      expect(facility).to respond_to(:address)
+      expect(facility).to respond_to(:services)
```
KatBrandt, line 32: If it's a facility object, we've already tested that it will respond to these methods (from the facility spec), but we can check that when we call these methods, we're not getting nil, so we know the values are being set.

KatBrandt (commented):
Lito! Overall, really great job on this. Please review my feedback and let me know if you have any questions.

PR #481, opened 2024-05-21
Candice Cirbo DMV Project

CCirbo:


lib/dmv.rb
```diff
@@ -9,7 +10,7 @@ def add_facility(facility)
   end
 
   def facilities_offering_service(service)
-    @facilities.find do |facility|
+    @facilities.select do |facility|
```
juliet-e, line 13: nice work! `select` or `find_all` are indeed better options

lib/facility.rb
```diff
@@ -1,14 +1,81 @@
...
+    @collected_fees = 0
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    return nil unless @services.include?('Vehicle Registration')
```
juliet-e, line 26: great use of a guard clause here to avoid nested if statements!

lib/facility.rb
```diff
@@ -1,14 +1,81 @@
...
+    if vehicle.antique?
```
juliet-e, line 27: Refactor Opportunity (for the future, we know time is limited!!): Consider, when you have a long method like this, how you can use helper methods to break up the logic. For example, you could have a helper method here that you call `process_antique_registration` and it can handle setting the plate type, setting registration date, collecting the right amount of fees for this type of car, etc.

lib/facility.rb
```diff
@@ -1,14 +1,81 @@
...
+    end
+  end
+
+  def renew_drivers_license(registrant)
+    return false unless @services.include?('Renew License')
+    if registrant.license_data[:written] == true && registrant.license_data[:license] = true
+      registrant.license_data[:renewed] = true
+    else
+      false  
+    end
```
juliet-e, line 74: These last few methods look great! Great conditional structuring and logic

lib/facility_factory.rb
```diff
@@ -0,0 +1,30 @@
+class FacilityFactory 
+    def initialize
+        
+    end
+
+    def create_facility(facility_raw_data)
+        facility_raw_data.map do |facility|
```
juliet-e, line 7: Well done making it through this method! Adding in new states adds a good deal of complexity really quickly. 

Similar to my last comment above, this becomes a pretty long method so using helper methods would be a cleaner option. Consider creating a helper method for each state and then delegating to that helper method from this top level create_facilities method:

```
def create_facilities(raw_data)
    if raw_data.first[:state] == "CO"
        return create_co_facilities(raw_data)
    elsif raw_data.first[:state] == "MO"
        return create_mo_facilities(raw_data)
    elsif raw_data.first[:state] == "NY" 
        return create_ny_facilities(raw_data)
    else 
       return []
    end
end

def build_co_facilities(raw_data)
    raw_data.map do |facility|
        # process facility and return Facility object
    end
end
```

lib/vehicle.rb
```diff
@@ -7,12 +7,18 @@ class Vehicle
               :model,
               :engine
 
+  attr_accessor :registration_date,
```
juliet-e, line 10: This is a great option for this project

On the job, however, you'll see that we generally want to avoid attr_accessors  as much as possible to maintain data integrity. if you wanted to avoid attr_accessors *completely*, the option would be to create some "setter" methods in this class called `assign_plate_type` and `assign_registration_date` and the Facility class could call those methods to make  changes to those attributes. The benefit to doing this is adding some controls in those methods (i.e. an if statement that makes sure that the plate_type is ONLY going to be set to the value of :ev, :antique, etc. Or a conditional to make sure the year is only set to an integer with 4 digits). We wouldn't expect you to add those controls into such a method as part of this project's requirements but I'm sharing this info so you know why we encourage that type of assignment rather than attr_accessors! Let me know if you have any questions about this

Again - using attr_accessors here is totally fine and we would never mark you down for using them, but it's good to know why you won't see them out in the world very much!

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,24 @@
+class VehicleFactory 
+    def initialize
```
juliet-e, line 2: You don't need to even include this initialize when there's no arguments or attributes to set!

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    end
+   
+    def create_vehicles(vehicle_raw_data)
+        vehicle_raw_data.map do |vehicle|
```
juliet-e, line 7: Great use of .map!

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+            expect(@ffactory.create_facility(@mo_facilities).length).to eq(178)
+            expect(@ffactory.create_facility(@mo_facilities)[0]).to be_an_instance_of(Facility)
+            expect(@ffactory.create_facility(@mo_facilities)[0].name).to eq("NEVADA")
+            expect(@ffactory.create_facility(@mo_facilities)[0].address).to eq("409 E AUSTIN BLVD NEVADA MO 64772")
+            expect(@ffactory.create_facility(@mo_facilities)[0].phone).to eq("(417) 448-5705")
+            expect(@ffactory.create_facility(@mo_facilities)[3].address).to eq("237 W MAIN WARSAW MO 65355")
+
+        end
+    end
+
```
juliet-e, line 53: Excellent robust testing!

spec/facility_spec.rb
```diff
@@ -1,26 +1,156 @@
...
+      @facility_2.add_service('Renew Drivers License')
+      @facility_2.add_service('Vehicle Registration')
+      expect(@facility.services).to eq(['New Drivers License', 'Renew Drivers License'])
+      expect(@facility_1.services).to eq(['New Drivers License', 'Vehicle Registration'])
+      expect(@facility_2.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
+    end   
+  end
+
+  describe '#register_vehicle' do
+    it 'can register vehicles' do
```
juliet-e, line 60: I would break this up into different scenarios just so it's easier to read and troubleshoot:
- it 'can register antique vehicles'
- it 'can register electric vehiecles'
- it 'can register basic vehicles as long as service is available'

You're covering all the scenarios but in one test case it just gets a bit long

spec/facility_spec.rb
```diff
@@ -1,26 +1,156 @@
...
+      expect(@facility_1.register_vehicle(@bolt)).to eq([@cruz, @camaro, @bolt])
+      expect(@bolt.plate_type).to eq(:ev)
+      expect(@bolt.registration_date).to eq(Date.today)
+      expect(@facility_1.collected_fees).to eq(325)
+      expect(@facility_1.registered_vehicles).to eq([@cruz, @camaro, @bolt]) 
+    end
+  end   
+
+  describe '#administer_written_test' do
+    it 'administer a written test' do
```
juliet-e, line 92: Same here -- can you break this up into different scenarios with specific descriptions? i.e. it `can administer a written test if registrant has a permit`

spec/facility_spec.rb
```diff
@@ -1,26 +1,156 @@
...
+
+      expect(@facility_1.administer_written_test(@registrant_3)).to eq(false)
+      @registrant_3.earn_permit
+      expect(@facility_1.administer_written_test(@registrant_3)).to eq(false)
+      expect(@registrant_3.license_data[:written]).to eq(false)
+      end
+    end   
+  
+  describe '#administer_road_test' do
+    it 'administer a road test' do
```
juliet-e, line 111: Here too! Multiple it statements in the same describe block would be great to make this more readable

spec/facility_spec.rb
```diff
@@ -1,26 +1,156 @@
...
+      @facility_1.add_service('Renew License')
+      expect(@facility_1.renew_drivers_license(@registrant_1)).to eq(true)
+      expect(@registrant_1.license_data[:renewed]).to eq(true)
+
+      expect(@facility_1.renew_drivers_license(@registrant_3)).to eq(false)
+      expect(@registrant_3.license_data[:renewed]).to eq(false)
+    
+      expect(@facility_1.renew_drivers_license(@registrant_2)).to eq(true)
+      expect(@registrant_2.license_data[:renewed]).to eq(true)
+    end
```
juliet-e, line 154: Your tests are excellent and robust. The focus will just be on breaking them up into separate cases, but well done covering all the scenarios!

spec/registrant_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+  before (:each) do
+    @registrant_1 = Registrant.new('Bruce', 18, true)  
+    @registrant_2 = Registrant.new('Penny', 16)
+    @registrant_3 = Registrant.new('Tucker', 15)
+        
+  end
+
+    describe '#initialize' do
+      it 'can initialize with attributes' do
+        expect(@registrant_1).to be_an_instance_of(Registrant)
```
juliet-e, line 20: It's sufficient to just test this with one registrant object. No need to do 3

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+        expect(@factory.create_vehicles(@wa_ev_registrations)[0].year).to eq("2024")
+        expect(@factory.create_vehicles(@wa_ev_registrations)[0].make).to eq("NISSAN")
+        expect(@factory.create_vehicles(@wa_ev_registrations)[0].model).to eq("Leaf")
+      
+        expect(@factory.create_vehicles(@wa_ev_registrations)[250].vin).to eq("5YJYGAEE3M")
+        expect(@factory.create_vehicles(@wa_ev_registrations)[250].year).to eq("2021")
+        expect(@factory.create_vehicles(@wa_ev_registrations)[250].make).to eq("TESLA")
+        expect(@factory.create_vehicles(@wa_ev_registrations)[250].model).to eq("Model Y")
+      
+    end
```
juliet-e, line 32: Awesome test!

lib/facility.rb
```diff
@@ -1,14 +1,81 @@
...
-  def initialize(name, address, phone)
-    @name = name
-    @address = address
-    @phone = phone
-    @services = []
+  def initialize(info_hash, services = [])
+    @name = info_hash[:name]
+    @address = info_hash[:address]
+    @phone = info_hash[:phone]
+    @services = services
```
juliet-e, line 16: This is indeed a possible solution for handling services! Nice work! You could also just keep the one argument as info_hash, and in this initialize here  with a conditional:
```
if info_hash.has_key?(:services)
   @services = info_hash[:services]
else
   @services = []
end
```

PR #482, opened 2024-05-21
Project Turn in - The Dmv

tnoble-cmd:
Final main Repo


lib/registrant.rb
```diff
@@ -0,0 +1,19 @@
...
+    @license_data = {written: false, license: false, renewed: false}
+  end
+
+  def permit?
+    @permit
+  end
+
+  def earn_permit
+    @permit = true
+  end
```
KatBrandt, line 18: Nice job with this class, looks good.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+     expect(registrant.name).to eq('Bill')
+     expect(registrant.age).to eq(25)
+     expect(registrant.permit?).to eq(true)
+    end
+  end
+
+  describe '#permit?' do
+    it 'returns permit status' do
+     registrant = Registrant.new('Bill', 25, true)
+     expect(registrant.permit?).to eq(true)
```
KatBrandt, line 17: No need for this test if you're testing it on line 10.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+  end
+
+  describe '#earn_permit' do
+    it 'can earn a permit' do
+     registrant = Registrant.new('Bill', 15)
+     expect(registrant.permit?).to eq(false)
+     registrant.earn_permit
+     expect(registrant.permit?).to eq(true)
```
KatBrandt, line 26: This is a good test, showing the initial state and then the changed state.

spec/vehicle_spec.rb
```diff
@@ -33,4 +39,40 @@
...
+      expect(@cruz.plate_type).to eq(nil)
+      expect(@bolt.plate_type).to eq(nil)
+      expect(@camaro.plate_type).to eq(nil)
+      @cruz.set_plate_type
+      @bolt.set_plate_type
+      @camaro.set_plate_type
+      expect(@cruz.plate_type).to eq(:regular)
+      expect(@bolt.plate_type).to eq(:ev)
+      expect(@camaro.plate_type).to eq(:antique)
+    end
```
KatBrandt, line 76: I recommend using white space in your tests to separate different sections, so I would put a white space under line 69 and under line 72. It helps us read tests. 
Great job testing your helper methods you created!

lib/vehicle.rb
```diff
@@ -22,4 +27,22 @@ def antique?
...
+  end
+
+  def set_plate_type
+    if @engine == :ev
+      @plate_type = :ev
+    elsif Date.today.year - @year > 25
+      @plate_type = :antique
+    else
+      @plate_type = :regular
+    end
```
KatBrandt, line 46: I like that you created these helper methods in the Vehicle class. And good job testing them.

lib/vehicle.rb
```diff
@@ -5,14 +5,19 @@ class Vehicle
               :year,
               :make,
               :model,
-              :engine
+              :engine,
+              :registration_date
+              
+  attr_accessor :plate_type
```
KatBrandt, line 11: We do not want an attr_accessor for plate_type since you created a method called `set_plate_type` that will do this for us. I like the helper method better than having an attr_accessor.

lib/facility.rb
```diff
@@ -1,14 +1,60 @@
 class Facility
   attr_reader :name, :address, :phone, :services
+  attr_accessor :collected_fees, :register_vehicles
```
KatBrandt, line 3: We don't need these in attr_accessors. Adding a method to attr_accessor allows other classes to change these attributes, so we don't want to expose that functionality unless we absolutely have to. These should just be attr_readers.

lib/facility.rb
```diff
@@ -1,14 +1,60 @@
...
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    vehicle.register
+    vehicle.set_plate_type
+    @register_vehicles << vehicle
+  end
```
KatBrandt, line 23: I like that this method has a bunch of helper methods in it, however, we're missing 2 pieces of logic from this method: 1. It should first be checking that the facility has the service of 'Vehicle Registration', if it doesn't, then we should not do any of these steps. 2. This method should also be calling the `calculate_fee` method I see below since upon registration, fees should be collected automatically.

lib/facility.rb
```diff
@@ -1,14 +1,60 @@
...
+  
+  def administer_written_test(registrant)
+    if @services.include?('Written Test') != true
+      return 'This facility does not offer written tests.'
+    elsif @services.include?('Written Test') && registrant.permit? && registrant.age >= 16
+      registrant.license_data[:written] = true   
+    elsif @services.include?('Written Test') && registrant.age < 16
+      return 'Registrant is too young to take the written test.'
+    end
```
KatBrandt, line 42: No need for the `return` keywords here. Ruby operates by implicit return so it will automatically return these strings if it goes down that branch of logic.

lib/facility.rb
```diff
@@ -1,14 +1,60 @@
...
+      @collected_fees += 25
+    else
+      @collected_fees += 100
+    end
+  end
+  
+  def administer_written_test(registrant)
+    if @services.include?('Written Test') != true
+      return 'This facility does not offer written tests.'
+    elsif @services.include?('Written Test') && registrant.permit? && registrant.age >= 16
```
KatBrandt, line 38: good job with the boolean logical operators in this conditional.

spec/facility_spec.rb
```diff
@@ -23,4 +23,127 @@
...
+    end
+
+    it 'can calculate the fee for an antique vehicle' do
+      camaro = Vehicle.new({vin: '1a2b3c4d5e6f', year: 1969, make: 'Chevrolet', model: 'Camaro', engine: :ice})
+      @facility.register_vehicle(camaro)
+      expect(@facility.register_vehicles.length).to eq(1)
+      expect(@facility.collected_fees).to eq(0)
+      @facility.calculate_fee(camaro)
+      expect(@facility.collected_fees).to eq(25)
+    end
```
KatBrandt, line 63: Good job testing the 3 different scenarios for this method.

spec/facility_spec.rb
```diff
@@ -23,4 +23,127 @@
...
+    it 'can not renew drivers license' do
+      registrant = Registrant.new('Bill', 25, true)
+      expect(@facility.services).to eq([])
+      @facility.add_service('Written Test')
+      @facility.add_service('Road Test')
+      @facility.administer_written_test(registrant)
+      @facility.administer_road_test(registrant)
+      @facility.renew_drivers_license(registrant)
+      expect(@facility.renew_drivers_license(registrant)).to eq('This facility does not offer road tests.')
+    end
```
KatBrandt, line 148: Overall good job testing all the conditions needed for these methods, it's a lot. You can definitely make more it blocks for each set of conditions you're testing for, and definitely start adding white space to your tests for readability.

spec/facility_spec.rb
KatBrandt: However, this it block was for the collect_fees method, so we would want to just test that method in isolation here. We also want to have another describe block for the register_vehicle method and in that block we would test all the different side effects that happen when we call that method.

lib/facility_generator.rb
```diff
@@ -0,0 +1,45 @@
...
+  def create_facility(facility)
+    facility.map do |data|
+      if data[:state] == "CO"
+        co_facility(data)
+      elsif data[:state] == "NY"
+        ny_facility(data)
+      elsif data[:state] == "MO"
+        mo_facility(data)
+      end
+    end
```
KatBrandt, line 11: Nice job with the helper methods here, definitely makes it easier to read and debug.

KatBrandt: Yeah, this class looks good, nice job.

spec/facility_generator_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+  end
+  describe '#co_facility' do
+    it 'can create a CO facility and create a new instance of facility' do
+      
+      expect(@fg.co_facility(@co[0])).to be_an_instance_of(Facility)
+      expect(@fg.co_facility(@co[0]).name).to eq("DMV Tremont Branch")
+      expect(@fg.co_facility(@co[0]).address).to eq("2855 Tremont Place, Suite 118, Denver, CO 80205")
+      expect(@fg.co_facility(@co[0]).phone).to eq("(720) 865-4600")
+    end
```
KatBrandt, line 29: The only problem with testing for specific data, is that this data is coming from the api, we don't control it. So tomorrow, these tests might fail if CO adds a new facility tomorrow morning. We want our tests to test the functionality of our method and fail only if our method fails, not if the data changes. So we could instead test that each of these attributes is not nil so we know they're set to something.

PR #485, opened 2024-05-21
James-Cochran/the_dmv

James-Cochran:


lib/facility.rb
```diff
@@ -1,14 +1,65 @@
...
+    @registered_vehicles = []
+    @collected_fees = 0
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
```
juliet-e, line 24: Make sure to double check here if the facility has "Vehicle Registration" as one of its services!

lib/facility.rb
```diff
@@ -1,14 +1,65 @@
...
+    vehicle.registration_date = Date.today
+    if vehicle.antique?
```
juliet-e, line 26: Refactor Opportunity (for the future, we know time is limited!!): Consider, when you have a long method like this, how you can use helper methods to break up the logic. For example, you could have a helper method here that you call `process_antique_registration` and it can handle setting the plate type, setting registration date, collecting the right amount of fees for this type of car, etc.

lib/facility.rb
```diff
@@ -1,14 +1,65 @@
...
+    elsif vehicle.electric_vehicle? 
+      vehicle.plate_type = :ev
+      @collected_fees += 200
+    else vehicle.plate_type = :regular
+      @collected_fees += 100
+    end
+    @registered_vehicles << vehicle
+  end 
+
+  def eligible_for_written_test?(registrant)
```
juliet-e, line 38: Great use of a helper method here! Excellent

lib/facility.rb
```diff
@@ -1,14 +1,65 @@
...
+    else
+      false
+    end
+  end
+
+  def renew_license(registrant)
+    if registrant.license_data[:license]
+      registrant.license_data[:renewed] = true
+    else
+      false
```
juliet-e, line 62: For all these methods, similar to my comment above, you need to check if the service is available at that facility. You could do this with a guard clause at the beginning of the method to avoid a nested if statement:

`return false unless @services.include("License Renewal")`

lib/registrant.rb
```diff
@@ -0,0 +1,24 @@
...
+    end
+
+    def permit?
+        @permit
+    end
+
+    def earn_permit
+        @permit = true
+    end
+
```
juliet-e, line 22: Great work!

lib/vehicle.rb
```diff
@@ -6,17 +6,22 @@ class Vehicle
               :make,
               :model,
               :engine
+              
+  attr_accessor :registration_date,
```
juliet-e, line 10: This is a great option for this project to allow vehicle attributes to be editable in the facility class

On the job, however, you'll see that we generally want to avoid attr_accessors  as much as possible to maintain data integrity. if you wanted to avoid attr_accessors *completely*, the option would be to create some "setter" methods in this class called `assign_plate_type` and `assign_registration_date` and the Facility class could call those methods to make  changes to those attributes. The benefit to doing this is adding some controls in those methods (i.e. an if statement that makes sure that the plate_type is ONLY going to be set to the value of :ev, :antique, etc. Or a conditional to make sure the year is only set to an integer with 4 digits). We wouldn't expect you to add those controls into such a method as part of this project's requirements but I'm sharing this info so you know why we encourage that type of assignment rather than attr_accessors! Let me know if you have any questions about this

Again - using attr_accessors here is totally fine and we would never mark you down for using them, but it's good to know why you won't see them out in the world very much!

spec/facility_spec.rb
```diff
@@ -1,26 +1,89 @@
...
+    end
+
+    it 'records registration date' do
+      @facility_1.register_vehicle(@cruz)
+      expect(@cruz.registration_date).to eq(Date.today)
+    end
+
+    it 'can collect fees' do
+      @facility_1.register_vehicle(@bolt)
+      expect(@facility_1.collected_fees).to eq(200)
```
juliet-e, line 50: It's a good idea to check that the fees are collected correctly for each type of plate, i.e. you have a scenario for registering every type of car and verify the collected fees are correct for that amount. Since you have a specific test for the fee collection, you could put all 3 of these expect statements in the same test case

spec/facility_spec.rb
```diff
@@ -1,26 +1,89 @@
...
+    end
+
+    it 'can determine plate type' do
+      @facility_1.register_vehicle(@camaro)
+      expect(@camaro.plate_type).to eq(:antique)
```
juliet-e, line 55: Same here -- check the plate type is assigned correctly based on the type of vehicle

spec/facility_spec.rb
```diff
@@ -1,26 +1,89 @@
...
+  end
+
+  describe '#renew_license' do
+    it  'verifies if registrant is elegible to renew' do
+      @registrant_3.license_data[:license] = true
+      expect(@facility_1.renew_license(@registrant_3)).to eq(true)
     end
   end
 end
+
```
juliet-e, line 84: the only test case you're missing here is the outcome when a facility tries to do something when it is a service that is not included in its @services array

spec/registrant_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+        it "can determine if registrant has a permit" do
+            expect(@registrant_1.permit?).to eq(true)
+            expect(@registrant_2.permit?).to eq(false)
+        end
+    end
+
+    describe "#earn_permit" do
+        it "can change if registrant has a permit" do
+            expect(@registrant_1.earn_permit).to eq(true)
+            expect(@registrant_2.earn_permit).to eq(true)
```
juliet-e, line 39: Consider checking the `permit?` method after this as well to verify it has changed

spec/registrant_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+        end
+    end
+end
+
```
juliet-e, line 43: Great test!

PR #486, opened 2024-05-22
Kaelin Salazar

kaelinpsalazar:


lib/registrant.rb
```diff
@@ -0,0 +1,23 @@
...
+    def permit?
+        @permit
+    end
+
+    def earn_permit
+        @permit = true
+    end
+
+   
+
```
KatBrandt, line 19: Are you leaving space for a party here? :)

lib/registrant.rb
```diff
@@ -0,0 +1,23 @@
+class Registrant 
+    attr_reader :name, :age, :permit, :license_data
```
KatBrandt, line 2: Careful here, we don't want to create a method called `permit` since we have the method `permit?` that does the same function and follows Ruby convention.

lib/vehicle.rb
```diff
@@ -1,25 +1,41 @@
...
     Date.today.year - @year > 25
   end
 
-  def electric_vehicle?
+  def electric?
     @engine == :ev
   end
+
+  def plate_type
+    calculate_plate_type
```
KatBrandt, line 27: So you've created a method here called `plate_type`, however your attr_reader has also exposed a method called `plate_type`, so you essentially have 2 methods called the same thing. I would just get rid of lines 26-28, set @plate_type on line 14 to nil, and then call `calculate_plate_type` from the Facility class when a vehicle is being registered. Or you can have it called here right away upon initialization.

lib/facility.rb
```diff
@@ -1,14 +1,78 @@
...
   end
-
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
-end
+
+  def register_vehicle(vehicle)
+    # return false unless @services.include?('Register Vehicle')
```
KatBrandt, line 17: yes, we do need to check first if the services include that string, otherwise we don't want to do any of this logic. You could have on line 17: `if @services.include?("Register Vehicle")` and then put all the logic inside that block.

lib/facility.rb
```diff
@@ -1,14 +1,78 @@
...
+    vehicle.registration_date = Date.today
+    @registered_vehicles << vehicle
+    vehicle
+  end
```
KatBrandt, line 21: We also want the facility to collect fees upon registration, so after line 19, we'd want to call `collect_fees`

KatBrandt: Ahhh, actually, I see what you're doing. You have a method below, `collected_fees` that will run through all the registered vehicles and calculate the total when that method is called. That certainly is a valid way to do that, however, if this is how you're doing the fees, then you don't need an instance variable of `@collected_fees` to store that data since it's being calculated dynamically whenever you call the `collected_fees` method. 
Another thing to keep in mind, is that usually variables are nouns like `collected_fees` whereas methods represent actions so we name name verbs like `collect_fees` or `calculate_fees`

lib/facility.rb
```diff
@@ -1,14 +1,78 @@
...
+    elsif vehicle.electric?
+      200
+    else
+      100
+    end
+  end
+
+  def administer_written_test(registrant)
+    # return false unless @services.include?('Written Test')
+    if registrant.permit? && registrant.age >= 16
```
KatBrandt, line 42: We can add that 3rd condition here to this line so it would read `if registrant.permit? && registrant.age >= 16 && @services.include?('Written Test')`

spec/facility_spec.rb
```diff
@@ -14,13 +29,92 @@
...
       @facility.add_service('Renew Drivers License')
       @facility.add_service('Vehicle Registration')
-      expect(@facility.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
+      @facility.add_service('Written Test')
+      @facility.add_service ('Renew License')
+      expect(@facility.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration','Written Test','Renew License'])
+    end
+  end
+
+
```
KatBrandt, line 46: Just one line of white space between an end and the next describe

spec/facility_spec.rb
```diff
@@ -14,13 +29,92 @@
...
+
+  describe 'register vehicle' do
+    it 'registers a vehicle' do
+      @facility.register_vehicle(@cruz)
+      expect(@facility.registered_vehicles).to include(@cruz)
```
KatBrandt, line 51: Instead of checking that the return value includes @cruz, we want to be more explicit. With this setup, there should be exactly one Vehicle in this array, so it should read `expect(@facility.registered_vehicles).to eq([@cruz])` The more explicit our tests, the better.

spec/facility_spec.rb
```diff
@@ -14,13 +29,92 @@
...
+
+    it 'returns $200 for electric vehicles' do
+      @facility.register_vehicle(@bolt)
+      expect(@facility.collected_fees).to eq(200)
+    end
+
+    it 'returns $100 for all other vehicles' do
+      @facility.register_vehicle(@cruz)
+      expect(@facility.collected_fees).to eq(100)
+    end
```
KatBrandt, line 78: Good job testing for all three scenarios in this method.

spec/facility_spec.rb
```diff
@@ -14,13 +29,92 @@
...
+  end
+
+
+  describe 'administer_written_test'do
+      it 'decides if the written test will be taken' do
+      expect(@facility.administer_written_test(@registrant_1)).to eq(true)
+      expect(@registrant_1.license_data[:written]).to eq(true)
+    end
```
KatBrandt, line 86: This method (and the following two methods) has 3 conditions that must be met in order for the return value to be true. That means we need to test all the different scenarios to ensure that we are always getting the correct return value. So we should be testing that if the facility doesn't have the service, it should return false, even if the registrant is 16 and they have their permit. But if the facility does have the service and the registrant is 13, then this should return false, etc.
When testing, we want to think about all possible return values and test for all of those.

lib/facility_creator.rb
```diff
@@ -0,0 +1,27 @@
...
+          phone: facility_info[:phone] || facility_info[:public_phone_number]
+        }
+  
+        facility = Facility.new(facility_details)
+        facilities << facility
+      end
+  
+      facilities
+    end
+  end
```
KatBrandt, line 21: Overall, this looks good. Great job using .each and checking for all possible keys of a given attribute.

spec/facility_creator_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+      @facilities_info = [
+        { address: '123 Main St', city: 'Anytown', state: 'ST', zip: '12345', phone: '555-1234' },
+        { address: '456 Elm St', city: 'Othertown', state: 'ST', zip: '54321', phone: '555-5678' }
+      ]
+    end
+
+    it 'creates a facility' do    
+      facility_creator = FacilityCreator.new
+
+      facilities = facility_creator.create_facilities(@facilities_info)
```
KatBrandt, line 15: I like that you're using fake data to test your method. However, because you are using fake data, that means you know exactly what data the Facility objects will have and we should be testing for that. So on line 19, in addition to testing that you've created a facility, we want to test that all data got set correctly so we can add `expect(facility.address).to eq("123 Main St"), etc. That way we know our method is properly assigning these values.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,23 @@
...
+        model: registration[:model],
+        engine: :ev,
+        registration_date: nil 
+      }
+
+      vehicle = Vehicle.new(vehicle_details)
+      vehicles << vehicle
+    end
+
+    vehicles
```
KatBrandt, line 21: Yup, great job.

PR #490, opened 2024-05-22
the_dmv_cp

CPPittman:


lib/dmv.rb
```diff
@@ -9,7 +9,7 @@ def add_facility(facility)
   end
 
   def facilities_offering_service(service)
-    @facilities.find do |facility|
+    @facilities.find_all do |facility|
```
juliet-e, line 12: Great choice of enumerables here

lib/registrant.rb
```diff
@@ -0,0 +1,25 @@
+require './lib/registrant'
+
```
juliet-e, line 2: Here, we would want to create the Registrant class something like:
```
class Registrant

   attr_reader :name, :age, :permit
 
          @name = name
          @age = age
          @permit = permit
     end
.... more methods below
```

Be careful of copying and pasting interaction pattern code. That is code that you can use to create tests in order to build out your functionality, but it shouldn't got straight into class files or spec files

spec/facility_spec.rb
```diff
@@ -24,3 +24,282 @@
     end
   end
 end
+
+require './lib/facility' = true
+
+require './lib/vehicle' = true
+
+facility_1 = Facility.new({name: 'DMV Tremont Branch', address: '2855 Tremont Place Suite 118 Denver CO 80205', phone: '(720) 865-4600'})
```
juliet-e, line 32: Try to use a reference spec to guide you to write your tests. The interaction pattern will show you what kind of test cases to write, but check out the rest of this facility spec to follow the pattern created there with `it` blocks and `expect` statements. 

Again, we don't want you to copy and paste the interaction pattern,  but use  it to answer these questions:
- what attributes should the class set up objects with
- what methods should the objects have?

From there, you can write a test for every method and attribute

PR #491, opened 2024-05-22
Karl Fallenius

SmilodonP:


lib/registrant.rb
```diff
@@ -0,0 +1,22 @@
+class Registrant
+  attr_reader :name,
+              :age,
+              :permit,
```
KatBrandt, line 4: Careful here, by having `:permit` in the attr_reader, you are exposing a method called `.permit`, but we never asked for that method AND we have a method called `.permit?` that gives us this functionality. We don't typically want two methods of the same name with the same functionality.

lib/registrant.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+  
+  def earn_permit
+    @permit = true
+  end
+
+  def permit?
+    @permit
+  end
+
```
KatBrandt, line 21: This cleans looks good, very clean

spec/registrant_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    @registrant_1 = Registrant.new('Bruce', 18, true)
+    @registrant_2 = Registrant.new('Penny', 16)
+    @registrant_3 = Registrant.new('Tucker', 15)
+  end
+
+  describe '#initialize' do
+    it 'can initialize' do
+      expect(@registrant).to be_an_instance_of(Registrant)
+      expect(@permit).to be false
+    end
```
KatBrandt, line 15: We sure to test all attr_reader methods, so we still want to test that `@registrant_1.name` is Bruce.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'spec_helper'
+
+RSpec.describe Registrant do
+  before(:each) do
+    @registrant = Registrant.new(@name, @age, @permit = false)
```
KatBrandt, line 5: We don't want this line. You have 3 registrant objects below that are the ones you'd want to use for your testing. This one on line 5 is a registrant object with it's attributes set to nil, so not helpful for testing.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  describe '#initialize' do
+    it 'can initialize' do
+      expect(@registrant).to be_an_instance_of(Registrant)
+      expect(@permit).to be false
+    end
+  end
+
+  describe 'license data' do
+    it 'has no license data by default' do
+      expect(@registrant.license_data).to eq({:written => false, :license => false, :renewed => false})
```
KatBrandt, line 20: Great job here testing the default values.

lib/facility.rb
```diff
@@ -1,14 +1,96 @@
...
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
-end
+
+  # VEHICLE REGISTRATION
+  def assign_plate_type(vehicle)
+    if vehicle.antique? == true
```
KatBrandt, line 26: If you have a method that evaluates to a boolean, then you don't need to compare it to a boolean. So `vehicle.antique?` will already return true or false, no need to then also compare it to true.

lib/facility.rb
```diff
@@ -1,14 +1,96 @@
...
+      vehicle.plate_type = :antique
+    elsif vehicle.electric_vehicle? == true
+      vehicle.plate_type = :ev
+    else 
+      vehicle.plate_type = :regular
+    end
+  end
+
```
KatBrandt, line 34: Only one line of white space between methods.

lib/facility.rb
```diff
@@ -1,14 +1,96 @@
...
+  def register_vehicle(vehicle)
+    if @services.include?("Register Vehicles")
+      vehicle.registration_date = Date.today
+      assign_plate_type(vehicle)
+      
+      @registered_vehicles << vehicle
+      fees_due(vehicle)
+
```
KatBrandt, line 53: No white space here.

lib/facility.rb
```diff
@@ -1,14 +1,96 @@
...
+      vehicle.registration_date = Date.today
+      assign_plate_type(vehicle)
+      
+      @registered_vehicles << vehicle
+      fees_due(vehicle)
+      
+    else
+      print "This Facility Does Not Offer This Service"
+    end
+  end
```
KatBrandt, line 57: I like that you used a bunch of helper methods to help with the logic of this method. It makes this method a lot more readable. Typically though, the main methods come before the helper methods in the class, that way we can quickly see the main functionality.

lib/facility.rb
```diff
@@ -1,14 +1,96 @@
...
+
+  #DRIVER'S LICENSE
+
+  #WRITTEN TEST
```
KatBrandt, line 61: Comments are ok for your beginning projects if you need them for guidance, but in production code, we don't want or need comments like these unless something is really unclear. The code usually speaks for itself. :)

lib/facility.rb
```diff
@@ -1,14 +1,96 @@
...
+
+  def drivers_license_test_prerequisites(registrant)
+    if registrant.permit? == true && registrant.age >= 16
+      true
+    end
```
KatBrandt, line 66: So here, we can refactor this to whole method to be only one line of code that reads `registrant.permit? && registrant.age >=16`, because that statement itself will evaluate to true or false because of the `&&`, so no need to  have an if statement. Nice job encapsulating this logic into a helper method.

KatBrandt: Also, remember that it is Ruby convention that any method that returns a boolean, should end in a `?`.

spec/facility_spec.rb
```diff
@@ -23,4 +34,105 @@
       expect(@facility.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
     end
   end
+
+# VEHICLE REGISTRATION
+  describe 'regular vehicle registration' do
+    it 'can register a regular vehicle' do
```
KatBrandt, line 40: So this is testing that it CAN register a vehicle, but we also want to test the conditions when it can't register a vehicle. So it would be good to have another test where the service is not added to the facility and you expect a different return value.

spec/facility_spec.rb
```diff
@@ -23,4 +34,105 @@
...
+      expect(@facility_1.administer_written_test?(@registrant_1)).to be true
+      expect(@registrant_1.license_data).to eq({:written => true, :license => false, :renewed => false})
+      
+      expect(@registrant_2.license_data).to eq({:written => false, :license => false, :renewed => false})
+      expect(@registrant_2.age).to eq(16)
+      expect(@registrant_2.permit?).to be false
+      expect(@facility_1.administer_written_test?(@registrant_2)).to be false
+      @registrant_2.earn_permit
+      expect(@facility_1.administer_written_test?(@registrant_2)).to be true
+      expect(@registrant_2.license_data).to eq({:written => true, :license => false, :renewed => false})
```
KatBrandt, line 99: Good job testing for the different conditions that need to be met in order to get the written test. Feel free to add more it blocks for each condition that is being tested, that will help with reading tests.

spec/facility_spec.rb
```diff
@@ -23,4 +34,105 @@
...
+      @facility_1.administer_written_test?(@registrant_1)
+      @facility_1.administer_road_test?(@registrant_1)
+      expect(@registrant_1.license_data).to eq({:written => true, :license => true, :renewed => false})
+    
+      @facility_1.add_service('Renew License')
+      expect(@facility_1.services).to eq ["Written Test", "Road Test", "Renew License"]
+      expect(@facility_1.renew_drivers_license?(@registrant_1)).to be true
+      expect(@registrant_1.license_data).to eq({:written=>true, :license=>true, :renewed=>true})
+    end
+  end
```
KatBrandt, line 136: So above here are integration tests because they're testing methods that are using helper methods. We'll also want to write unit tests for those helper methods you created to ensure they work as expected in isolation.

lib/facility_factory.rb
```diff
@@ -0,0 +1,44 @@
...
+        facility_info[:name] = dmv_info[:name]
+        facility_info[:address] = dmv_info[[:address1,:city, :state, :zipcode]]
+        facility_info[:phone] = dmv_info[:phone]
+        facility_info[:services] = dmv_info[] if dmv_info[nil]
+        facility_info[:registered_vehicles] = dmv_info[] if dmv_info[nil]
+        facility_info[:collected_fees] = dmv_info[] if dmv_info[nil]
+        Facility.new(facility_info)
+      end
+    end
+  end
```
KatBrandt, line 43: Overall, the logic of this looks really good. Nice job getting that api data into your app as objects. You could definitely have made some helper methods here though to encapsulate each of the if/else flows logic, that would help this be even more readable.

lib/facility_factory.rb
```diff
@@ -0,0 +1,44 @@
+class FacilityFactory
+
+  def initialize
+  end
+
+  def create_facility(dmv_info)
+    if dmv_info == @co_dmv_office_locations
```
KatBrandt, line 7: I'm unclear on what we are comparing dmv_info to here. There are no instance variables defined in this class, so the value of `@co_dmv_office_locations` is actually `nil` here. That means all of your data is running through your `else` logic flow. Your tests did not catch this because they're not specific enough, but if you put a binding.pry in your tests for this method, you'll see that you are getting back Facility objects that have no data filled in

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    @mo_dmv_office_locations = DmvDataService.new.mo_dmv_office_locations
+    factory = FacilityFactory.new
+
+    output1 = factory.create_facility(@co_dmv_office_locations)
+    output2 = factory.create_facility(@ny_dmv_office_locations)
+    output3 = factory.create_facility(@mo_dmv_office_locations)
+
+    expect(output1).to be_an(Array)
+    expect(output1.first).to be_a(Facility)
+
```
KatBrandt, line 21: To dig further into testing here, we could test that output1.first.address is a String, or check that it's not nil, and we could do that for all of the attributes that we expect to be set after this method runs.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,25 @@
...
+  
+
+  end
+
+end
+
+# It looks to me like the DmvDataService is going to be getting you a bunch of data
+#  from an API, and your job is to set up the VehicleFactory to take in that data
+#  and use it to instantiate Vehicle objects with it. I would probably start by prying into 
+# the DmvDataService near the end and seeing what format the data is that it's returning
```
KatBrandt, line 25: Yeah, great idea! And that's exactly what we're doing. Your create_vehicle method here looks good. Note: if nothing is in the initialize method, you do not need to include it.

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    expect(factory).to be_an_instance_of(VehicleFactory)
+  end
+
+  it "can create vehicle objects from an external database" do
+    @wa_ev_registrations = DmvDataService.new.wa_ev_registrations
+    factory = VehicleFactory.new
+    output = factory.create_vehicles(@wa_ev_registrations)
+    expect(output).to be_an(Array)
+    expect(output.first).to be_a(Vehicle)
+  end
```
KatBrandt, line 15: Again, we can go a little deeper here and check that all the attributes we expect to be filled in are not nil and are the correct data_type.

KatBrandt (commented):
Overall, you're doing a good job and this project looks good. Your tests are on the right track, we just want to make sure we are testing every method that we create for a class, including attr_readers and helper methods. Please review the feedback I provided here and let me know if you have any questions.

PR #492, opened 2024-07-11
Fixed all failures for part one of iteration 1

MDelarosa1993:


lib/Dmv_Facility_Factory.rb
```diff
@@ -0,0 +1,35 @@
+class DmvFacilityFactory
```
KatBrandt, line 1: Remember that our naming convention for files names is snake case, so this file should be `dmv_facility_factory.rb`

lib/Dmv_Facility_Factory.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def new_york_facilities(new_york_facilities)
+      facilities = new_york_facilities.map do |new_york_facility|
+      new_york_facility.default = "N/A"
+      name = new_york_facility[:office_name]
+      address = "#{new_york_facility[:street_address_line_1]}, #{new_york_facility[:city]}, #{new_york_facility[:state]}, #{new_york_facility[:zip_code]}"
+      phone = new_york_facility[:public_phone_number]
+      Facility.new({name: name, address: address, phone: phone})  
+    end
+    facilities
```
KatBrandt, line 22: Yup, nice work with this. No need to store your .map's return value in a variable though since you're not doing anything else to that variable afterwards.

lib/Dmv_Facility_Factory.rb
```diff
@@ -0,0 +1,35 @@
...
+  #     address = "#{co_dmv_office_location[:address_li]}, #{co_dmv_office_location[:address__1]}, #{co_dmv_office_location[:city]}, #{co_dmv_office_location[:state]} #{co_dmv_office_location[:zip]}"
+  #     phone = co_dmv_office_location[:phone]
+  #     Facility.new({name: name, address: address, phone: phone})
+  #   end
+  #  facilities
+  # end
+
+  def new_york_facilities(new_york_facilities)
+      facilities = new_york_facilities.map do |new_york_facility|
+      new_york_facility.default = "N/A"
```
KatBrandt, line 16: Remember to indent your code block, everything between the do/end should be indented.

lib/facility.rb
```diff
@@ -1,14 +1,67 @@
...
+    elsif vehicle.electric_vehicle?
+      @collected_fees += 200
+      vehicle.plate_type = :ev
+    else
+      @collected_fees += 100
+      vehicle.plate_type = :regular
+    end
+
+    @collected_fees
+
```
KatBrandt, line 38: This method is looking good. You accounted for almost all of the side effects we wanted with the register_vehicle method. The one part that is missing it to set the vehicle's registration date. 

We'll also want to start seeing if we can break this method up into smaller methods, so that each method is responsible for one piece of functionality.

lib/facility.rb
```diff
@@ -1,14 +1,67 @@
...
+      @collected_fees += 100
+      vehicle.plate_type = :regular
+    end
+
+    @collected_fees
+    
+  end
+  
+  def administer_written_test(registrant)
+    if @services.include?('Written Test') && registrant.permit? && registrant.age >= 16
```
KatBrandt, line 42: Great job with your logical operators in this statement!

spec/facility_spec.rb
```diff
@@ -1,26 +1,136 @@
...
+        expect(@cruz.registration_date).to eq(nil)
+        expect(@cruz.plate_type).to eq(:regular)
+        expect(@facility_1.registered_vehicles).to include(@cruz)
+        expect(@facility_1.collected_fees).to eq(100)
+      end
+
+      it 'registers plate_type :antique' do 
+        @facility_1.add_service('Vehicle Registration')
+        @facility_1.register_vehicle(@camaro)
+        expect(@camaro.registration_date).to eq(nil)
```
KatBrandt, line 51: We actually want this registration_date to NOT be nil once the vehicle is registered.

spec/facility_spec.rb
```diff
@@ -1,26 +1,136 @@
...
+        expect(@facility_1.renew_drivers_license(@registrant_1)).to be true
+        expect(@facility_1.renew_drivers_license(@registrant_2)).to be true
+      end
+
+      it 'does not renew license' do 
+        @registrant_3 = Registrant.new('Tucker', 15)
+        expect(@facility_1.renew_drivers_license(@registrant_3)).to eq false
+        expect(@registrant_3.license_data[:renewed]).to be false
+      end
+    end
```
KatBrandt, line 134: Good job with these tests and testing all the different conditions .

lib/registrant.rb
```diff
@@ -0,0 +1,27 @@
+class Registrant 
+  attr_reader :name, :age, :permit, :license_data
```

lib/vehicle.rb
```diff
@@ -6,13 +6,17 @@ class Vehicle
               :make,
               :model,
               :engine
+              
+attr_accessor :plate_type, :registration_date
 
-  def initialize(vehicle_details)
+def initialize(vehicle_details)
```
KatBrandt, line 12: indent all methods inside the class

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,16 @@
...
+      model = wa_ev_registration[:model]
+      vin = wa_ev_registration[:vin_1_10]
+      year = wa_ev_registration[:model_year]
+      engine = :ev
+      
+      Vehicle.new(make: make, model: model, vin: vin, year: year, engine: engine)
+    end
+    vehicles
+  end
+end
```
KatBrandt, line 16: Overall this looks good. .map will return the transformed array so no need to store it in a variable.

KatBrandt (commented):
Overall, this looks good. Some indentation issues and missing one piece of logic with the registration date. Tests are looking good and good job with enumerables and hashes.

PR #499, opened 2024-07-16
Candice Cirbo

CCirbo:
DMV, all tests were passing, now only Colorado is not passing due to the problem with the co dmv office locations.


lib/dmv_facility_factory.rb
```diff
@@ -0,0 +1,49 @@
...
+        state = bulk_facilities.first[:state]
+        if state == "CO"
+            build_co_facilities(bulk_facilities)
+        elsif state == "MO"
+            build_mo_facilities(bulk_facilities)
+        elsif state == "NY"
+            build_ny_facilities(bulk_facilities)
+        else
+            []
+        end
```
KatBrandt, line 13: I like how you put the logic into different helper methods in this control flow.

lib/dmv_facility_factory.rb
```diff
@@ -0,0 +1,49 @@
...
+    def build_mo_facilities(bulk_facilities)
+        bulk_facilities.map do |facility|
+            facility_information = {
+                name: facility[:name],  
+                address: "#{facility[:address1]} #{facility[:city]} #{facility[:state]} #{facility[:zipcode]}",
+                phone: facility[:phone]
+            }
+            Facility.new(facility_information)
+        end
+    end
```
KatBrandt, line 48: yeah, these all look good Candice! Nice job working with .map, hashes, and creating new Facility objects.

lib/facility.rb
```diff
@@ -1,14 +1,60 @@
...
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    return nil unless @services.include?("Vehicle Registration")
+    vehicle.assign_registration_date
+    vehicle.assign_plate_type
```
KatBrandt, line 25: I really like how you created helper methods in the Vehicle class that will change the value of some of the Vehicle attributes.

lib/facility.rb
```diff
@@ -1,14 +1,60 @@
...
+    registration_fee(vehicle)
+    @registered_vehicles << vehicle
+    @registered_vehicles
+  end
 end
+
+  def registration_fee(vehicle)
```
KatBrandt, line 32: Remember that method names should be verbs so maybe something like `calculate_reg_fee`

lib/facility.rb
```diff
@@ -1,14 +1,60 @@
...
+    if vehicle.antique?
+    @collected_fees += 25
```
KatBrandt, line 34: Remember to indent the code under each control flow. 

Great job using your existing methods of vehicle.antique? and vehicle.electric_vehicle?

lib/facility.rb
```diff
@@ -1,14 +1,60 @@
...
+    elsif vehicle.electric_vehicle?
+    @collected_fees += 200
+    else
+    @collected_fees += 100
```
KatBrandt, line 38: we're missing an `end` for the if statement.

lib/facility.rb
```diff
@@ -1,14 +1,60 @@
...
+    return false unless @services.include?('Road Test')
+    return false unless registrant.license_data[:written] == true
+    registrant.license_data[:license] = true
+  end
+
+  def renew_drivers_license(registrant)
+    return false unless @services.include?('Renew License')
+    return false unless registrant.license_data[:written] == true && registrant.license_data[:license] == true
+    registrant.license_data[:renewed] = true
+  end
```
KatBrandt, line 57: Good job with these three methods. Because we're checking three conditions though, I would rather just have all three conditions being checked at the same time instead of checking one condition, then the other two. Let's just do it all in one go. so line 42 could like like this: `if @services.include?('Written Test') && registrant.age >= 16 && registrant.permit?` then just put line 44 inside the if block.

spec/facility_spec.rb
```diff
@@ -22,5 +35,163 @@
...
+      @facility_1.add_service('Vehicle Registration')
+      expect(@facility_1.services).to eq(['New Drivers License', 'Vehicle Registration'])
+
+      @facility_2.add_service('New Drivers License')
+      @facility_2.add_service('Renew Drivers License')
+      expect(@facility_2.services).to eq(['New Drivers License','Renew Drivers License'])
+    end
+  end
+
+  describe '#vehicles services' do
```
KatBrandt, line 50: I like that you created a describe block to house all the it blocks that describe the behavior of the register_vehicle method. I would have put that as the describe block though: `describe "#register_vehicle" do ` that way we know all those it blocks pertain to that method.

spec/facility_spec.rb
```diff
@@ -22,5 +35,163 @@
...
+      @facility_1.administer_road_test(@registrant_2)
+
+      expect(@facility_1.renew_drivers_license(@registrant_2)).to eq(true)
+      expect(@registrant_2.license_data).to eq({:written=>true, :license=>true, :renewed=>true})
+    end
+    
+    it 'will not Renew License if registrant is missing the Written or Road Test' do
+      expect(@facility_1.renew_drivers_license(@registrant_3)).to eq(false)
+      expect(@registrant_3.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+    end
```
KatBrandt, line 193: Great job testing all the condition combinations for these methods!

lib/registrant.rb
```diff
@@ -0,0 +1,26 @@
+class Registrant
+    attr_reader :name,
+                :age,
+                :permit,
```

lib/vehicle.rb
```diff
@@ -22,4 +26,19 @@ def antique?
...
+
+  def assign_registration_date
+    @registration_date = Date.today
+  end
+
+  def assign_plate_type
+    if antique?
+      @plate_type = :antique
+    elsif
+      electric_vehicle?
```
KatBrandt, line 38: The elsif condition needs to be on the same line as the elsif

lib/vehicle.rb
```diff
@@ -22,4 +26,19 @@ def antique?
...
+      @plate_type = :ev
+    else
+      @plate_type = :regular
+    end
+  end
```
KatBrandt, line 43: I really like that you added these helper methods to this class. Make sure when you do that though, that you are testing these methods in the vehicle_spec file. You are testing them within register_vehicle, that's an integration test. We also want to test them in isolation as a unit test.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,14 @@
...
+         Vehicle.new(
+                vin: vehicle_details[:vin_1_10],
+                year: vehicle_details[:model_year],
+                make: vehicle_details[:make],
+                model: vehicle_details[:model],
+                engine: :ev
+            )
+        end
+    end
+end
```
KatBrandt, line 14: Great!

KatBrandt (commented):
Overall, really great job!!!!! Your tests are looking good, code looks good too. Seems like you are understanding how to work with Classes and objects, hashes, and enumerables. Nice work!

PR #500, opened 2024-07-16
Joe Chirchirillo

jchirch:


lib/dmv.rb
```diff
@@ -9,8 +11,8 @@ def add_facility(facility)
   end
 
   def facilities_offering_service(service)
-    @facilities.find do |facility|
+    @facilities.find_all do |facility|
```
KatBrandt, line 14: Good choice with this find_all enumerable.

lib/facility.rb
```diff
@@ -1,14 +1,47 @@
...
   end
-end
+
+  def register_vehicle(vehicle)
+    return unless @services.include?('Vehicle Registration')
+      vehicle.set_registration_date
+      vehicle.set_plate_type
+      @collected_fees += vehicle.plate_fees
+      @registered_vehicles << vehicle
+  end
```
KatBrandt, line 28: Indentation is off here. You do not have any code blocks so all this code inside the method should be at the same indentation.

KatBrandt: I really like how you constructed this method! You have the guard clause on line 23 that will check the condition, then I like that you created methods in Vehicle to handle the setting of the plate_type and registration_date attributes. That's a great separation of responsibility. I think the method of `plate_fees` could have gone in this class though, because that seems like something the Facility would set, not the Vehicle. So line 26 could have looked like this: `@collected_fees += calculate_plate_fees` and `calculate_plate_fees` would have been a method here in the Facility class.

lib/facility.rb
```diff
@@ -1,14 +1,47 @@
...
+    return false unless @services.include?('Road Test')
+    return false unless administer_written_test(registrant)
+    registrant.pass_road
+  end
+  
+  def renew_drivers_license(registrant)
+    return false unless @services.include?('Renew License')
+    return false unless registrant.license? == true && administer_road_test(registrant)
+    registrant.renew_license
+  end
```
KatBrandt, line 46: With these three methods, I like that you created a method in the Registrant class that handles changing the Registrant's attributes. I like the responsibility for that being in that class. 
For your conditional checks though, I am not sure breaking them up like this makes them more readable. I would have just used a single if statement for these ones since you were checking 3 different things, so it would like like this: `if @services.include?('Written Test') && registrant.permit? && registrant.age >=16`

On line 44, part of your conditional is calling the `administer_road_test` method, but that will actually execute that method which is not quite what we want. We just want to check if the registrant has met those conditions so you could check the registrants attributes.

spec/facility_spec.rb
```diff
@@ -1,26 +1,206 @@
...
-      @facility.add_service('Renew Drivers License')
-      @facility.add_service('Vehicle Registration')
-      expect(@facility.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
+      expect(@facility_1.services).to eq([])
+
+      @facility_1.add_service('New Drivers License')
+      @facility_1.add_service('Renew Drivers License')
+      @facility_1.add_service('Vehicle Registration')
+
+      expect(@facility_1.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
```
KatBrandt, line 38: This is a good test because you're checking the state of .services both before and after running the add_service method, so we can see the change of that command method. Nice work.

spec/facility_spec.rb
```diff
@@ -1,26 +1,206 @@
...
+    it 'can access list of registered vehicles' do
+      expect(@facility_1.registered_vehicles).to eq([])
+    end
+
+    it 'can register a vehicle' do
+      @facility_1.add_service('Vehicle Registration')
+      @facility_1.register_vehicle(@cruz)
+
+      expect(@facility_1.services).to eq(['Vehicle Registration'])
+      expect(@facility_1.registered_vehicles[0]).to be_an_instance_of(Vehicle)
```
KatBrandt, line 52: don't need this line. The assertion below it makes sure we have the Vehicle instance.

spec/facility_spec.rb
```diff
@@ -1,26 +1,206 @@
...
+
+    it 'can set a registration date' do
+      @facility_1.add_service('Vehicle Registration')
+      @facility_1.register_vehicle(@cruz)
+
+      expect(@cruz.registration_date).to eq(Date.today)
+    end
+  end
+
+  describe '#payment' do
```
KatBrandt, line 71: These methods are still part of the side effects of the #register_vehicle method, so I'd want these tests to be in the describe block above for the #register_vehicle method. Test all side effects of a given method inside that describe block.

spec/facility_spec.rb
```diff
@@ -1,26 +1,206 @@
...
+    it 'updates license data when passes administered written test' do
+      @facility_1.add_service('Written Test')
+
+      expect(@registrant_1.license_data[:written]).to be false
+
+      @facility_1.administer_written_test(@registrant_1)
+
+      expect(@registrant_1.license_data[:written]).to be true
+    end
+  end
```
KatBrandt, line 121: Good job testing all the different conditions for these methods.

lib/facility_factory.rb
```diff
@@ -0,0 +1,43 @@
...
+            location_data = {}
+            location.default = "N/A"
+
+            location_data[:name] = location[:name]
+            location_data[:address] = "#{location[:address1]} #{location[:city]} #{location[:state]} #{location[:zipcode]}"
+            location_data[:phone] = location[:phone]
+
+            Facility.new(location_data)
+        end
+    end
```
KatBrandt, line 42: Nice job with your hash work in taking this data and making new Facility objects!

lib/registrant.rb
```diff
@@ -0,0 +1,39 @@
+class Registrant
+
+    attr_reader :name, :age, :permit, :license_data
```
KatBrandt, line 3: Careful here, you have exposed a method of `.permit` that reads back the value of the `@permit` attribute, but you have another method `.permit?` that does that same thing. We only want one, the `.permit?` one so we don't want `.permit` in the attr_reader.

lib/registrant.rb
```diff
@@ -0,0 +1,39 @@
...
+        @permit
+    end
+
+    def earn_permit
+        @permit = true
+    end
+
+    def license?
+        @license_data[:license]
+    end
```
KatBrandt, line 26: I think this method. Could also create more for written and renewed so you could use those in the Facility methods.

lib/registrant.rb
```diff
@@ -0,0 +1,39 @@
...
+        @license_data[:written] = true
+    end
+
+    def pass_road
+        @license_data[:license] = true
+    end
+
+    def renew_license
+        @license_data[:renewed] = true
+    end
```
KatBrandt, line 38: Like I said, I like that you created these methods in this class. You made it the Registrant's responsibility to change it's own attributes.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+            expect(@registrant_2.earn_permit).to be true
+        end
+    end
+    describe '#license data' do
+        it 'can access license data' do
+            expect(@registrant_1.license_data[:written]).to be false
+            expect(@registrant_1.license_data[:license]).to be false
+            expect(@registrant_1.license_data[:license]).to be false
+        end
+    end
```
KatBrandt, line 40: We want to make sure that any methods we've written in a class, has a corresponding test. Looks like you did not test all your additional methods you created in the Registrant class.

lib/vehicle.rb
```diff
@@ -22,4 +26,28 @@ def antique?
...
+  end
+
+  def set_plate_type
+    if antique?
+      @plate_type = :antique
+    elsif electric_vehicle?
+      @plate_type = :ev
+    else
+      @plate_type = :regular
+    end
```
KatBrandt, line 41: Great job with your control flow.

lib/vehicle.rb
```diff
@@ -22,4 +26,28 @@ def antique?
...
+  end
+
+  def plate_fees
+    if @plate_type == :antique
```
KatBrandt, line 45: Careful here, you'd probably want to use your methods you have in this class of `.antique?` and `.electric_vehicle?` since you already have those written.

spec/vehicle_spec.rb
```diff
@@ -33,4 +34,16 @@
...
+  describe '#plate type' do
+    it 'can assign correct plate type' do
+      @cruz.set_plate_type
+      @bolt.set_plate_type
+      @camaro.set_plate_type
+      
+      expect(@cruz.plate_type).to eq(:regular)
+      expect(@bolt.plate_type).to eq(:ev)
+      expect(@camaro.plate_type).to eq(:antique)
+    end
```
KatBrandt, line 47: Good job testing this method.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,18 @@
...
+            vehicle_data = {}
+            registration.default = "N/A"
+           
+            vehicle_data[:vin] = registration[:vin_1_10]
+            vehicle_data[:year] = registration[:model_year]
+            vehicle_data[:make] = registration[:make]
+            vehicle_data[:model] = registration[:model]
+            vehicle_data[:engine] = :ev
+            Vehicle.new(vehicle_data)
+        end
```
KatBrandt, line 16: Nice work with this method and good use of .map. 
On line 2, since your initialize method isn't doing anything, you don't actually need to include it. :)

KatBrandt (commented):
Overall, really great work! Your testing is looking good, just make sure we're testing all methods we add to our classes. Really good job creating helper methods for changing attributes in a class.

lib/facility.rb
jchirch: Originally I did have the conditionals on lines 43 and 44 as `if @services.include?('Written Test') && registrant.permit? && registrant.age >=16` ! I refactored them because the code line was really long. Single statement does make more sense. Lesson learned!

PR #501, opened 2024-07-16
Jim Macur

jimmacur:


lib/facility.rb
```diff
@@ -1,14 +1,76 @@
...
+    @registered_vehicles = []
+    @collected_fees = 0
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def has_service?(service)
```
KatBrandt, line 22: Nice helper method!!! I'm hoping I find a test for this helper method in the spec file.

lib/facility.rb
```diff
@@ -1,14 +1,76 @@
...
+    @services.include?(service)
+   end
+
+  def allow_service(service)
```
KatBrandt, line 26: I'm not sure this method is necessary since you created the `has_service?` method to check, that method will return a true/false that you can use in conditionals.

lib/facility.rb
```diff
@@ -1,14 +1,76 @@
...
+    if has_service?(service)
+      yield
+    else
+      false
+    end
+  end
+
+  def register_vehicle(vehicle)
+    if
```
KatBrandt, line 35: our `if` needs a condition to know whether it should execute the rest of the code.

lib/facility.rb
```diff
@@ -1,14 +1,76 @@
...
+    else
+      false
+    end
+  end
+
+  def register_vehicle(vehicle)
+    if 
+      @registered_vehicles << vehicle
+      vehicle.set_date
+      vehicle.give_plate
```
KatBrandt, line 38: I like that you created these two helper methods in the Vehicle class to handle the changing of the attributes.

lib/facility.rb
```diff
@@ -1,14 +1,76 @@
...
+      elsif vehicle.plate_type == :antique
+        @collected_fees += 25
+      else
+        @collected_fees += 100
+      end
+      return [vehicle]
+    else 
+      return false
+    end
+  end
```
KatBrandt, line 50: Good job with the logic here. Moving forward, we're going to want to start practicing SRP (Single Responsibility PRinciple) where each method is only responsible for doing one thing. This method is doing a lot. :) Creating more helper methods to move this logic into is how we solve that.

lib/facility.rb
```diff
@@ -1,14 +1,76 @@
...
+
+  def administer_written_test(registrant)
+    return false unless registrant.age > 15 && registrant.permit?
+
+    allow_service('Written Test') do
```
KatBrandt, line 55: I'd rather just have you check this condition on line 53 with an if statement. So line 53 would say: `if registrant.age > 15 && registrant.permit? && has_service("Written Test")` and then put in that if block the logic of what to do if all those conditions are true. Then we only have to read one line to know what's going on, keeping all the conditions in one place.

spec/facility_spec.rb
```diff
@@ -17,10 +19,190 @@
...
     end
   end
+
+  describe '#has service?' do
+    it 'can check if a facility has a service' do
+      expect(@facility.has_service?('Vehicle Registration')).to eq(false)
+
+      @facility.add_service('Vehicle Registration')
+
+      expect(@facility.has_service?('Vehicle Registration')).to eq(true)
```
KatBrandt, line 37: YAY! Nice job adding a test for your helper method.

spec/facility_spec.rb
```diff
@@ -17,10 +19,190 @@
...
+  end
+
+  describe '#allow_service' do
+    it 'doesnt allow a service that hasnt been added' do
+      expect(@facility.allow_service('Vehicle Registration'){'Registered'}).to eq(false)     
+    end
+  end
+
+  describe '#registered vehicles' do
+    it 'can list vehicles that have been registered at a facility' do
```
KatBrandt, line 58: Based on your it statement, I would think this test is holding assertions showing us adding vehicles. Maybe instead this it block could say `it "starts with no vehicles registered"`. We want to be specific and explicit about what the it block is demonstrating.

spec/facility_spec.rb
```diff
@@ -17,10 +19,190 @@
...
+    end
+  end
+
+  describe '#registered vehicles' do
+    it 'can add to registered vehicles array' do
+      @facility.add_service('Vehicle Registration')
+      @cruz = Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice} )
+      @facility.register_vehicle(@cruz)
+
+      expect(@facility.registered_vehicles).to eq([@cruz])
```
KatBrandt, line 84: This is the same test as the test on line 70.

spec/facility_spec.rb
```diff
@@ -17,10 +19,190 @@
...
+    end
+  end
+
+  describe '#collect fees' do
+    it 'can collect fees upon registration' do
+      @cruz = Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice} )
+      @facility.add_service('Vehicle Registration')
+      @facility.register_vehicle(@cruz)
+
+      expect(@facility.collected_fees).to eq(100)
```
KatBrandt, line 122: Yes, this is a good start, then we'd also want to register another vehicle and test that the collected fees attribute incremented how we expect.

spec/facility_spec.rb
```diff
@@ -17,10 +19,190 @@
...
+      @facility.add_service('Written Test')
+      @facility.add_service('Road Test')
+      @facility.add_service('Renew License')
+      @facility.administer_written_test(registrant_2)
+      @facility.administer_road_test(registrant_2)
+
+      expect(@facility.renew_drivers_license(registrant_2)).to eq(true)
+      expect(registrant_2.license_data[:license]).to eq(true)
+    end
+  end
```
KatBrandt, line 206: Good job testing all these conditions for these methods.

lib/facility_factory.rb
```diff
@@ -0,0 +1,25 @@
...
+    def create_mo_facility(dmv_locations)
+        dmv_locations.map do |location|
+          Facility.new(
+            name: location[:name],
+            address: "#{location[:address1]}, #{location[:city]}, #{location[:state]}, #{location[:zipcode]}",
+            phone: location[:phone]
+          )
+        end
+    end
+
```
KatBrandt, line 24: Nice job with this class. Good use of using .map to instantiate new Facility instances. One note, when your initialize method is empty, you don't need to include it at all. :)

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+            expect(facility[0].name).to be_an_instance_of(String)
+            expect(facility[0].name).not_to eq("")
+
+            expect(facility[0].address).to be_an_instance_of(String)
+            expect(facility[0].address).not_to eq("")
+            
+            expect(facility[0].phone).to be_an_instance_of(String)
+            expect(facility[0].phone).not_to eq("")
+        end
+    end
```
KatBrandt, line 41: Nice job with this testing.

lib/registrant.rb
```diff
@@ -0,0 +1,21 @@
+class Registrant 
+    attr_reader :name, :age, :permit
```

lib/vehicle.rb
```diff
@@ -22,4 +26,19 @@ def antique?
   def electric_vehicle?
     @engine == :ev
   end
+
+  def set_date
+    @registration_date = Date.today
+  end
+
+  def give_plate
```
KatBrandt, line 34: I'd call this method something like `set_plate_type`

lib/vehicle.rb
```diff
@@ -22,4 +26,19 @@ def antique?
...
+    if engine == :ev
+      @plate_type = :ev
+    elsif @engine == :ice && Date.today.year - @year > 25
+      @plate_type = :antique
+    else
+      @plate_type = :regular
+    end
+  end
```
KatBrandt, line 42: Remember to look at the methods you already have to avoid repeating code. We had the methods `.electric_vehicle?` and `.antique?` that we could have used for the conditions of the flow control.

lib/vehicle.rb
KatBrandt: Also, I do not see tests for the set_date and give_plate methods in the vehicle_spec. Remember to always add tests for methods we add to classes.

KatBrandt (commented):
Overall, great work here. It seems like you understand how to use enumerables and hashes. We'll want to start working on having our it blocks testing single behaviors and work on making our methods smaller, having single responsibility.

lib/facility.rb
jimmacur: I agree and have seen other versions of projects that don't include it. Those look cleaner and are more simple. I was overcomplicating it.

lib/facility.rb
jimmacur: That makes sense. I thought I was creating efficiency by putting them together, but I see now that it complicates the readability and functionality. I will use more helper methods in the future.

lib/vehicle.rb
jimmacur: Yes. I missed that.

lib/vehicle.rb
jimmacur: This was another element that I saw in other people's projects that looks cleaner and more readable. I will keep my eye out for that in the future.

lib/facility.rb
jimmacur: I thought about that but it felt lie a really long line

KatBrandt: That's true, but you can actually stack them vertically to make it not a long line ;) So after each logical operator, put the next condition on the next line

lib/registrant.rb
jimmacur: Thank makes sense!

jimmacur: That

PR #502, opened 2024-07-16
Renee Messersmith

reneemes:


lib/facility.rb
```diff
@@ -1,14 +1,73 @@
...
   end
+
+  def register_vehicle(vehicle)
+    
+    if @services.include?('Vehicle Registration') == true
+      @registered_vehicles << vehicle
+      vehicle.set_registration_date
+      vehicle.set_plate_type
+
+      if vehicle.antique? == true
```
nicktu12, line 29: When you have a long method like this, think about how you can use helper methods to break up the logic. For example, you could have a helper method here that you call `process_antique_registration` to handle setting the plate type and collecting the right amount of fees for this type of car.

Using helper methods can help a lot when we have complex or nested conditionals, because those get hard to read. Extract some of that logic into a separate location and give it a readable name to indicate to another dev what is happening.

lib/facility.rb
```diff
@@ -1,14 +1,73 @@
...
+        @collected_fees += 25
+      elsif vehicle.electric_vehicle? == true
+        @collected_fees += 200
+      else
+        @collected_fees += 100
+      end
+    end
+
+    if @registered_vehicles.empty? == true
```
nicktu12, line 38: `@registered_vehicles.empty?` will evaluate to true, so checking that it is `== true` is redundant. Same could be said for your elsif before

lib/facility.rb
```diff
@@ -1,14 +1,73 @@
...
+    if @services.include?('Written Test') == true
+      if registrant.age >= 16 && registrant.permit == true
+        registrant.license_data[:written] = true
+      end
+    end
+    return registrant.license_data[:written]
+  end
+
+  def administer_road_test(registrant)
+    if @services.include?('Road Test') == true
```
nicktu12, line 56: Consider chaining together the conditional checks here with &&
```
if @services.include?('Renew License') == true && registrant.license_data[:license] == true
        registrant.license_data[:renewed] = true
      end

```

nicktu12, line 56: Consider chaining together the conditional checks here with &&
```
if @services.include?('Renew License') == true && registrant.license_data[:license] == true
        registrant.license_data[:renewed] = true
      end

```

nicktu12: Again, you dont need to explictly check that these values == true, as `include?` will evaluate to `true` or `false` inherinetly

spec/facility_catalog_spec.rb
```diff
@@ -0,0 +1,45 @@
+require 'spec_helper'
+
+RSpec.describe VehicleFactory do
+    before(:each) do
```
nicktu12, line 4: Great use of beforeEach!

lib/facility.rb
```diff
@@ -1,14 +1,73 @@
...
+    @collected_fees = 0
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+
```
KatBrandt, line 23: No white space here.

spec/facility_spec.rb
```diff
@@ -23,4 +32,81 @@
...
+      expect(@facility_1.registered_vehicles).to eq([])
+      @facility_1.add_service('Vehicle Registration')
+      expect(@facility_1.register_vehicle(@cruz)).to eq([@cruz])
+    end
+
+    it 'only registers a vehicle if the facility offers that service' do
+      expect(@facility_2.registered_vehicles).to eq([])
+      expect(@facility_2.register_vehicle(@bolt)).to eq(nil)
+      expect(@facility_2.registered_vehicles).to eq([])
+      expect(@facility_2.collected_fees).to eq(0)
```
KatBrandt, line 47: Great job testing that you can not register and none of these attributes are effected if that first condition isn't met.

spec/facility_spec.rb
```diff
@@ -23,4 +32,81 @@
...
+    end
+  end
+
+  describe '#administer_written_test' do
+    it 'offers a written test' do
+      expect(@facility_1.administer_written_test(@registrant_1)).to eq(false)
+      @facility_1.add_service('Written Test')
+      expect(@facility_1.administer_written_test(@registrant_1)).to eq(true)
+    end
+    it 'changes :written from flase to true when requirements are met' do
```
KatBrandt, line 70: white space between end and the next it block

spec/facility_spec.rb
```diff
@@ -23,4 +32,81 @@
...
+      expect(@facility_1.administer_road_test(@registrant_1)).to eq(true)
+      expect(@facility_1.renew_drivers_license(@registrant_1)).to eq(false)
+      @facility_1.add_service('Renew License')
+      expect(@facility_1.renew_drivers_license(@registrant_1)).to eq(true)
+    end
+    it 'does not renew if requirements are not met' do
+      @facility_1.add_service('Renew License')
+      expect(@facility_1.renew_drivers_license(@registrant_3)).to eq(false)
+    end
+  end
```
KatBrandt, line 110: good job testing for all these conditions for these methods!

lib/facility_catalog.rb
```diff
@@ -0,0 +1,38 @@
+class FacilityCatalog
+   
+    attr_reader :facility_catalog
+
+    def initialize
```
KatBrandt, line 5: No need to have an initialize method if it's not doing anything. :)

lib/facility_catalog.rb
```diff
@@ -0,0 +1,38 @@
...
+            end
+           
+            if office[:phone]
+                office_info[:phone] = office[:phone]
+            elsif office[:public_phone_number]
+                office_info[:phone] = office[:public_phone_number]
+            end
+            Facility.new(office_info)
+        end
+    end
```
KatBrandt, line 36: Great job here using .map and hashes to create these facilities.

lib/registrant.rb
```diff
@@ -0,0 +1,22 @@
+class Registrant
+
+    attr_reader :name,
+                :age,
+                :permit,
```

lib/vehicle.rb
```diff
@@ -22,4 +26,19 @@ def antique?
...
+  
+  def set_plate_type
+    if antique? == true
+      @plate_type = :antique
+    elsif electric_vehicle? == true
+      @plate_type = :ev
+    else
+      @plate_type = :regular
+    end
+  end
```
KatBrandt, line 42: I love these two methods you created here in the Vehicle class. Since we're changing a vehicle attribute, it makes sense that the class that holds those attributes should be in charge of changing those attributes.

lib/vehicle.rb
```diff
@@ -22,4 +26,19 @@ def antique?
   def electric_vehicle?
     @engine == :ev
   end
+
+  def set_registration_date
+    @registration_date = Date.today
+  end
+  
+  def set_plate_type
+    if antique? == true
```
KatBrandt, line 35: just `if antique?` since that will already return a boolean.

spec/vehicle_spec.rb
```diff
@@ -33,4 +34,22 @@
...
+    end
+  end
+
+  describe '#set_plate_type' do
+    it 'checks for antique and electric' do
+      expect(@cruz.set_plate_type).to eq(:regular)
+      expect(@bolt.set_plate_type).to eq(:ev)
+      expect(@camaro.set_plate_type).to eq(:antique)
+    end
+  end
```
KatBrandt, line 53: Good job testing the extra methods you created in the Vehicle class.

KatBrandt (commented):
Overall, really great work here!

PR #504, opened 2024-07-16
John Hill

jphill19:
Completed all the iterations and went above and beyond the scope of expectations by challenging myself.

Key Highlights:

- Strong Typing: Enhanced methods to accept only certain data types, reducing room for error.

- Custom Errors: Implemented custom errors to handle edge cases effectively, with corresponding tests.

Iteration 4 (Option 3):

- Option 3 Completed: Created iteration_4.rb to output analytics as outlined in Option 3.

- FacilityBuilder Adjustments: Modified methods to account for daily hours and holidays closed. Unable to fully test Colorado DMV API as it was unresponsive during testing, therefor I left the colorado method untouched.

- VehicleFactory Enhancements: Added a method to handle ny_registration vehicles, which the data coming from was API was for boats, so this was a bit tricky to tackle. I attempted what I think works best and created  corresponding tests.

I am open to feedback and look forward to it!

Thanks,
John


Iteration_4.rb
```diff
@@ -0,0 +1,66 @@
+require './lib/vehicle_factory'
+require './lib/dmv_data_service'
+require './lib/vehicle.rb'
+
+###### Option 3 ---> Analytics #######
```
KatBrandt, line 5: Remember, we want to practice OOP which means these methods could be encapsulated in a new class. Maybe a class called Analytics or something, then it should have a corresponding test file to test those methods. :)

Iteration_4.rb
```diff
@@ -0,0 +1,66 @@
...
+# Hash for years, make, model
+vehicle_make_count = Hash.new(0)
+vehicle_model_count = Hash.new(0)
+vehicle_years_count = Hash.new(0)
+
+#Hash for counties
+county_count = Hash.new(0)
+
+
+def max_count (data)
```
KatBrandt, line 22: No space between method name and opening parenthesis

Iteration_4.rb
```diff
@@ -0,0 +1,66 @@
...
+puts "\nThe most popular model regisred is: #{max_model[0]}, with a total of #{max_model[1]}/#{factory.vehicles_created.length}\n"
+puts "\nThe most popular make registered is #{max_make[0]}, with a total of #{max_make[1]}/#{factory.vehicles_created.length}\n"
+
+puts "\nVehicle By Years"
+vehicle_years_sorted.each do |key, value|
+  puts "\tYear: #{key}, Total: #{value}"
+end
+
+
+puts "\nMost popular county is #{most_popular_county[0]} with a total of #{most_popular_county[1]}/#{factory.vehicles_created.length}"
```
KatBrandt, line 66: Overall, great job with these analytics! You got lots of good practice with enumerables and hashes.

lib/facility.rb
```diff
@@ -1,14 +1,142 @@
...
+  end
+end
+
+class VehicleAlreadyRegisted < StandardError
+  attr_reader :message
+  def initialize(msg="This vehicle is already registered")
+    @message = msg
+    super(@message)
+  end
+end
```
KatBrandt, line 26: I like this, but I think it could help be more reusable if we had one CustomError class and then it can take in an argument of the message since the structure of these classes is the same. Regardless, great job playing with this!!!!

lib/facility.rb
```diff
@@ -1,14 +1,142 @@
...
+        @collected_fees += 200
+
+      elsif vehicle.antique?
+        vehicle.set_platetype(:antique)
+        @collected_fees += 25
+
+      else
+        vehicle.set_platetype(:regular)
+        @collected_fees += 100
+
```
KatBrandt, line 70: Too many white spaces here, remove white space on line 50, 58, 62, 66, and 70. You _could_ put a white space before an if block though, so a space above line 55.

lib/facility.rb
```diff
@@ -1,14 +1,142 @@
...
+      vehicle.set_registration_date
+
+      return @registered_vehicles << vehicle
+
+      # return "#{vehicle.model} has been registered"
+
+    else
+      return @registered_vehicles
+    end
+
```
KatBrandt, line 82: Good job with the logic here, but this method is doing a lot. How could we refactor this so that we follow SRP - Single Responsiblility Principle and this method is doing one thing? Could be that one thing is orchestrating the flow of data by calling other methods that are holding this logic. More helper methods. :)

lib/facility.rb
```diff
@@ -1,14 +1,142 @@
...
+        vehicle.set_platetype(:antique)
+        @collected_fees += 25
+
+      else
+        vehicle.set_platetype(:regular)
+        @collected_fees += 100
+
+      end
+
+      vehicle.set_registration_date
```
KatBrandt, line 73: I like that you put the logic for setting the reg date in the Vehicle class. The class that owns the attribute should be the one changing that attribute.

lib/facility.rb
```diff
@@ -1,14 +1,142 @@
...
+    else
+      return @registered_vehicles
+    end
+
+  end
+
+  def administer_written_test(registrant)
+    if @services.include?('Written Test')
+      raise TypeError, "registrant must be a registrant object" unless registrant.is_a?(Registrant)
+      raise WrittenTestAlreadyTakenError.new() unless registrant.license_data[:written] == false
```
KatBrandt, line 88: In Ruby, we do not use open parenthesis to invoke a method, so just `.new` not `.new()`

spec/facility_spec.rb
```diff
@@ -2,25 +2,258 @@
...
+      @facility_1.add_service('New Drivers License')
+      @facility_1.add_service('Renew Drivers License')
+      @facility_1.add_service('Vehicle Registration')
+      expect(@facility_1.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
+    end
+  end
+
+
+  describe '#register_vehicle' do
+
```
KatBrandt, line 39: Do not use white space between a describe and an it, or on line 45 between the last line and the end.

spec/facility_spec.rb
```diff
@@ -2,25 +2,258 @@
...
+    it "will only function if facility has the service to do so" do
+      expect(@facility_1.register_vehicle(@cruz)).to eq([])
+
+      @facility_1.add_service('Vehicle Registration')
+      expect(@facility_1.register_vehicle(@cruz)).to eq([@cruz])
```
KatBrandt, line 44: Good job testing the initial state, running the command method, then showing the next value of the attribute.

spec/facility_spec.rb
```diff
@@ -2,25 +2,258 @@
...
+
+      expect(@facility_1.collected_fees).to eq(25)
+      expect(@facility_1.registered_vehicles).to eq([@camaro])
+      expect(@camaro.registration_date).to be_an_instance_of(Date)
+      expect(@camaro.plate_type).to eq(:antique)
+
+
     end
+
+    it 'can appropriately register regular vehicle' do
```
KatBrandt, line 86: Good job testing each of the different kinds of vehicles.

spec/facility_spec.rb
```diff
@@ -2,25 +2,258 @@
...
+      expect(@facility_1.renew_drivers_license(@registrant_1)).to eq(true)
+
+      # Registrant now has renewed_license
+      expect(@registrant_1.license_data).to eq({:written=>true, :license=>true, :renewed=>true})
+
+    end
+    
+    # No edge case test for renewed_license as the assumption is that a user with a renewed_license can renew again
+
+  end
```
KatBrandt, line 255: Good job testing all the conditions for these methods!!!!!

lib/facility_builder.rb
```diff
@@ -0,0 +1,92 @@
+class FacilityBuilder
+
```
KatBrandt, line 2: No space on line 2, but DO want a space between the attr_readers and the def initialize

lib/facility_builder.rb
```diff
@@ -0,0 +1,92 @@
...
+  attr_reader :co_facilities, :ny_facilities, :mo_facilities
+  def initialize
+    @co_facilities = []
+    @ny_facilities = []
+    @mo_facilities = []
+  end
+
+
+  def create_co_facilities(data)
+
```
KatBrandt, line 12: No white space here

lib/facility_builder.rb
```diff
@@ -0,0 +1,92 @@
...
+      @co_facilities << new_facility
+
+    end
+    return @co_facilities
+  end
+
+  def create_ny_facilities(data)
+
+    @ny_facilities = data.map do |data|
```
KatBrandt, line 38: You used a .each for co, but a .map for ny. Probably just the .each would work for both since you're storing them in the instance variable

lib/facility_builder.rb
```diff
@@ -0,0 +1,92 @@
+class FacilityBuilder
+
+  attr_reader :co_facilities, :ny_facilities, :mo_facilities
+  def initialize
+    @co_facilities = []
+    @ny_facilities = []
+    @mo_facilities = []
```
KatBrandt, line 7: I like your choice of storing these as instance variables.

lib/registrant.rb
```diff
@@ -0,0 +1,32 @@
+#type error link: https://www.exceptionalcreatures.com/bestiary/TypeError.html
+
+class Registrant
+  attr_accessor :license_data
+  attr_reader :name, :age, :permit
```

lib/registrant.rb
```diff
@@ -0,0 +1,32 @@
...
+  def permit?
+    @permit
+  end
+
+  def earn_permit
+    if @permit
+      return @permit
+    else @permit
+      @permit = true
+    end
```
KatBrandt, line 28: If we just do `@permit = true` it will accomplish all this behavior, no need for an if statement here.

lib/vehicle.rb
```diff
@@ -1,18 +1,22 @@
 require 'date'
 
 class Vehicle
+  attr_accessor :registration_date
```
KatBrandt, line 4: No need for an attr_accessor here since you created a method of `set_registration_date` that handles changing this attribute.

lib/vehicle.rb
```diff
@@ -22,4 +26,13 @@ def antique?
...
     @engine == :ev
   end
+
+  def set_platetype(type)
+    raise TypeError, "plate_type must be a symbol" unless type.is_a?(Symbol)
+    @plate_type = type
+  end
+
+  def set_registration_date
+    @registration_date = Date.new(Date.today.year, Date.today.month, Date.today.day)
```
KatBrandt, line 36: Also could just do `@registration_date = Date.today` :) Nice job making a method for this attribute change!

KatBrandt (commented):
Overall, really great job! Nice work challenging yourself with extensions. It looks like you have a good grasp on enumerables and hashes. Testing looks good overall. Let's work on reducing white spaces (helps with readability) and work on SRP - using helper methods to ensure each method is doing one thing.

PR #505, opened 2024-07-16
Shane Galvin

Sgalvin36:
Followed all interaction patterns to create the registrant, vehicle_factory, and facility_factory classes. Provided some functionality with the ability to expand fairly easily.
Went into iteration 4 and added some analytic functionality for the vehicle registration. Got JavaScript registrant class working according to the accompanying test file.


lib/DMV_setup.rb
```diff
@@ -0,0 +1,50 @@
...
+                address_string[4] = value
+            elsif key.include?("address") && value.include?("Suite")
+                address_string[1] = value
+            elsif key.include?("address") && !(key.include?("id")) && !(value.include?("Suite"))
+                address_string[0] = value
+            end
+        end
+        address_string = address_string.compact
+        address_string.join(" ")
+    end
```
KatBrandt, line 49: Really great use of helper methods here to adhere to single repsonsibility for your methods.

lib/DMV_setup.rb
```diff
@@ -0,0 +1,50 @@
+class DMVSetup
```
KatBrandt, line 1: Remember that our convention for file names is snake case. So you might have wanted to call this class `DmvSetup` so the file could be called `dmv_setup.rb`

lib/facility.rb
```diff
@@ -1,14 +1,85 @@
...
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    if @services.include?("Vehicle Registration") && vehicle.registration_date == nil
+      @collected_fees += collect_fees(vehicle)
+      @registered_vehicles << vehicle
+      vehicle.registration_date = Time.new
+    end
```
KatBrandt, line 22: Overall this method looks good. To refactor it, one line 21, I would have pushed the responsibility of changing the vehicle's attribute into the Vehicle class, so creating a method in the Vehicle class called `set_registration_date` that encapsulates this logic.

lib/facility.rb
```diff
@@ -1,14 +1,85 @@
...
+  end
+
+  def collect_fees(vehicle)
```
KatBrandt, line 25: I like that you created a helper method for this logic. It could be argued though that this method is doing 2 things and we want to strive for SRP (Single Responsibility Principle), so could break this into two different methods, maybe the setting of the plate happens in the vehicle class.

lib/facility.rb
```diff
@@ -1,14 +1,85 @@
...
+  def administer_written_test(registrant)
+    if @services.include?("Written Test") && registrant.permit? && registrant.age >= 16
+      registrant.license_data[:written] = true
+    else
+      false
+    end
+  end
+
+  def administer_road_test(registrant)
+    if @services.include?("Road Test") && registrant.license_data[:written] == true
```
KatBrandt, line 47: Nice job checking all this logic. Technically, we don't need to do a comparison to true though for `registrant.license_data[:written]` since that statement will already evaluate to true or false.

lib/facility.rb
```diff
@@ -1,14 +1,85 @@
...
+    end
+  end
+
+  
+  def all_of_one_attribute(attribute)
+    @all_types = []
+    @registered_vehicles.each do |vehicle|
+      @all_types << (vehicle.instance_variable_get("@#{attribute}"))
+    end
+    @all_types
```
KatBrandt, line 68: I'm curious why you're setting ivars in this method. Is that necessary? If yes, then we want to establish the ivar in the initialize method, that way we always have all the ivars in one place so we can easily see what we have.

lib/facility.rb
```diff
@@ -1,14 +1,85 @@
...
+  end
+
+  def unique_types(attribute)
+    @unique_types = all_of_one_attribute(attribute)
+    @unique_types.uniq
```
KatBrandt, line 73: don't need another ivar here. You have the method to calculate these dynamically when the method is called.

lib/facility.rb
```diff
@@ -1,14 +1,85 @@
...
+
+  def quantity_unique_types(attribute)
+    quantities = {}
+    types = unique_types(attribute)
+    total_data = all_of_one_attribute(attribute)
+    types.map do |type|
+      quantities[type] = total_data.count(type)
+    end
+    quantities
+  end
```
KatBrandt, line 84: Really great use of helper methods, hashes, and enumerables.

spec/facility_spec.rb
```diff
@@ -2,25 +2,276 @@
...
+      expect(@cruz.registration_date).to be_an_instance_of Time
+    end
+
+    it 'adds vehicle to registered vehicles list' do
+      @facility_1.add_service('Vehicle Registration')
+      expect(@facility_1.registered_vehicles).to eq []
+
+      @facility_1.register_vehicle(@cruz)
+      expect(@facility_1.registered_vehicles).to include(Vehicle)
+      expect(@facility_1.registered_vehicles.count).to eq 1
```
KatBrandt, line 53: I'd rather us be more explicit with this assertion. The assertions you have are true, but they what if some other random Vehicle got inserted? We can rewrite the assertion to be `expect(@facility_1.registered_vehicles).to eq([@cruz])` and this ensures that the return value is an Array, has exactly one element, that element is the specific Vehicle instance that we registered.

spec/facility_spec.rb
```diff
@@ -2,25 +2,276 @@
...
+      expect(@facility_1.collected_fees).to eq 0
+
+      @facility_1.register_vehicle(@cruz)
+      expect(@facility_1.collected_fees).to eq 100
+
+      @facility_1.register_vehicle(@bolt)
+      expect(@facility_1.collected_fees).to eq 300
+
+      @facility_1.register_vehicle(@camaro)
+      expect(@facility_1.collected_fees).to eq 325
```
KatBrandt, line 72: Nice job testing that not only can we calculate one registered vehicle's fees, but it successfully adds those fees together.

spec/facility_spec.rb
```diff
@@ -2,25 +2,276 @@
...
+      expect(@facility_1.quantity_unique_types('year').count).to eq 14
+      p @facility_1.quantity_unique_types('year')
+    end
+
+    it 'can reveal other information for vehicle registration' do
+      expect(@facility_1.quantity_unique_types('model')).to be_an_instance_of Hash
+      expect(@facility_1.quantity_unique_types('model').count).to eq 67
+      expect(@facility_1.quantity_unique_types('make')).to be_an_instance_of Hash
+      expect(@facility_1.quantity_unique_types('make').count).to eq 27
     end
```
KatBrandt, line 275: Great job adding tests to your additional methods you created. 
Remember to try to test explicitly for the return value you're expecting.

lib/vehicle.rb
```diff
@@ -6,17 +6,22 @@ class Vehicle
               :make,
               :model,
               :engine
+              
+  attr_accessor :registration_date,
```
KatBrandt, line 10: You technically added additional methods with this attr_accessor which means there should be additional tests in the spec file to test that we can indeed change these attributes.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,34 @@
+class VehicleFactory
+    attr_accessor :vehicle_lot
```
KatBrandt, line 2: Do we need this to be an accessor? We want to not expose methods unless they're needed.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,34 @@
...
+          elsif key == :model || key.to_s.include?("body")
+            filtered_set[:model] = value
+          elsif key.to_s.include?("electric") || key == :engine
+            filtered_set[:engine] = :ev
+          elsif key.to_s.include?('fuel')
+            filtered_set[:engine] = value
+          end
+        end
+        filtered_set
+    end
```
KatBrandt, line 33: Nice job with all this logic. There are a lot of keys to check here. And to answer your question, there might be a way to further abstract all of this into one class that creates objects, but I'm not sure I would make that abstraction, could make everything more complex and harder to read.

KatBrandt (commented):
Really great work here!

PR #506, opened 2024-07-16
Finished DMV Project

SThompson05:
## Reflection
Before moving on to the next portion of this iteration, reflect on your process of learning a new codebase. Take notes on the following questions so that you can speak to these themes at the end of the project:

** Please add the below reflections to your Mod 1 Project Reflections gist within the Second Solo Project section. **

**Describe the steps you took to dig in to this code base. What was your process? If you don’t feel you had a process, define one that you might like to try next time.**

I was good for iteration 1 of this project, but I don't know what it was but I had a moment of panic starting iteration 2. I was totally clueless about what was being asked. I hung out in Study Hall quite a bit for this one which is a double edge blade because some people are so far ahead of you and you think you aren't getting it. However, people are more than willing to help you. Next time, I want to sign up for Student Support.

**What was hard about working with code you did not write?**

I couldn't for the life of me to get New York's DMV hours of operation to map. I used ChatGPT on it and still couldn't get it to work. It was frustrating.

**What was easier than you expected about jumping into an unfamiliar codebase? What made it easy? If nothing felt easy, what would’ve helped you feel more comfortable more quickly?**

Not being able to see and touch the data. It feels very incorporeal. We can't see it unless we use specific methods. It's hard for me to visualize what it is that we're doing until it's done. It'll probably click for me in the future. *fingers crossed*

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,12 @@
 # The DMV
 
 This is the starter repo for the BE Mod1 DMV project.
+
+Reflection
+Before moving on to the next portion of this iteration, reflect on your process of learning a new codebase. Take notes on the following questions so that you can speak to these themes at the end of the project:
+
+** Please add the below reflections to your Mod 1 Project Reflections gist within the Second Solo Project section. **
+
+Describe the steps you took to dig in to this code base. What was your process? If you don’t feel you had a process, define one that you might like to try next time.
+What was hard about working with code you did not write?
+What was easier than you expected about jumping in to an unfamiliar codebase? What made it easy? If nothing felt easy, what would’ve helped you feel more comfortable more quickly?
--- a/lib/dmv.rb
+++ b/lib/dmv.rb
@@ -1,5 +1,5 @@
 class Dmv
-
+  attr_reader :facilities
   def initialize
     @facilities = []
   end
--- a/lib/facility.rb
+++ b/lib/facility.rb
@@ -1,10 +1,10 @@
 class Facility
   attr_reader :name, :address, :phone, :services
 
-  def initialize(name, address, phone)
-    @name = name
-    @address = address
-    @phone = phone
+  def initialize(info)
+    @name = info[:name]
+    @address = info[:address]
+    @phone = info[:phone]
     @services = []
   end
 
--- a/lib/facility.rb
+++ b/lib/facility.rb
@@ -41,6 +41,14 @@ class Facility
     true
   end
 
+  def renew_drivers_license(registrant)
+    return false unless @services.include?('Renew License')
+    return false unless registrant.license_data[:license]
+
+    registrant.license_data[:renewed] = true
+    true
+  end
+
   private
 
   def update_plate_type_and_fees(vehicle)
--- a/lib/facility.rb
+++ b/lib/facility.rb
@@ -26,12 +26,20 @@ class Facility
 
   def administer_written_test(registrant)
     return false unless @services.include?('Written Test')
-    return false unless registrant.permit?
-  
+    return false unless registrant.permit? && registrant.age >=16
+
     registrant.license_data[:written] = true
     true
   end
   
+  def administer_road_test(registrant)
+    return false unless @services.include?('Road Test')
+    return false unless registrant.license_data[:written]
+    return false unless registrant.permit?
+
+    registrant.license_data[:license] = true
+    true
+  end
 
   private
 
--- a/lib/facility.rb
+++ b/lib/facility.rb
@@ -15,6 +15,33 @@ class Facility
     @services << service
   end
 
+  def register_vehicle(vehicle)
+    return unless @services.include?('Vehicle Registration')
+    return unless vehicle.is_a?(Vehicle)
+
+    update_plate_type_and_fees(vehicle)
+    set_registration_date(vehicle)
+    @registered_vehicles << vehicle
+  end
+
+  private
+
+  def update_plate_type_and_fees(vehicle)
+    if vehicle.antique?
+      vehicle.plate_type = :antique
+      @collected_fees += 25
+    elsif vehicle.electric_vehicle?
+      vehicle.plate_type = :ev
+      @collected_fees += 200
+    else
+      vehicle.plate_type = :regular
+      @collected_fees +=100
+    end
+  end
+
+  def set_registration_date(vehicle)
+    vehicle.registration_date = Date.today
+  end
 
 
 end
--- a/lib/facility.rb
+++ b/lib/facility.rb
@@ -24,6 +24,15 @@ class Facility
     @registered_vehicles << vehicle
   end
 
+  def administer_written_test(registrant)
+    return false unless @services.include?('Written Test')
+    return false unless registrant.permit?
+  
+    registrant.license_data[:written] = true
+    true
+  end
+  
+
   private
 
   def update_plate_type_and_fees(vehicle)
--- a/lib/facility.rb
+++ b/lib/facility.rb
@@ -1,14 +1,20 @@
 class Facility
-  attr_reader :name, :address, :phone, :services
+  attr_reader :name, :address, :phone, 
+              :services, :registered_vehicles, :collected_fees
 
   def initialize(info)
     @name = info[:name]
     @address = info[:address]
     @phone = info[:phone]
     @services = []
+    @registered_vehicles = []
+    @collected_fees = 0
   end
 
-  def add_services(service)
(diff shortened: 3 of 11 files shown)


README.md
```diff
@@ -1,3 +1,12 @@
 # The DMV
 
 This is the starter repo for the BE Mod1 DMV project.
+
```
KatBrandt, line 4: Thank you for this insightful reflection!

lib/dmv.rb
```diff
@@ -1,5 +1,5 @@
 class Dmv
-
+  attr_reader :facilities
   def initialize
     @facilities = []
```
KatBrandt, line 4: Were you able to make this bug pass? We needed to change line 12 of this class to a different enumerable to get this one working.

lib/facility.rb
```diff
@@ -1,14 +1,72 @@
...
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    return unless @services.include?('Vehicle Registration')
+    return unless vehicle.is_a?(Vehicle)
+
+    update_plate_type_and_fees(vehicle)
+    set_registration_date(vehicle)
```
KatBrandt, line 23: I do like that you created helper methods to move some of the logic out of this register_vehicle method. Makes it a lot more readable! 
I do think we could talk more about which class these helper methods should be in though. The class that is the owner of the attribute we're changing should be the class that is changing the attribute. For example, `set_registration_date`, we're changing a Vehicle attribute, so we could put that method in the Vehicle class and then on line 23 call `vehicle.set_registration_date`.

lib/facility.rb
```diff
@@ -1,14 +1,72 @@
...
+    return false unless @services.include?('Renew License')
+    return false unless registrant.license_data[:license]
+
+    registrant.license_data[:renewed] = true
+    true
+  end
+
+  private
+
+  def update_plate_type_and_fees(vehicle)
```
KatBrandt, line 54: This is good, however moving forward, we want to start thinking about Single Responsibility Principle (SRP) where every method should be doing one thing. The fact that you have an `and` in your method name is a great indicator that this should be split up into 2 methods.

lib/facility.rb
```diff
@@ -1,14 +1,72 @@
...
+    if vehicle.antique?
+      vehicle.plate_type = :antique
+      @collected_fees += 25
+    elsif vehicle.electric_vehicle?
+      vehicle.plate_type = :ev
+      @collected_fees += 200
+    else
+      vehicle.plate_type = :regular
+      @collected_fees +=100
+    end
```
KatBrandt, line 64: Good work with this control flow.

lib/facility.rb
```diff
@@ -1,14 +1,72 @@
...
+    return unless vehicle.is_a?(Vehicle)
+
+    update_plate_type_and_fees(vehicle)
+    set_registration_date(vehicle)
+    @registered_vehicles << vehicle
+  end
+
+  def administer_written_test(registrant)
+    return false unless @services.include?('Written Test')
+    return false unless registrant.permit? && registrant.age >=16
```
KatBrandt, line 29: For these 3 methods, since we're checking for all these conditions to be true, I think it's easier to read if we just put them all into one if statement instead of breaking them up into multiple guard clauses. So this would look like `if @services.include?("Written Test") && registrant.permit? && registrant.age >= 16` then put the logic inside the if statement. 

But, great job checking for all these conditions!

spec/facility_spec.rb
```diff
@@ -1,26 +1,174 @@
...
-      expect(@facility.services).to eq([])
-      @facility.add_service('New Drivers License')
-      @facility.add_service('Renew Drivers License')
-      @facility.add_service('Vehicle Registration')
-      expect(@facility.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
+      expect(@facility_1.services).to eq([])
+      @facility_1.add_service('New Drivers License')
+      @facility_1.add_service('Renew Drivers License')
+      @facility_1.add_service('Vehicle Registration')
+      expect(@facility_1.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
```
KatBrandt, line 35: This is a great test because it shows the state of the services attribute before and after running the command method of add_service.

spec/facility_spec.rb
```diff
@@ -1,26 +1,174 @@
...
+    end
+  end
+
+  describe 'registered_vehicles' do
```
KatBrandt, line 39: I like that you used a describe block to organize the it blocks that pertain to this method. 

Our it blocks we also want to practice SRP, so one it block for each behavior we're trying to test.

spec/facility_spec.rb
```diff
@@ -1,26 +1,174 @@
...
+  end
+
+  describe 'Written Test' do
+    it 'administers written test and updates license_data' do
+      expect(@registrant_1.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+      expect(@registrant_1.permit?).to eq(true)
+      # below is false b/c facility doesn't have the written test service
+      expect(@facility_1.administer_written_test(@registrant_1)).to eq(false)
+
+      # Written test can only be given to registrants with a permit && who are at least 16 years old.
```
KatBrandt, line 76: All these comments that indicate what you're testing, these should be separate it blocks. :)

spec/facility_spec.rb
```diff
@@ -1,26 +1,174 @@
...
+      expect(@registrant_1.license_data).to eq({:written=>true, :license=>true, :renewed=>true})
+
+      expect(@facility_1.renew_drivers_license(@registrant_2)).to eq(true)
+      expect(@registrant_2.license_data).to eq({:written=>true, :license=>true, :renewed=>true})
+
+      expect(@facility_1.renew_drivers_license(@registrant_3)).to eq(false)
+      expect(@registrant_3.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+    end
+  end
+
```
KatBrandt, line 172: Overall, really great job testing all the different combinations of these conditions for these methods.

lib/facility_factory.rb
```diff
@@ -0,0 +1,89 @@
+class FacilityInfo
+    attr_reader :geom, :id, :office, :address_1, :address_2,
```
KatBrandt, line 2: So in coding, we have this idea of only coding what we need at that moment. Which means not including all these attributes. We only want to bring in the attributes that we need to create a Facility, so looking back at the Facility class, there are only a handful that we need. Most of these can be left out.

lib/facility_factory.rb
```diff
@@ -0,0 +1,89 @@
+class FacilityInfo
```
KatBrandt, line 1: Remember our convention, the name of the class needs to match the file name. So this should be `class FacilityFactory`

lib/facility_factory.rb
```diff
@@ -0,0 +1,89 @@
...
+    #         ending = details[end_key]
+    #         "#{day.capitalize}: #{start} - #{ending}" if start && ending
+    #     end.compact
+    #     hours.join(', ')
+  
+    # end
+
+end
+
+class FacilityFactory
```
KatBrandt, line 55: oh, interesting. So every class should have it's own file. We don't want multiple classes in one file. I don't think you need the Facility Info class though, all that ingestion of data should happen in the FacilityFactory

lib/facility_factory.rb
```diff
@@ -0,0 +1,89 @@
...
+        def create_office(info)
+            info.map do |office_info|
```
KatBrandt, line 57: Good use of map and creating new Facility objects.

lib/registrant.rb
```diff
@@ -0,0 +1,19 @@
+class Registrant
+    attr_reader :name, :age, :permit, :license_data
```

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,14 @@
...
+  def create_vehicles(data)
+    data.map do |record|
+      Vehicle.new(
+        vin: record[:vin_1_10],
+        year: record[:model_year],
+        make: record[:make],
+        model: record[:model],
+        engine: :ev
+      )
+    end
```
KatBrandt, line 12: Yeah! This is good. This is more what the FacilityFactory should look like. Only the data coming in that we need to create the object.

KatBrandt (commented):
Overall this looks good. Your tests are verifying behavior and it looks like you have a good grasp on enumerables and hashes.

PR #508, opened 2024-07-16
Terra Manning

TDManning:
DMV Project Final Submission 

*Please note that due to external data source errors CO DMV related code has been temporarily commented out.


README.md
```diff
@@ -1,3 +1,10 @@
 # The DMV
 
-This is the starter repo for the BE Mod1 DMV project.
+## Reflections after debugging, Iteration 1, pt.1
+
+1. Describe the steps you took to dig in to this code base. What was your process?**<br />
+In order to build out this code base I started by reading the full interaction pattern of each iteration. After getting a through understanding of the interaction pattern I sequentially wrote tests that would ensure that I was showing what my code is going to do. Next I wrote code to pass the tests. In order to confirm what I thought was happening accurately, I tested the interaction pattern in pry to verify my expected output.
+
+2. What was hard about working with code you did not write?**<br /> At first the code looked unfamiliar (becasue it was!) and it took some time to look through the files to understand where I was jumping in. Overall it wasn't particularly hard because I liked the ways it was structured overall. Bugs are to be expected and having another coder to look for bugs can be helpful in trying to fix them.
```
KatBrandt, line 8: Great insight!

lib/facility.rb
```diff
@@ -1,14 +1,59 @@
 class Facility
-  attr_reader :name, :address, :phone, :services
+  attr_reader :name, :address, :phone, :services, :collected_fees, :registered_vehicles, :registration_date
```
KatBrandt, line 2: We do not want or need the `:registration_date` in the attr_reader here because we don't have an attribute called `@registration_date` that we need to read.

lib/facility.rb
```diff
@@ -1,14 +1,59 @@
...
   end
-end
+
+  def register_vehicle(vehicle)
+    if @services.include?("Vehicle Registration") 
+      @registered_vehicles << vehicle 
+      vehicle.registration_date = Date.today
+      collect_fees(vehicle)
+      @registered_vehicles
+    end
```
KatBrandt, line 23: Nice job here including all the logic for all the side effects this method needed to accomplish. I like that you abstracted some of that logic into helper methods to keep this method very clean and readable.

KatBrandt: Actually, we're missing one thing! This method should also be setting the plate_type of our vehicle.

lib/facility.rb
```diff
@@ -1,14 +1,59 @@
...
+      @collected_fees += 25
+    elsif vehicle.plate_type == :ev
+      @collected_fees += 200
+    else
+      @collected_fees += 100
+    end 
+  end 
+
+  def administer_written_test(registrant)
+    if @services.include?("Written Test") && registrant.permit? && registrant.age >= 16
```
KatBrandt, line 37: Great job with all your logical operators here

lib/facility.rb
```diff
@@ -1,14 +1,59 @@
...
+      registrant.license_data[:written] = true
+    else
+      false
+    end
+  end
+
+  def administer_road_test(registrant)
+    if registrant.license_data[:written] == true
```
KatBrandt, line 45: This should be looking for more conditions to determine if they can take the road test. The facility needs to have this service for one.

spec/facility_spec.rb
```diff
@@ -20,7 +33,160 @@
...
+    end
+  end
+
+  describe "#register_vehicle camaro" do 
+    it 'can register a specific vehicle, antique type' do 
+      expect(@camaro).to be_an_instance_of(Vehicle)
+      expect(@facility.add_service('Vehicle Registration')).to eq(["Vehicle Registration"])
+      @facility.register_vehicle(@camaro)
+
+      expect(@camaro.plate_type).to eq(:antique)
```
KatBrandt, line 70: So this line is passing because it's calling the method of `plate_type` in the Vehicle class, however this is not happening in the register_vehicle method.

lib/vehicle.rb
```diff
@@ -22,4 +25,14 @@ def antique?
   def electric_vehicle?
     @engine == :ev
   end
-end
+
+  def plate_type
```
KatBrandt, line 29: This does work, but this plate_type should be set upon successful registration of the vehicle so we need to tie this method into the register_vehicle method.

spec/facility_spec.rb
```diff
@@ -20,7 +33,160 @@
...
+      @facility.administer_written_test(@registrant_3)
+      @facility.add_service('Road Test')
+      @facility.administer_road_test(@registrant_1)
+      @facility.administer_road_test(@registrant_2)
+      @facility.administer_road_test(@registrant_3)
+      @facility.add_service('Renew License')
+      @facility.renew_drivers_license(@registrant_1)
+      @facility.renew_drivers_license(@registrant_2)
+      @facility.renew_drivers_license(@registrant_3)
+
```
KatBrandt, line 186: Ok to use white space between "sections" of this test for readability.

lib/registrant.rb
```diff
@@ -0,0 +1,20 @@
+class Registrant
+  attr_reader :name, :age, :permit, :license_data
```

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,27 @@
...
+        registration_date: nil,
+        vin: registration[:vin_1_10], 
+        year: registration[:model_year]
+      }
+
+    vehicle = Vehicle.new(specific_vehicle)
+    created_vehicles << vehicle
+    end
+    created_vehicles
+  end
```
KatBrandt, line 25: Nice job with this, however it's not wired up quite right. On line 5, you're setting your instance variable equal to the local variable of `created_vehicles` that that variable has not been assigned so it's currently nil. And then in your create_vehicles method, you're not reseting the value of your instance variable so it's not being updated. This method will still work by itself though since the return value of the method is the array of vehicle objects. Also, you don't need the accumulator on line 9 since you're using .map which already returns the transformed array. 
Good work with setting up that hash and creating the new Vehicle objects.

KatBrandt (commented):
Overall good work here. We do want to make sure we're fully understanding the iteration patterns and what each method should be doing and ensuring that our tests are accurately testing that functionality.

PR #510, opened 2024-07-16
Stefan J Bloom

stefanjshepard:
1st 3 iterations complete.

Colorado data set broken when embarking on final iteration, so it is missing.


lib/facility.rb
```diff
@@ -1,14 +1,65 @@
...
-    @address = address
-    @phone = phone
+  def initialize(facility_info)
+    @name = facility_info[:name]
+    @address = facility_info[:address]
+    @phone = facility_info[:phone]
     @services = []
+    @registered_vehicles = []
+    @collected_fees = 0
+    @facility_created = []
```
KatBrandt, line 17: So this looks like it's an attribute you've defined for the FacilityFactory class, not the Facility class, so we don't want this in here.

lib/facility.rb
```diff
@@ -1,14 +1,65 @@
...
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    if @services.include?('Vehicle Registration') == true
+      @registered_vehicles << vehicle
+      vehicle.set_registration_date
+      vehicle.give_plate
```
KatBrandt, line 28: I really like that you made helper methods in the Vehicle class to handle the responsibility of changing the vehicle attribute.

lib/facility.rb
```diff
@@ -1,14 +1,65 @@
...
+          @collected_fees += 200
+        elsif vehicle.antique? == true
+          @collected_fees += 25
+        else
+          @collected_fees += 100
+        end
+        return [vehicle] 
+    else 
+      return nil
+    end
```
KatBrandt, line 39: Overall, great work with all this logic and including all the side effects we wanted register_vehicle to display. Moving forward, I want us to start thinking about how we can break this logic up into more helper methods, that way each method can adhere to SRP (Single Responsibility Principle).

lib/facility.rb
```diff
@@ -1,14 +1,65 @@
...
+  end
+
+  def administer_written_test(registrant)
+    if @services.include?('Written Test') == true && registrant.age >= 16 && registrant.permit? == true
```
KatBrandt, line 43: Yeah, great job with your Logical Operators here. One thing we can do to clean this up, is remove comparison operators when they already return a boolean, so line 43 should be: `if @services.include?('Written Test') && registrant.age >= 16 && registrant.permit? `

spec/facility_spec.rb
```diff
@@ -16,11 +24,125 @@
...
+      @facility.register_vehicle(@camaro)
+
+      expect(@facility.registered_vehicles).to eq([@cruz, @bolt, @camaro])
+    end
+
+    it 'sets todays date as registration date' do
+      @facility = Facility.new({name: 'DMV Tremont Branch', address: '2855 Tremont Place Suite 118 Denver CO 80205', phone: '(720) 865-4600'})
+      @facility.register_vehicle(@cruz)
+
+      expect(@cruz.set_registration_date).to eq(Date.today)
```
KatBrandt, line 58: In this assertion, we don't want to call `@cruz.set_registration_date` because that is actually running that method, which means we're not testing the behavior of the `register_vehicle` method. We'd want to change this line to `expect(@cruz.registration_date).to eq(Date.today)` which would prove that it was the `register_vehicle` method that changed the registration date for the vehicle.

spec/facility_spec.rb
```diff
@@ -16,11 +24,125 @@
...
+    end
+
+    it 'gives a license plate' do
+      @facility = Facility.new({name: 'DMV Tremont Branch', address: '2855 Tremont Place Suite 118 Denver CO 80205', phone: '(720) 865-4600'})
+      @facility.register_vehicle(@cruz)
+
+      expect(@cruz.give_plate).to eq(:regular)
```
KatBrandt, line 65: Same thing here

spec/facility_spec.rb
```diff
@@ -16,11 +24,125 @@
...
+     expect(@facility.collected_fees).to eq(0)
+    end
+
+    it 'increases with every vehicle registration' do
+      @facility = Facility.new({name: 'DMV Tremont Branch', address: '2855 Tremont Place Suite 118 Denver CO 80205', phone: '(720) 865-4600'})
       @facility.add_service('Vehicle Registration')
-      expect(@facility.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
+      @facility.register_vehicle(@cruz)
+
+      expect(@facility.collected_fees).to eq(100)
```
KatBrandt, line 81: Yup, good test. Now we'd want to add one more vehicle to ensure that collected_fees will successfully add the additional registration fees.

spec/facility_spec.rb
```diff
@@ -16,11 +24,125 @@
...
+      @facility.administer_written_test(@registrant)
+
+      expect(@facility.administer_written_test(@registrant)).to eq(true)
+    end
+
+    it 'can only be administered to registrants 16 and older and with a permit' do
+      @facility = Facility.new({name: 'DMV Tremont Branch', address: '2855 Tremont Place Suite 118 Denver CO 80205', phone: '(720) 865-4600'})
+      @registrant = Registrant.new('Penny', 16 )
+
+      expect(@facility.administer_written_test(@registrant)).to eq(false)
```
KatBrandt, line 102: Yes, good job testing the false value here.

lib/registrant.rb
```diff
@@ -0,0 +1,45 @@
+class Registrant
+    
+    attr_reader :name,
+                :age,
+                :permit,
```

lib/registrant.rb
```diff
@@ -0,0 +1,45 @@
...
+                :license_data,
+                :take_written_test
```
KatBrandt, line 7: This should not be in the attr_reader, we dont' have an attribute called `@take_written_test` that we're trying to read.

lib/registrant.rb
```diff
@@ -0,0 +1,45 @@
...
+    def earn_permit
+        @permit = true
+    end
+
+    def take_written_test
+        if @license_data[:written] = true
+            true
+        else @age < 16 && permit? == false
+            false
+        end
```
KatBrandt, line 33: I'm not sure why we have all these conditions in this method. We already tested all these conditions in the Facility class method.

lib/vehicle.rb
```diff
@@ -22,4 +26,18 @@ def antique?
   def electric_vehicle?
     @engine == :ev
   end
+
+  def set_registration_date
+    @registration_date = Date.today
+  end
+
+  def give_plate
```
KatBrandt, line 34: Maybe `set_plate_type` would be a better name for this method.

spec/vehicle_spec.rb
```diff
@@ -33,4 +37,31 @@
...
+      expect(@cruz.plate_type).to eq(nil)
+
+      @cruz = Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice, plate_type: :regular} )
+      @bolt = Vehicle.new({vin: '987654321abcdefgh', year: 2019, make: 'Chevrolet', model: 'Bolt', engine: :ev, plate_type: :ev} )
+      @camaro = Vehicle.new({vin: '1a2b3c4d5e6f', year: 1969, make: 'Chevrolet', model: 'Camaro', engine: :ice, plate_type: :antique} )
+
+      expect(@cruz.plate_type).to eq(:regular)
+      expect(@bolt.plate_type).to eq(:ev)
+      expect(@camaro.plate_type).to eq(:antique)
+    end
```
KatBrandt, line 64: Good job writing tests for the additional methods you created in the Vehicle class.

KatBrandt (commented):
Overall, great work! You're on the right track.

PR #511, opened 2024-07-16
Bryan Willett

bwillett2003:
Second Solo project submission pull request.

### DIFF
--- a/lib/facility.rb
+++ b/lib/facility.rb
@@ -8,7 +8,7 @@ class Facility
     @services = []
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
 end
--- a/lib/facility.rb
+++ b/lib/facility.rb
@@ -1,14 +1,20 @@
 class Facility
-  attr_reader :name, :address, :phone, :services
+  attr_reader :name, :address, :phone, :services, :registered_vehicle, :collected_fees
 
   def initialize(details)
     @name = details[:name]
     @address = details[:address]
     @phone = details[:phone]
     @services = []
+    @registered_vehicles = []
+    @collected_fees = 0
   end
 
   def add_service(service)
     @services << service
+  end
+
+  def register_vehicle(vehicle)
+    
   end
 end
--- a/lib/facility.rb
+++ b/lib/facility.rb
@@ -23,5 +23,20 @@ class Facility
     else
       @collected_fees += 100
     end
+    vehicle.plate_type = vehicle.type
+    @registered_vehicles << vehicle
+  end
+
+  def administer_written_test(registrant)
+    if registrant.permit? && registrant.age >= 16
+      registrant.taken_written_test = true
+    end
+  end
+
+  def administer_road_test(registrant)
+    if registrant.taken_written_test && registrant.passed_written_test
+      registrant.taken_road_test = true
+      registrant.issue_license = true
+    end
   end
 end
--- a/lib/facility.rb
+++ b/lib/facility.rb
@@ -15,6 +15,13 @@ class Facility
   end
 
   def register_vehicle(vehicle)
-    
+    case vehicle.type
+    when :antique
+      @collected_fees += 25
+    when :ev
+      @collected_fees += 200
+    else
+      @collected_fees += 100
+    end
   end
 end
--- a/lib/facility.rb
+++ b/lib/facility.rb
@@ -36,7 +36,13 @@ class Facility
   def administer_road_test(registrant)
     if registrant.taken_written_test && registrant.passed_written_test
       registrant.taken_road_test = true
-      registrant.issue_license = true
+      registrant.issued_license = true
+    end
+  end
+
+  def renew_license(registrant)
+    if registrant.issued_license
+      registrant.license_renewed = true
     end
   end
 end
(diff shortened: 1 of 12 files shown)


lib/facility.rb
```diff
@@ -1,14 +1,73 @@
...
+       @collected_fees += 200
+        vehicle.plate_type = :ev
+      else
+        @collected_fees += 100
+        vehicle.plate_type = :regular
+      end
+    
+    vehicle.registration_date = Date.today
+    @registered_vehicles << vehicle
+  end
```
KatBrandt, line 41: Your logic in this method looks good, nice job working through all the things this method needs to accomplish. Moving forward, I'd love to see you start to break this logic up into helper methods. Every method should do ONE thing, and this method is doing about 5 different things. :) We can start putting logic in helper methods to have each helper method taking care of one piece of this logic. 

Also, I'm noticing that you forgot to check that the Facility had the correct services before performing any of this logic.

lib/facility.rb
```diff
@@ -1,14 +1,73 @@
...
+
+  def administer_written_test(registrant)
+    if @services.include?('Written Test') && registrant.permit? && registrant.age >= 16
```
KatBrandt, line 44: Great job with your Logical operators here to check all these conditions.

spec/facility_spec.rb
```diff
@@ -3,6 +3,11 @@
 RSpec.describe Facility do
   before(:each) do
     @facility = Facility.new({name: 'DMV Tremont Branch', address: '2855 Tremont Place Suite 118 Denver CO 80205', phone: '(720) 865-4600'})
+    @cruz = Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice} )
+    @bolt = Vehicle.new({vin: '987654321abcdefgh', year: 2019, make: 'Chevrolet', model: 'Bolt', engine: :ev} )
+    @camaro = Vehicle.new({vin: '1a2b3c4d5e6f', year: 1969, make: 'Chevrolet', model: 'Camaro', engine: :ice} )
+
+
```
KatBrandt, line 10: There are no tests in here for the methods that you added to the Facility class. Remember, every method should have a test in it's corresponding test file that accurately tests ALL side effects of that method. I encourage you to do a code share with a cohortmate to see what those tests could look like for this class.

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    #   expect(facilities).to be_an(Array)
+    #   expect(facilities.first).to be_a(Facility)
+    #   expect(facilities.first.name).to be_a(String)
+    #   expect(facilities.first.address).to be_a(String)
+    #   expect(facilities.first.phone).to be_a(String)
+    # end
+
+    it 'creates Facility objects from NY data' do
+      facilities = @factory.create_facilities(@ny_facilities, :ny)
```
KatBrandt, line 26: This works, passing in the :ny, but it makes this less dynamic. Is there something we could have looked for in the data set where in the method in the FacilityFactory class, the code could determine where this data was coming from? Like a key in the data of state or something? We want to remove the human input as much as possible and have the code do the work.

lib/facility_factory.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+  end
+
+  def create_ny_facilities(data)
+    data.map do |facility_data|
+      name = facility_data[:office_name]
+      address = "#{facility_data[:street_address_line_1]}, #{facility_data[:city]} #{facility_data[:state]} #{facility_data[:zip_code]}"
+      phone = facility_data[:public_phone_number]
+      Facility.new(name: name, address: address, phone: phone)
+    end
```
KatBrandt, line 23: Good job here using .map and creating new Facility objects

lib/registrant.rb
```diff
@@ -0,0 +1,39 @@
+class Registrant
+    attr_reader :name, 
+                :age, 
+                :permit,
```

lib/registrant.rb
```diff
@@ -0,0 +1,39 @@
...
+        }
+    end
+
+    def permit?
+        @permit
+    end
+
+    def earn_permit
+        @permit = true
+        return
```
KatBrandt, line 24: don't need this return keyword

lib/registrant.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+
+    def has_permit?
```
KatBrandt, line 27: We already have this method, it's `.permit?` both of these methods are currently doing the same thing, only need one of them.

lib/registrant.rb
```diff
@@ -0,0 +1,39 @@
...
+        permit?
+    end
+    
+    def passed_tests?
+        @license_data[:written] && @license_data[:license]
```
KatBrandt, line 32: Nice logic here!

KatBrandt: Remember though, for every method you add, we need to add a test in the spec file.

KatBrandt (commented):
You're on the right track here, looks like you have a good understanding of hashes and enumerables which is great. I would love to see you focus on writing well organized, robusts tests for every method. I'd also like to see you start to create helper methods so that each method you write is only doing one thing.

PR #512, opened 2024-07-16
Brody DMV

BrodyNics77:
Finished all tasked!

### DIFF
--- a/lib/facility.rb
+++ b/lib/facility.rb
@@ -5,7 +5,7 @@ class Facility
     @name = name
     @address = address
     @phone = phone
-    @services = []
+    @services = services
   end
 
   def add_services(service)
@@ -35,5 +35,18 @@ class Facility
     if registrant.has_license?
       registrant.renew_license
     end
+  end
+
+  def ny_dmv_office_locations
+    [
+      {name: 'DMV huntington' phone: '7184774820' address: 'street_address_line_1' }
+      [name: 'DMV Millbrook' phone: '8456774080' address: '15 Merritt anvenue']
+    ]
+  end
+
+  def mo_dmv_office_locations
+    [
+      {name: }
+    ]
   end
 end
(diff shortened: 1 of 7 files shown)


lib/facility.rb
```diff
@@ -1,14 +1,52 @@
 class Facility
   attr_reader :name, :address, :phone, :services
 
-  def initialize(name, address, phone)
+  def initialize(name, address, phone, services)
```
KatBrandt, line 4: So we don't need to add services here in the argument list since we're making it a default on line 8. Also, line 8 should be what you had originally @services = []

lib/facility.rb
```diff
@@ -1,14 +1,52 @@
...
-    @services = []
+    @services = services
   end
 
   def add_services(service)
     @services << service
   end
+
+  def register_vehicles(vehicle)
+    case vehicle.type
```
KatBrandt, line 16: I don't believe Vehicle has a method or attribute of .type. There was a plate_type potentially though.

lib/facility.rb
```diff
@@ -1,14 +1,52 @@
...
+    when :antique
+      vehicle.register(:antique, 25)
```
KatBrandt, line 18: This implies that there is a method called register in the Vehicle class, but I"m not seeing that you added that.

lib/facility.rb
```diff
@@ -1,14 +1,52 @@
...
+    when :ev
+      vehicle.register(:ev, 200)
+    else
+      vehicle.register(:regular, 100)
+    end
+  end
+
+  def administer_written_test(registrant)
+    if registrant.has_permit? && registrant.age >= 16
```
KatBrandt, line 27: Good job checking these two conditions, we also wanted to check the condition that the Facillity had the service that would allow for the written test.

PR #515, opened 2024-07-16
Kyle Delaney

kylomite:


lib/facility.rb
```diff
@@ -1,14 +1,77 @@
...
+              :phone,
+              :services,
+              :registered_vehicles,
+              :collected_fees
 
-  def initialize(name, address, phone)
-    @name = name
-    @address = address
-    @phone = phone
+  def initialize(fac_info)
```
KatBrandt, line 11: Just 'info' would have been a fine variable name. That is a nit pick comment though. Take it or leave it. :)

lib/facility.rb
```diff
@@ -1,14 +1,77 @@
...
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    if @services.include?('Vehicle Registration')
+      add_to_registered_vehicle(vehicle) 
+      update_vehicle_registration(vehicle)
+      plate_service(vehicle)
+    end
```
KatBrandt, line 29: I really like that you made the responsibility of this method to be more of a delegator, just calling other methods and orchestrating the chain of events, nice work with that. 
I would change the name of some of your methods though. Like 26 I would have added an `s` to the end to indicate we're adding the vehicle to a collection. And line 28, maybe call `set and calculate plate fees` or something, but then that shows that method is doing 2 things, and maybe that logic should be two separate methods.

lib/facility.rb
```diff
@@ -1,14 +1,77 @@
...
+  end
+
+  def add_to_registered_vehicle(vehicle)
+    @registered_vehicles << vehicle
+  end
+
+  def update_vehicle_registration(vehicle)
+    vehicle.registration_date = Date.today
```
KatBrandt, line 37: I would have prefered this method to have been in the Vehicle class, maybe called `set_registration_date` or something, so then in the register_vehicle method, you could have just called `vehicle.set_registration_date` and then this logic would be housed in the Vehicle class since we're changing a Vehicle attribute.

lib/facility.rb
```diff
@@ -1,14 +1,77 @@
...
+  end
+
+  def plate_service(vehicle)
+    if vehicle.antique? == true
```
KatBrandt, line 41: `vehicle.antique?` already returns a boolean, so no need to compare it to a boolean.

lib/facility.rb
```diff
@@ -1,14 +1,77 @@
...
+    end
+  end
+
+  def administer_written_test(registrant)
+
+    if @services.include?("Written Test") && registrant.permit == true
+      registrant.license_data[:written] = true
+    else
+      false
+    end
```
KatBrandt, line 59: Good job checking for the services and the permit status, also needed to check the condition that the registrant is >= 16.

spec/facility_spec.rb
```diff
@@ -23,4 +37,173 @@
...
+    end
+
+    it 'will be unavailable at facilities that do not offer the service' do
+      facility_2 = Facility.new({name: 'DMV Northeast Branch', address: '4685 Peoria Street Suite 101 Denver CO 80239', phone: '(720) 865-4600'})
+      facility_2.register_vehicle(@cruz)
+
+      expect(facility_2.registered_vehicles).to be_empty
+    end
+
+    it 'will use all helper methods to fulfill processed' do
```
KatBrandt, line 52: Let's be more specific with our it statements. 
So it "sets plate type and collects plate fees"

spec/facility_spec.rb
```diff
@@ -23,4 +37,173 @@
...
+  
+  describe '#add_to_registered_vehicle' do
+    before(:each) do
+      @facility.add_service('Vehicle Registration')
+    end
+
+    it 'will add specified vehicle to @register_vehicles array' do 
+      expect(@facility.registered_vehicles).not_to include @cruz
+
+      expect{@facility.add_to_registered_vehicle(@cruz)}.to change{@facility.registered_vehicles}.to([@cruz])
```
KatBrandt, line 72: Interesting syntax, that works.

spec/facility_spec.rb
```diff
@@ -23,4 +37,173 @@
...
+    end
+  end
+
+  describe '#update_vehicle_registration' do
+    before(:each) do
+      @facility.add_service('Vehicle Registration')
+    end
+    
+    it 'will assign current date as vehicles registration date' do
```
KatBrandt, line 81: This is a great it statement. :)

spec/facility_spec.rb
```diff
@@ -23,4 +37,173 @@
...
+
+      expect{facility_2.administer_written_test(@registrant)}.not_to change{@registrant.license_data}
+    end
+    
+    it 'will be unavailable to registrants who have not earned a permit' do
+      expect{@facility.administer_written_test(@registrant_2)}.not_to change{@registrant_2.license_data}
+    end
+
+    it 'will update registrants license_data attribute to include a passed written test' do
+      expect{@facility.administer_written_test(@registrant)}.to change{@registrant.license_data}.from({:written=>false, :license=>false, :renewed=>false}).to({:written=>true, :license=>false, :renewed=>false})
```
KatBrandt, line 137: This syntax is fine for testing, but it does create very long horizontal lines which can be harder to read than shorter, vertical lines. :)

spec/facility_spec.rb
```diff
@@ -23,4 +37,173 @@
...
+    it 'will be unavailable to registrants who have not passed road test' do
+      @facility.administer_written_test(@registrant_2)
+
+      expect{@facility.renew_drivers_license(@registrant_2)}.not_to change{@registrant_2.license_data}
+    end
+
+    it 'will update registrants license_data attribute to include a renewed license' do
+      @facility.administer_written_test(@registrant)
+      @facility.administer_road_test(@registrant)
+
```
KatBrandt, line 205: Good job testing all the conditions here for these methods. You did forget the age condition, but I'm sure you would have tested that too.

lib/facility_factory.rb
```diff
@@ -0,0 +1,54 @@
...
+            :co => [],
+            :ny => [],
+            :mo => []
+        }
+    end
+    
+    def create_facilities(facilities_list)
+        if facilities_list[0][:state] == "CO"
+            facilities_list.each {|facility|
+            transform_co_facility_data(facility)
```
KatBrandt, line 21: I like that you used this flow control and then encapsulated the logic in helper methods.

lib/facility_factory.rb
```diff
@@ -0,0 +1,54 @@
...
+            }
+        else facilities_list[0][:state] == "MO"
+            facilities_list.each {|facility|
+            transform_mo_facility_data(facility)
+            }
+        end
+    end
+#COLORADO
+    def transform_co_facility_data(facility)
+        @facility_details [:name] = facility[:dmv_office]
```
KatBrandt, line 35: Should not have a space between the name of the hash and the bracket

lib/facility_factory.rb
```diff
@@ -0,0 +1,54 @@
...
+    
+    def create_facilities(facilities_list)
+        if facilities_list[0][:state] == "CO"
+            facilities_list.each {|facility|
+            transform_co_facility_data(facility)
+            }
+        elsif facilities_list[0][:state] == "NY"
+            facilities_list.each {|facility|
+            transform_ny_facility_data(facility)
+            }
```
KatBrandt, line 26: Curly braces need to be a one-line syntax. Otherwise use do/end blocks.

lib/facility_factory.rb
```diff
@@ -0,0 +1,54 @@
...
+require './lib/facility'
+
+
+class Facility_Factory 
+    attr_accessor :facility_details, 
+                  :facilities
+
+    def initialize 
+        @facility_details = {:name => nil, :address => nil, :phone => nil, :services => [], :registered_vehicles => [], :collected_fees => 0}
+        @facilities = {
```
KatBrandt, line 11: I like that you created an attribute here to store all the Facility objects you've created.

lib/facility_factory.rb
```diff
@@ -0,0 +1,54 @@
+require './lib/dmv_data_service'
+require './lib/facility'
+
+
+class Facility_Factory 
+    attr_accessor :facility_details, 
+                  :facilities
+
+    def initialize 
+        @facility_details = {:name => nil, :address => nil, :phone => nil, :services => [], :registered_vehicles => [], :collected_fees => 0}
```
KatBrandt, line 10: I'm going back and forth on how I feel about this attribute. I don't think it's necessary because depending on the state, some of your values remain nil or 0 or empty. I think this could be utilized better by keeping the necessary keys within the methods below so you only establish the ones you need... but again, I'm going back and forth on it. Either way, great job with this solution as a whole.

lib/registrant.rb
```diff
@@ -0,0 +1,24 @@
+class Registrant
+    attr_reader :name,
+                :age,
+                :permit
```

lib/registrant.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    def initialize(name, age, permit = false)
+        @name = name
+        @age = age
+        @permit = permit
+        @license_data = {:written=>false, :license=>false, :renewed=>false}
+    end
+
+    def earn_permit
+        if @permit == false && @age > 15
```
KatBrandt, line 16: Remember in the directions, there was no age limit to earn a permit. If there was though, you'd want to use your method call that indicates a boolean return, so line 16 would be `if permit? && age > 15`

lib/vehicle.rb
```diff
@@ -5,14 +5,22 @@ class Vehicle
               :year,
               :make,
               :model,
-              :engine
+              :engine,
+              :registration_date,
+              :antique,
+              :electric_vehicle
```
KatBrandt, line 11: Again, don't need these attr_readers because you have a method that reads these attrbitues already with a ? at the end.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,30 @@
+require './lib/dmv_data_service'
+require './lib/vehicle'
+
+class VehicleFactory
+
+    attr_accessor :vehicle_details,
```
KatBrandt, line 6: I don't think you need an attr_accessor. An attr_reader would be fine. We don't want to expose a method that can change these attributes until we absolutely need one.

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,30 @@
...
+                  :vehicles
+
+    def initialize
+        @vehicle_details = {vin: nil, year: nil, make: nil, model: nil, engine: :ev, plate_type: nil, registration_date: nil}
```
KatBrandt, line 10: Ok, the reason I'm not liking this attribute in the Factories is because they are not an attribute of the VehicleFactory class, these are attributes of the Vehicles you're creating. I could see an argument to keeping them here in the initialize, but I also think it would be better to keep these local within the method you're writing to reduce unnecessary setting of default values. 
There are also other ways of setting default values so you don't need to explicitly write out all the keys like this. Actuallly, since most of these default values are already nil, we don't need to set the hash at all like this since the defaults will already start as nil.

KatBrandt (commented):
Nice job with this and great job getting to the Iteration 4 extension! Your code is looking good as are your tests. We could use some better organization in the tests with describe blocks and it blocks in a few places, but overall, this is looking good.

PR #516, opened 2024-07-16
Joel Davalos 

jdavalos98:


README.md
```diff
@@ -1,3 +1,11 @@
 # The DMV
 
 This is the starter repo for the BE Mod1 DMV project.
+
+Reflection on Debugging:
+
+1. This time around the first step I took was to look over read the code and get a bit familiar then ran my spec files. From there pulled up each of the 3 spec files that had failing tests and also split screened with the lib files. From there I worked from the bottom up from the list of tests. Felt it was easier to start with debugging one test then moving to two then three until I was fully passing. 
+
+2. I think the hardest thing I found working on code I didnt write was trying to understand what the original author was trying to do with some of the code and navigating is test files. I was a bit confused at first with the added describe line with every test which is something I have not done.
```
KatBrandt, line 9: Did you find it easier to read through the tests with that added describe line? I find it helps with organizing it blocks.

lib/facility.rb
```diff
@@ -1,14 +1,34 @@
...
+      vehicle.plate_type = :ev
+      @collected_fees += 200
+    else
+      vehicle.plate_type = :regular
+      @collected_fees += 100
+    end
+    vehicle.registration_date = Date.today
+    @registered_vehicles << vehicle
+    @registered_vehicles
+  end
```
KatBrandt, line 33: Great job with all the logic here and all the side effects we wanted this register_vehicle method to have. Something to work on moving forward, is thinking about how we can refactor this method so that this method is only doing one thing. How can we break out some of these pieces into helper methods so that each method is doing one thing?

spec/facility_spec.rb
```diff
@@ -2,25 +2,67 @@
...
+      expect(@facility_2.services).to eq([])
+      @facility_2.add_service('New Drivers License')
+      @facility_2.add_service('Renew Drivers License')
+      @facility_2.add_service('Vehicle Registration')
+      expect(@facility_2.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
+    end
+  end
+  describe '#register_vehicle' do
+    it 'can register a vehicle if service is offered' do
+      @facility_1.add_service('Vehicle Registration')
```
KatBrandt, line 48: We also want to test what happens when this service is NOT added.

spec/facility_spec.rb
```diff
@@ -2,25 +2,67 @@
...
+      expect(@camaro.registration_date).to eq(Date.today)
+      expect(@camaro.plate_type). to eq(:antique)
+      expect(@facility_1.collected_fees).to eq(125)
+
+      @facility_1.register_vehicle(@bolt)
+      expect(@bolt.registration_date).to eq(Date.today)
+      expect(@bolt.plate_type).to eq(:ev)
+      expect(@facility_1.collected_fees).to eq(325)
+
+      expect(@facility_1.registered_vehicles).to eq([@cruz, @camaro, @bolt])
```
KatBrandt, line 65: Yeah, good job testing all these behaviors. We can start now to create more it blocks where each it block tests a different piece of this behavior. The it blocks should act as documentation to allow someone to read all the desired side effects and behaviors.

KatBrandt (commented):
You're off to a good start here Joel. Just keep practicing.

PR #517, opened 2024-07-17
Cameron Pittman

CPPittman:


lib/registrant.rb
```diff
@@ -0,0 +1,26 @@
+class Registrant 
+    attr_reader :name, :age, :permit, :license_data
```

lib/registrant.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    def initialize(name, age, permit, license_data)
```
KatBrandt, line 4: We don't need license_data here in the arguments since it's just an attribute we're setting for every Registrant object. We also wanted permit to have a default value so line 4 would like like this: `def initialize(name, age, permit = false)`

lib/registrant.rb
```diff
@@ -0,0 +1,26 @@
...
+    #    @age = age
+    # end
+
+    # def permit
+    #    @permit = permit
+    # end
+
+    # def license_data
+    #    @license_data = license_data
+    # end
```
KatBrandt, line 25: Good job recognizing that there are methods you've created with your attr_readers. We did want a method of `def permit?` to be the attr_reader for the `@permit` attribute since it returns a boolean.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+RSpec.describe @registrant do
+    before(:all) do
+    end
+
+    it 'is a registrant' do
+        registrant_1 = Registrant.new('Bruce', 18, true, "license") 
+    end
+
+    # it 'has a name' do
+    #   @name = name
```
KatBrandt, line 12: We don't have access to the `@name` variable in the spec file. We'd need to test this by first instantiating a new Registrant object on line 12 (like you have on line 8), then we can test the method of `.name`

PR #531, opened 2024-09-04
Danielle Cardona

dcardona23:


lib/dmv.rb
```diff
@@ -1,6 +1,14 @@
 class Dmv
 
-  def initialize
+  attr_reader :name, 
+              :address, 
+              :phone, 
+              :facilities
+
+  def initialize(attributes = {})
+    @name = attributes[:name]
```
nicktu12, line 9: Nice work assigning keys/values to this empty hash on initialize!

lib/dmv_factory.rb
```diff
@@ -0,0 +1,74 @@
...
+            collected_fees: facility_data[:collected_fees])
+        
+            add_service(facility, facility_data[:services_p])
+    
+            facility 
+            end
+        end
+           
+    def add_service(facility, services_string)
+        return unless services_string
```
nicktu12, line 68: Great work with the guard statement here

lib/dmv_factory.rb
```diff
@@ -0,0 +1,74 @@
...
+        registered_vehicles: facility_data[:registered_vehicles],
+        collected_fees: facility_data[:collected_fees])
+    
+        add_service(facility, facility_data[:services_p])
+        facility 
+        end
+    end
+
+    def create_ny_facilities(ny_dmv_office_locations)
+    ny_dmv_office_locations.map do |facility_data|
```
nicktu12, line 26: Good use of map in these methods! Some complex hash work, too

lib/facility.rb
```diff
@@ -1,14 +1,79 @@
 class Facility
-  attr_reader :name, :address, :phone, :services
+  attr_reader :name,
```
nicktu12, line 2: attr_accessor for these attributes would be preferred - this gives you read and write functionality! No need to have both for all the attributes

lib/facility.rb
```diff
@@ -1,14 +1,79 @@
...
+        vehicle.registration_date = Date.today
+        assign_plate_type(vehicle)
+        calculate_fees(vehicle)
+        @registered_vehicles << vehicle
+    else
+      false
+    end
+  end
+
+  def calculate_fees(vehicle)
```
nicktu12, line 50: LOVE the SRP here!! Plate type, fees, and registration are all handled in their own spaces

lib/facility.rb
```diff
@@ -1,14 +1,79 @@
...
+  end
+
+  def administer_written_test(registrant)
+    return false unless @services.include?("Written Test")
+    return false unless registrant.permit? == true && registrant.age >= 16
+
+    registrant.license_data[:written] = true
+  end
+
+  def administer_road_test(registrant)
```
nicktu12, line 67: Good job with these three methods. Because we're checking three conditions though, I would rather just have all three conditions being checked at the same time instead of checking one condition, then the other two. Let's just do it all in one go. so line 61 could like like this: if @services.include?('Written Test') && registrant.age >= 16 && registrant.permit? then just put line 61 inside the if block.

lib/registrant.rb
```diff
@@ -0,0 +1,25 @@
+class Registrant
+
+    attr_reader :name, 
+    :age, 
+    :license_data, 
+    :permit
```
nicktu12, line 6: no need for this reader on permit, as permit? will expose this data any way

lib/vehicle.rb
```diff
@@ -5,18 +5,31 @@ class Vehicle
               :year,
               :make,
               :model,
-              :engine
+              :engine,
+              :plate_type,
+              :registration_date
 
-  def initialize(vehicle_details)
+  attr_writer :vin,
```
nicktu12, line 12: Same as above - attr_accessor!

spec/registrant_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+            expect(@registrant_1.permit?).to eq(true)
+            expect(@registrant_2.permit?).to eq(false)
+        end
+
+        it 'adds permit when earned' do
+            @registrant_2.earn_permit
+            @registrant_3.earn_permit
+        
+            expect(@registrant_2.permit?).to eq(true)
+            expect(@registrant_3.permit?).to eq(true)
```
nicktu12, line 44: nice testing all around!

spec/registrant_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+
+            expect(@registrant_1.license_data).to eq({written: true, license: true, renewed: false})
+        end
+
+        it 'cannot take the road test without passing the written test' do
+            expect(@facility_1.administer_road_test(@registrant_1)).to eq(false)
+        end
+    end
+
+        describe 'license renewal' do
```
nicktu12, line 79: Some spacing nitpicks - dont include a newline here between your describe and it blocks, and watch your indentation

spec/spec_helper.rb
```diff
@@ -3,3 +3,6 @@
 require './lib/facility'
 require './lib/vehicle'
 require './lib/dmv_data_service'
+require './lib/registrant'
+require './lib/vehicle_factory'
+require './lib/dmv_factory'
```
nicktu12, line 8: 🔥

PR #560, opened 2024-10-22
Shadeau Christensen

shadeauchristensen:
- created vehicle_factory_spec and class files
-- cleaned up code in previous files to run with corresponding tests and classes
--- rspec run and have 0 failures


spec/facility_spec.rb
```diff
@@ -2,25 +2,227 @@
...
+    )
+    end
+
+    it 'sets up plate_type for facility_3' do
+      facility_3 = Facility.new(
+        { name: 'DMV Northwest Branch', 
+        address: '3698 W. 44th Avenue Denver CO 80211', 
+        phone: '(720) 865-4600' }
+    )
+    end
```
memcmahon, line 109: None of these tests have an `expect` line, so they are missing the Assertion phase of the test!  Remember S.E.A.T.

spec/facility_spec.rb
```diff
@@ -2,25 +2,227 @@
...
+      expect(@facility_1.renew_license(registrant_2)).to eq(true)
+      expect(registrant_2.license_data[:renewed]).to eq(true)
+   
+      registrant_3.pass_written_test
+      registrant_3.pass_road_test
+      expect(@facility_1.renew_license(registrant_3)).to eq(true)
+      expect(registrant_3.license_data[:renewed]).to eq(true)
+   
+# it does not renew license if registrant has not passed written test
+# it does not renew license if registrant has not passed road test
```
memcmahon, line 222: Looks like you maybe meant to add more tests here?

lib/facility.rb
```diff
@@ -1,14 +1,74 @@
...
+    else
+      100
+    end
+  end
+
+  private
+
+  def vehicle_age(vehicle)
+    Date.today.year - vehicle.year
+  end
```
memcmahon, line 69: When I see a method like this, that takes in an object, and only uses that object (and built in ruby) to return a value, it is an indicator to me that the method maybe should live on the object that is being passed in.  In this case, it would make more sense for the Vehicle to be able to tell you its own age (you would have a `age` method) on the vehicle class.

lib/facility.rb
```diff
@@ -1,14 +1,74 @@
...
+      true
+    else
+      false
+    end
+  end
+
+  def registered_vehicle(vehicle)
+    if @services.include?('Vehicle Registration')
+      vehicle.assign_registration_date
+      vehicle.assign_plate_type
```
memcmahon, line 49: I like these as helper methods!!

lib/vehicle.rb
```diff
@@ -1,25 +1,42 @@
...
+    elsif @engine == :ev
+      @plate_type = :ev
+    else
+      @plate_type = :regular
+    end
+  end
+
+  def assign_registration_date
+    @registration_date = Date.today
+  end
```
memcmahon, line 41: When you add additional methods to a class, make sure you also add tests for these methods!!!

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,20 @@
+class VehicleFactory
+  attr_reader :data, 
+              :vin, 
+              :make, 
+              :model, 
+              :year, 
+              :engine 
+              
+  def create_vehicles(data)
+    data.map do |vehicle_data|
```
memcmahon, line 10: excellent use of .map!

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,20 @@
+class VehicleFactory
+  attr_reader :data, 
+              :vin, 
+              :make, 
+              :model, 
+              :year, 
+              :engine
```
memcmahon, line 7: None of these are attributes of this class, so you don't need them (these are attributes of the `Vehicle`)

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+  
+    describe '#create_vehicles' do
+      it 'creates vehicles from wa_ev_registration data' do
+        vehicles = @factory.create_vehicles(@wa_ev_registrations)
+  
+        expect(vehicles).to be_an(Array)
+        expect(vehicles.any? { |vehicle| vehicle.make == 'TOYOTA' }).to be(true)
+        expect(vehicles.any? { |vehicle| vehicle.model == 'Prius Plug-in' }).to be(true)
+        expect(vehicles.any? { |vehicle| vehicle.year == '2013' }).to be(true)
```
memcmahon, line 23: Nice use of `any?` to verify that a specific Vehicle was created!

lib/registrant.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+
+    def collected_fees
+        @collected_fees ||= 0
+      end
+      
+      def register_vehicle(vehicle)
+        @register_vehicles ||= []
+        @register_vehicles << vehicle
+      end
```
memcmahon, line 44: These definitely don't belong on the Registrant!

spec/registrant_spec.rb
```diff
@@ -0,0 +1,50 @@
+require 'spec_helper'
+
+RSpec.describe Registrant do
```
memcmahon, line 3: We are missing tests for: `pass_written_test`, `pass_road_test`, `renew_license`, and `pass_license`

PR #561, opened 2024-10-22
Jono Sommers DMV Project (Mod1 - 2410)

JonoSommers:
Completed iterations 1 - 3 of the DMV project. 
The main functions of the project are creating vehicles, creating facilities, and determining if registrants are eligible for their license, permit, etc. 
All tests are included and passing.
All branches are pushed and pulled.


lib/facility.rb
```diff
@@ -1,14 +1,66 @@
...
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    if @services.include?('Vehicle Registration')
+      if vehicle.antique?
+        @collected_fees += 25
+        vehicle.set_plate_type
+        vehicle.set_registration_date
```
memcmahon, line 24: Love these as helper methods!

spec/registrant_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'spec_helper'
+
+RSpec.describe Registrant do
```
memcmahon, line 3: Oh no!  Looks like we are missing tests for `take_written_test`, `take_road_test` and `renew_license`. All methods should have at least one test!!

spec/vehicle_spec.rb
```diff
@@ -33,4 +35,12 @@
...
     end
   end
+
+  describe '#set_registration_date' do
+    it 'sets the registration date to the current day when called on a vehicle' do
+      expect(@cruz.registration_date).to be(nil)
+      @cruz.set_registration_date
+      expect(@cruz.registration_date).to eq(@today)
+    end
+  end
```
memcmahon, line 45: I think we are missing a test for `set_plate_type`

lib/creating_facilities.rb
```diff
@@ -0,0 +1,14 @@
+class CreatingFacilities
```
memcmahon, line 1: I maybe would have called this class `FacilityFactory` to match the `VehicleFactory`

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,14 @@
+class VehicleFactory
+    attr_reader :vehicles_made
+
+    def initialize
+        @vehicles_made = []
+    end
+
+    def create_vehicles(wa_ev_registrations)
```
memcmahon, line 8: Very technically, this method should have returned an array of the Vehicles created 😬 , but you have a nice workaround with the attribute `vehicles_made`.

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'spec_helper'
+
+RSpec.describe VehicleFactory do
```
memcmahon, line 3: Very nice factory tests!

lib/facility.rb
```diff
@@ -1,14 +1,66 @@
 class Facility
-  attr_reader :name, :address, :phone, :services
-
-  def initialize(name, address, phone)
-    @name = name
-    @address = address
-    @phone = phone
-    @services = []
+  attr_reader :name, :address, :phone, :services, :collected_fees, :registered_vehicles
+  def initialize(facility_info)
```
memcmahon, line 3: Great job getting the data from different sources to work! One thing you could think about is moving some of this data manipulation to a helper method.  You could imagine that all 50 states may have different data formats and that could make this method really overwhelming.  That refactor could look like:

```ruby
@name = set_name_attribure(facility_info)

...

def set_name_attribute(facility_info)
  facility_info[:name] || facility_info[:dmv_office] || facility_info[:office_name]
end
```

PR #563, opened 2024-10-22
Patrick Shim

pshim17:
Hi, Megan,

I forgot to commit frequently and create branches/pull requests until I started working on iteration 4.

Please check the link below to check the additional commits made and branches/pull requests I created to meet the requirements.

https://github.com/pshim17/the_dmv_extension_js


lib/facility.rb
```diff
@@ -1,14 +1,83 @@
...
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def registered_vehicles
+    @registered_vehicles
+  end
+
+  def collected_fees
```
memcmahon, line 22: I LOVE that you are calculating the collected fees based on the registered vehicles that you are already storing.  This is great Object Oriented Design!

lib/registrant.rb
```diff
@@ -0,0 +1,21 @@
+class Registrant
+    attr_accessor :name, :age, :permit, :license_data
```
memcmahon, line 2: Be careful with how many attributes you expose as attr_accessors.  In this case, you only need :permit, and :license_data - the others could remain attr_readers!

spec/registrant_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+            #Penny
+            expect(@registrant_2.name).to eq("Penny")
+            expect(@registrant_2.age).to eq(15)
+            expect(@registrant_2.permit?).to eq(false)
+            expect(@registrant_2.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+            expect(@registrant_2.earn_permit).to eq(true)
+            expect(@registrant_2.permit?).to eq(true)
+        end
+    end
+
```
memcmahon, line 31: When you change an attribute into an accessor, you should also add tests to verify that `.attribute = new_value` is working and is expected behavior.  You are creating new methods with those accessors!

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+        @factory = VehicleFactory.new
+        @wa_ev_registrations = DmvDataService.new.wa_ev_registrations
+        @factory.create_vehicles(@wa_ev_registrations)
+    end
+
+    describe '#initialize' do
+        it 'can initialize' do
+            expect(@factory).to be_an_instance_of(VehicleFactory)
+        end
+    end
```
memcmahon, line 14: You do want to test that the VehicleFactory does in fact create vehicles - this might look like:

```ruby
        it 'creates array of vehicles from datasource' do
            factory = VehicleFactory.new
            wa_ev_registrations = DmvDataService.new.wa_ev_registrations
            vehicles = factory.create_vehicles(wa_ev_registrations)
            expect(vehicles).to be_a(Array)
            vehicles.each do |vehicle|
                expect(vehicle).to be_a(Vehicle)
                expect(vehicle.vin).to_not eq(nil)
                expect(vehicle.year).to_not eq(nil)
                expect(vehicle.make).to_not eq(nil)
                expect(vehicle.model).to_not eq(nil)
                expect(vehicle.engine).to eq(:ev)
            end
            ```

File: lib/facility_factory.rb [hunk 5]
```diff
@@ -0,0 +1,13 @@
+class FacilityFactory
+    attr_reader :facility_list
+
+    def initialize
+        @facility_list = []
+    end
+
+    def create_facilities(facilities)
```
memcmahon, line 8: Though this is creating new facilities, it is not working exactly as it should.  Not every state has the same data structures - so this will give us facilities that may or may not have some attributes created (like name, phone number, or zipcode)

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,13 @@
+class VehicleFactory
+    attr_reader :vehicle_list
+
+    def initialize
+        @vehicle_list = []
+    end
+
+    def create_vehicles(vehicles)
+        @vehicle_list = vehicles.map do |vehicle_details|
+            Vehicle.new(vehicle_details)
```
memcmahon, line 10: because your vehicle_factory_spec tests are not quite robust enough, you are not seeing that the vehicles that you are creating are missing some attributes: vin, year, and engine are examples.

PR #565, opened 2024-10-22
Kevin Newland

kevin-newland:


lib/facility.rb
```diff
@@ -1,14 +1,89 @@
...
-    @phone = phone
+  attr_reader :name, 
+              :address, 
+              :phone, 
+              :services, 
+              :registered_vehicles, 
+              :collected_fees          
+  attr_accessor :age
+                
+  def initialize(facilities_details)
```
memcmahon, line 10: Great job getting the data from different sources to work! One thing you could think about is moving some of this data manipulation to a helper method.  You could imagine that all 50 states may have different data formats and that could make this method really overwhelming.  That refactor could look like:

```ruby
@name = set_name_attribure(facility_info)

...

def set_name_attribute(facility_info)
  facility_info[:name] || facility_info[:dmv_office] || facility_info[:office_name]
end
```

lib/facility.rb
```diff
@@ -1,14 +1,89 @@
...
-    @name = name
-    @address = address
-    @phone = phone
+  attr_reader :name, 
+              :address, 
+              :phone, 
+              :services, 
+              :registered_vehicles, 
+              :collected_fees          
+  attr_accessor :age
```
memcmahon, line 8: I'm pretty sure a facility doesn't have an age ;)

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+    describe '#initialize' do
+      it 'can initialize' do
+        @colorado.create_facilities(@co_dmv_office_locations)
+        expect(@colorado).to be_an_instance_of(FacilityFactory)
+        expect(@colorado.new_facilities).to be_a(Array)
+      end
+    end
+
+    describe 'create_facilities' do
+      it 'can create facilities' do
```
memcmahon, line 24: Great job testing that a facility is created AND created with the correct attributes!

lib/vehicle.rb
```diff
@@ -6,20 +6,26 @@ class Vehicle
               :make,
               :model,
               :engine
-
+              
+  attr_accessor :registration_date, :plate_type
```
memcmahon, line 10: converting these to attr_accessors actually creates new methods, so you should have tests to make sure that `.registration_date = <somevalue>` and `.plate_type = <somevalue>` work as expected.

lib/facility.rb
```diff
@@ -1,14 +1,89 @@
...
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    unless  @services.include?('Vehicle Registration')
+      return nil
+    else
+      if vehicle.antique?
+        vehicle.plate_type = :antique
```
memcmahon, line 29: You could think about moving the set plate_type and set registration_date to into helper methods in the vehicle class.

PR #566, opened 2024-10-22
Dustin Peukert DMV Project

DustinPeukert:


spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+      expect(created_facilities[2].address).to eq("3698 W. 44th Avenue Denver CO 80211") #does not have address line 2 in data
+      expect(created_facilities[2].phone).to eq("(720) 865-4600")
+    end
+    
+    it 'can create facilities using data sets with different formats' do
+      ny_dmv_info = DmvDataService.new.ny_dmv_office_locations
+      mo_dmv_info = DmvDataService.new.mo_dmv_office_locations
+
+      expect(ny_dmv_info.count).to eq(173)
+      expect(mo_dmv_info.count).to eq(178)
```
DustinPeukert, line 42: Did not think about the fact that API data can change. Should have used:
expect(mo_dmv_info.count == mo_facilities.count).to be true

lib/facility.rb
```diff
@@ -1,14 +1,80 @@
...
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    if @services.include?('Vehicle Registration')
+      vehicle.set_registration_date
+      vehicle.set_plate_type
```
memcmahon, line 25: Love these as helper methods

lib/facility.rb
```diff
@@ -1,14 +1,80 @@
...
-    @name = name
-    @address = address
-    @phone = phone
+  def initialize(facility_info)
+    @name = facility_info[:name]
+    @address = facility_info[:address]
+    @phone = facility_info[:phone]
     @services = []
+    @registered_vehicles = []
+    @collected_fees = 0
```
memcmahon, line 15: Since we have a collection of registered vehicles, you don't strictly _need_ to store collected_fees.  You could calulate this on the fly with a method that sums fees based on registered vehicles.  This is pretty nit-picky, so not a huge deal - just something to think about :)

lib/facility_factory.rb
```diff
@@ -0,0 +1,43 @@
+class FacilityFactory
+  def create_facilities(facility_data)
+    created_facilities = []
+
+    facility_data.each do |facility|
```
memcmahon, line 5: Great job getting the data from different sources to work! One thing you could think about is moving some of this data manipulation to a helper method.  You could imagine that all 50 states may have different data formats and that could make this method really overwhelming.  That refactor could look like:

```ruby
name = set_name_attribure(facility)

...

def set_name_attribute(facility)
  facility[:name] || facility[:dmv_office] || facility[:office_name]
end
```

spec/registrant_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+
+      expect(@registrant_2.permit?).to be true
+
+      @registrant_2.earn_permit #might take these out
+
+      expect(@registrant_2.permit?).to be true
+    end
+  end
+
+  describe '#pass_written_test' do
```
memcmahon, line 45: GREAT job testing all of your helper methods!

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    @vehicle_factory = VehicleFactory.new
+  end
+
+  describe '#initialize' do
+    it 'is an instance of VehicleFactory' do
+      expect(@vehicle_factory).to be_an_instance_of(VehicleFactory)
+    end
+  end
+
+  describe '#create_vehicles' do
```
memcmahon, line 14: This is a very robust test - I appreciate that you are verifying that a specific vehicle is being created with all of the correct attributes!

PR #567, opened 2024-10-22
Kristin Weiland

KMPWeiland:


lib/dmv_directory.rb
```diff
@@ -0,0 +1,16 @@
+class DmvDirectory
+  
+     def create_facilities(dmv_directory_array)
+        dmv_directory_array.map do |dmv_directory_hash|
+            Facility.new({
```
memcmahon, line 5: Great job getting the data from different sources to work! One thing you could think about is moving some of this data manipulation to a helper method.  You could imagine that all 50 states may have different data formats and that could make this method really overwhelming.  That refactor could look like:

```ruby
name: set_name_attribure(dmv_directory_hash)

...

def set_name_attribute(dmv_directory_hash)
  dmv_directory_hash[:name] || dmv_directory_hash[:dmv_office] || dmv_directory_hash[:office_name]
end
```

lib/vehicle.rb
```diff
@@ -5,14 +5,17 @@ class Vehicle
...
+              :registration_date
 
   def initialize(vehicle_details)
     @vin = vehicle_details[:vin]
     @year = vehicle_details[:year]
     @make = vehicle_details[:make]
     @model = vehicle_details[:model]
     @engine = vehicle_details[:engine]
+    @registration_date = nil
+    @plate_type = nil
```
memcmahon, line 18: Since you have a method that returns the plate_type, you don't need to have an instance variable as well!

spec/dmv_directory_spec
```diff
@@ -0,0 +1,65 @@
...
+            
+            expect(facilities).to be_a(Array)
+        end
+
+        it 'test that the first items in the facilities array is a Facility object' do
+        #aka tests that the first element is an instance of the Vehicle class
+            @new_co_dmv_offices = DmvDataService.new.co_dmv_office_locations
+            facilities = @directory.create_facilities(@new_co_dmv_offices)
+
+            expect(facilities.first).to be_an_instance_of(Facility)
```
memcmahon, line 24: I would also make sure to test that not only is a Facility created but that it has the attributes that you expect.  It could be something like:

```ruby
            facilities.each do |facility|
                expect(facility).to be_a(Facility)
                expect(facility.name).to_not eq(nil)
                expect(facility.address).to_not eq(nil)
                expect(facility.phone).to_not eq(nil)
            end
            ```

File: spec/dmv_directory_spec
Comment (memcmahon) [inline spec/dmv_directory_spec:1]: Make sure this file has the .rb extension!!!

File: spec/registrant_spec.rb [hunk 4]
```diff
@@ -0,0 +1,73 @@
...
+    end
+
+    describe '#earn_permit?' do
+      it 'can determine if a registrant earned their permit' do
+       @registrant_2.earn_permit
+       expect(@registrant_2.permit?).to eq(true)
+      end
+    end
+
+    # describe "#updates to license data" do
```
memcmahon, line 36: I like the decision to comment out these tests, since these are really more related to the facility 👍

spec/vehicle_spec.rb
```diff
@@ -33,4 +33,21 @@
       expect(@camaro.electric_vehicle?).to eq(false)
     end
   end
+
+  describe '#register' do
```
memcmahon, line 37: Excellent job testing your helper methods that you added to the vehicle!

PR #568, opened 2024-10-22
Jillian Bleggi- DMV Mod1 Project Submission

jbleggi:


lib/vehicle.rb
```diff
@@ -1,25 +1,21 @@
...
-              :engine
+  attr_reader :vin, :year, :make, :model, :engine, :registration_date, :registration_fee, :plate_type
 
   def initialize(vehicle_details)
     @vin = vehicle_details[:vin]
     @year = vehicle_details[:year]
     @make = vehicle_details[:make]
     @model = vehicle_details[:model]
     @engine = vehicle_details[:engine]
+    @plate_type = []
```
memcmahon, line 10: Since a vehicle only ever has one plate type, I think I would have set this with a default value of `nil` and then set the value to `:ev` instead of `[:ev]`

lib/vehicle.rb
```diff
@@ -1,25 +1,21 @@
...
   end
 
   def antique?
-    Date.today.year - @year > 25
+    @year <= 1999
```
memcmahon, line 14: If this program exists into 2025, this method will no longer be accurate!

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+        expect(factory.wa_ev_vehicles).to be_an(Array)
+    end
+
+    it 'creates an instance of Vehicle with attributes' do
+        factory = VehicleFactory.new
+        wa_ev_registrations = DmvDataService.new.wa_ev_registrations
+        factory.create_vehicles(wa_ev_registrations)
+        
+        vehicle = factory.wa_ev_vehicles.first
```
memcmahon, line 19: This doesn't _exactly_ match the interaction pattern, because we wanted the method `create_vehicles` to return the new instances of the vehicle class, but it's close enough for this project :)

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+        co_dmv_facilities = DmvDataService.new.co_dmv_office_locations
+        factory.new_facility(co_dmv_facilities)
+
+        facility = factory.co_dmv_facilities.first
+
+        expect(facility).to be_a(Facility)
+        
+        expect(facility.name).to eq('DMV Tremont Branch')
+        expect(facility.address).to eq('2855 Tremont Place')
+        expect(facility.phone).to eq('(720) 865-4600')
```
memcmahon, line 25: great expectations to confirm this facility is created appropriately!

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+RSpec.describe FacilityFactory do
+    
+    it 'can create an array of facilities' do
+        factory = FacilityFactory.new
+        co_dmv_facilities = DmvDataService.new.co_dmv_office_locations
+        factory.new_facility(co_dmv_facilities)
+
+        expect(factory.co_dmv_facilities).to be_an(Array)
+    end
+
```
memcmahon, line 13: You are missing tests to confirm that NY and MO facilities are created correctly!

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    it 'can create an instance of a Facility with attributes' do
+        factory = FacilityFactory.new
+        co_dmv_facilities = DmvDataService.new.co_dmv_office_locations
+        factory.new_facility(co_dmv_facilities)
+
+        facility = factory.co_dmv_facilities.first
```
memcmahon, line 19: It looks like this test might not run, because the method is named `new_co_dmv_facilities` 😢

PR #569, opened 2024-10-22
Jackson Bickler

JacksonBick:


lib/facility.rb
```diff
@@ -1,14 +1,76 @@
...
-  def initialize(name, address, phone)
-    @name = name
-    @address = address
-    @phone = phone
+  def initialize(data)
+    @name = data[:name]
+    @address = data[:address]
+    @phone = data[:phone]
     @services = []
+    data = {name:, address:, phone:}
```
memcmahon, line 12: I don't think you are ever using this variable, so you could take out this line

spec/facility_spec.rb
```diff
@@ -23,4 +24,108 @@
       expect(@facility.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
     end
   end
+
+  describe 'register_vehicle' do
```
memcmahon, line 28: Nice job writing thorough tests for the register vehicle method!

spec/facility_spec.rb
```diff
@@ -23,4 +24,108 @@
...
+      @facility.add_service('Vehicle Registration')
+      cruz = Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet',model: 'Cruz', engine: :ice} )
+      bolt = Vehicle.new({vin: '987654321abcdefgh', year: 2019, make: 'Chevrolet', model: 'Bolt', engine: :ev} )
+      @facility.register_vehicle(cruz)
+      @facility.register_vehicle(bolt)
+      expect(cruz.plate_type).to eq(:regular)
+      expect(bolt.plate_type).to eq(:electric_vehicle)
+    end
+  end
+
```
memcmahon, line 71: Since you have a helper method for `determine_plate_type_collect_fees` I would expect to see a unit test for that method as well as the integration tests for registering a vehicle.

lib/facility_factory.rb
```diff
@@ -0,0 +1,37 @@
+require './lib/facility'
+
+class FacilityFactory
+
```
memcmahon, line 4: Great job figuring out a way to deal with differences in the faciilty data based on location!

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,16 @@
+require './lib/vehicle'
+
+class VehicleFactory
+
+  def create_vehicles(registrations)
+    registrations.map do |data|
```
memcmahon, line 6: perfect use of `.map`!

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,50 @@
+require 'spec_helper'
+require 'rspec'
+
+RSpec.describe FacilityFactory do
```
memcmahon, line 4: These tests are very thorough - love it!

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,87 @@
+require 'spec_helper'
+require 'rspec'
+ 
+RSpec.describe VehicleFactory do
+  describe "create_vehicles" do
+    it "takes an array of vehicle data and makes them into instances of vehicle class" do
```
memcmahon, line 6: NICE - I love that you are using a smaller set of data to really verify that the vehicles that are being created are in fact the vehicles that you expect.

PR #570, opened 2024-10-22
Elysa Ward

elysableu:


spec/facility_builder_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      expect(@builder).to be_an_instance_of(FacilityBuilder)
+    end
+  end
+
+  describe '#build facility' do
+    it 'can build an array of new facility objects' do
+      expect(@co_facilities).to be_empty
+      co_facilities = @builder.build_facility_data(@co_dmv_office_locations)
+      expect(co_facilities.length).to eq(@co_dmv_office_locations.length)
+      expect(co_facilities[0..2].map(&:name)).to eq(["DMV Tremont Branch", "DMV Northeast Branch", "DMV Northwest Branch"])
```
memcmahon, line 25: Nice use of this map styntax to verify specfic instances of facilities!

spec/facility_builder_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    it 'can accept facility data from YET another data source' do
+      expect(@mo_facilities).to be_empty
+      @mo_facilities = @builder.build_facility_data(@mo_dmv_office_locations)
+      expect(@mo_facilities.length).to eq(@mo_dmv_office_locations.length)
+      expect(@mo_facilities[0..2].map(&:name)).to eq(["FERGUSON-OFFICE CLOSED UNTIL FURTHER NOTICE", "BUTLER", "CENTRAL WEST END"])
+    end
+  end
+
+  describe '#format_address'do
```
memcmahon, line 43: Great job testing these helper methods!

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+  describe '#initialize' do
+    it 'can create new VehicleFactory object' do
+      expect(@factory).to be_an_instance_of(VehicleFactory)
+    end
+  end
+
+  describe '#create_vehicles' do
+    it 'can create an array of new vehicle objects from DmvDataService' do
+      expect(@factory.created_vehicles).to be_empty
+      @factory.create_vehicles(@wa_ev_registrations)
```
memcmahon, line 17: This is a bit nit-picky, but based on the interaction pattern, I would expect that `create_vehicles` returns the list of vehicles that were created.  Not a huge deal on this project, just something to keep an eye on for future challenges!

lib/facility_builder.rb
```diff
@@ -0,0 +1,52 @@
+class FacilityBuilder
+  def build_facility_data(dmv_data)
+    facilities = []
+    dmv_data.each do |dmv_entry|
+      
+      
+      entry_name = nil
```
memcmahon, line 7: Great job figuring out how to handle data from multiple sources!  You could also move some of this logic to helper methods - you could imagine that all 50 states might have different formats, and that would make this method really difficult to read through.  That refactor might look like:

```ruby
entry_name = set_name_attribute(dmv_entry)

...

def set_name_attribute(dmv_entry)
  dmv_entry[:dmv_office] || dmv_entry[:office_name] || dmv_entry[:name]
end

File: lib/facility_builder.rb [hunk 5]
```diff
@@ -0,0 +1,52 @@
...
+      new_facility = Facility.new(name: entry_name, address: entry_address, phone: entry_phone)
+      
+      format_facility_services(new_facility, dmv_entry)
+      facilities << new_facility      
+    end
+    return facilities
+  end
+
+  def format_address(dmv_entry, source_num)
+    if source_num == 1
```
memcmahon, line 34: I would probably set this up so that instead of getting a number as the source, we would get a state code like "CO".  That would be a little more readable, especially if we end up needing to deal with all 50 states' data.

PR #571, opened 2024-10-22
Will Fox

willfox0409:


lib/facility.rb
```diff
@@ -1,14 +1,62 @@
 class Facility
-  attr_reader :name, :address, :phone, :services
+  attr_reader :name, :address, :phone, :services, :registered_vehicles, :hours, :address_id
+  attr_accessor :collected_fees
```
memcmahon, line 3: because you are manipulating the collected fees from within this class, I don't think this needs to be an attr_accessor.

lib/facility.rb
```diff
@@ -1,14 +1,62 @@
...
-    @name = name
-    @address = address
-    @phone = phone
-    @services = []
+  def initialize(details)
+    @name = details[:name]
+    @address = details[:address]
+    @phone = details[:phone]
+    @services = details[:services] || []   
+    @hours = details[:hours] || "Hours not available"
```
memcmahon, line 10: Nice!

lib/facility.rb
```diff
@@ -1,14 +1,62 @@
...
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    return nil unless @services.include?('Vehicle Registration')
+
+    vehicle.registration_date = Date.today
+
+    if vehicle.year <= Date.today.year - 25
```
memcmahon, line 25: Don't forget about the vehicle methods! You could have called `vehicle.antique?` here ;)

lib/facility.rb
```diff
@@ -1,14 +1,62 @@
...
+    else
+      vehicle.plate_type = :regular
+      @collected_fees += 100
+    end
+
+    @registered_vehicles << vehicle
+  end
+
+  def administer_written_test(registrant)
+    return false unless @services.include?('Written Test') &&
```
memcmahon, line 40: I like the use of these guard statements to control when tests are performed.

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+RSpec.describe FacilityFactory do 
+  before(:each) do
+    @factory = FacilityFactory.new
+    @service = DmvDataService.new
+
+    @co_dmv_data = @service.co_dmv_office_locations
+    @ny_dmv_data = @service.ny_dmv_office_locations
+  end
+
```
memcmahon, line 11: Make sure that you test all methods - looks like we are missing one for `format_address` and `format_hours`

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  describe '#create_facilities' do
```
memcmahon, line 12: These are good, robust tests.  I really like that you are verifying not only that a Facility exists, but also that it is the exact facility that you expect.

lib/facility_factory.rb
```diff
@@ -0,0 +1,36 @@
+require './lib/facility'
+
+class FacilityFactory
+  def create_facilities(data)
+    data.map do |entry|
+
+    Facility.new(
+      name: entry[:dmv_office] || entry[:office_name],
```
memcmahon, line 8: Great job getting the data from different sources to work! One thing you could think about is moving some of this data manipulation to a helper method.  You could imagine that all 50 states may have different data formats and that could make this method really overwhelming.  That refactor could look like:

```ruby
name: set_name_attribure(entry)

...

def set_name_attribute(facility)
  facility[:name] || facility[:dmv_office] || facility[:office_name] #etc...
end
```

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,15 @@
+ require './lib/vehicle'
+ 
+ class VehicleFactory   
+    def create_vehicles(data)
+        data.map do |entry|
```
memcmahon, line 5: great use of .map

lib/facility_factory.rb
```diff
@@ -0,0 +1,36 @@
+require './lib/facility'
+
+class FacilityFactory
```
memcmahon, line 3: Though you didn't get to the third dataset, you've shown enough that I am confident that you could add additional datasets, so I'm going to consider this as iteration3 finished.

PR #573, opened 2024-10-23
Coral Ruschak

Coralruschak:


lib/facility.rb
```diff
@@ -1,14 +1,34 @@
...
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    if @services.include?('Vehicle Registration')
+      vehicle.set_plate_type
+      vehicle.collect_fees
```
memcmahon, line 20: This method does not exist on a vehicle - it exists in this class so I would expect something like `collect_fees(vehicle)`

lib/facility.rb
```diff
@@ -1,14 +1,34 @@
...
+      @registered_vehicles << vehicle
+    end
+  end
+
+  def collect_fees
+    if vehicle.plate_type == :antique
```
memcmahon, line 26: there is no vehicle that exists in this method.  If you want a vehicle in the method, you need to pass it in as an argument.

lib/vehicle.rb
```diff
@@ -22,4 +25,15 @@ def antique?
   def electric_vehicle?
     @engine == :ev
   end
+
+  def set_plate_type
```
memcmahon, line 29: I like this as a helper method!

spec/vehicle_spec.rb
```diff
@@ -33,4 +33,20 @@
       expect(@camaro.electric_vehicle?).to eq(false)
     end
   end
+
+  describe 'set_plate_type' do
```
memcmahon, line 37: great job writing a test for this method that you added!

PR #574, opened 2024-10-23
Iteration 3 features Pull Request # 3

qoryhanisagal:


lib/facility_factory.rb
```diff
@@ -0,0 +1,26 @@
...
+  # It maps over the data, creating and returning an array of Facility objects based on the data.
+  #
+  # Parameters:
+  # - data: An array of hashes. Each hash contains information about a facility (name, address, phone, etc.).
+  #
+  # Returns:
+  # - An array of Facility objects.
+  def create_facilities(data)
+    # Iterate over each entry (facility_data) in the provided data array
+    data.map do |facility_data|
```
memcmahon, line 14: great use of .map

lib/facility_factory.rb
```diff
@@ -0,0 +1,26 @@
...
+      # Create and return a new instance of Facility using the data from the current hash (facility_data).
+      Facility.new({
```
memcmahon, line 16: This will work for data from the CO dataset, but will not work as expected for NY or MO.  Those datasets have different formats.

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,49 @@
+# Required files for testing
+require 'spec_helper'         # Load RSpec configuration and helpers
+require './lib/facility_factory'  # Load the FacilityFactory class
+require './lib/facility'          # Load the Facility class
+
+# Describe the test suite for the FacilityFactory class
+RSpec.describe FacilityFactory do
```
memcmahon, line 7: These are great tests for the colorado "style" of data - very thorough!  I also like that you are using a sample of data, instead of the entire dataset. Just keep an eye out that different states have differently formatted data.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+    registrant_2.earn_permit
+
+    # Test if the permit status is updated
+    expect(registrant_2.permit?).to eq(true)
+  end
+
+  # Edge Case 1: Registrants of any age can earn a permit
+  it 'allows registrants of any age to earn a permit' do
+    registrant_young = Registrant.new('Baby Bruce', 5)
+    registrant_old = Registrant.new('Grandpa Joe', 85)
```
memcmahon, line 40: Great naming for these variables - it makes it super clear what you are testing!

PR #576, opened 2024-10-23
Shadeau Christensen

shadeauchristensen:
- removed "hard code" found new methods to cut code blocks down
-rspec passed all tests, removed tests that were unused and/or lacking
-rewrote entire second half of it2 & finished it3
- removed all if not most seelf comments to avoid confusion of written tests
-used comments made to reconfigure tests to make more sense


lib/vehicle.rb
```diff
@@ -1,25 +1,43 @@
...
     @model = vehicle_details[:model]
     @engine = vehicle_details[:engine]
+    @plate_type = nil
+    @registration_date = nil
   end
 
+  
   def antique?
-    Date.today.year - @year > 25
+    @year < 1995
```
memcmahon, line 23: This math isn't mathing - 1995 was 29 years ago....

lib/vehicle.rb
```diff
@@ -1,25 +1,43 @@
...
-
+  
   def electric_vehicle?
     @engine == :ev
   end
+
+  def assign_plate_type
+    if @year < Date.today.year - 25
+      @plate_type = :antique
+    elsif @engine == :ev
```
memcmahon, line 33: this works, you could also use `elsif electric_vehicle?`

lib/facility.rb
```diff
@@ -1,14 +1,89 @@
...
+      true
+    else
+      false
+    end
+  end
+
+  def registered_vehicle(vehicle)
+    if @services.include?('Vehicle Registration')
+      vehicle.assign_registration_date
+      vehicle.assign_plate_type
```
memcmahon, line 53: yay helper methods!

lib/facility.rb
```diff
@@ -1,14 +1,89 @@
...
+  def vehicle_age(vehicle)
+    Date.today.year - vehicle.year
+  end
+
+  def calculate_fees(vehicle)
+    registration_cost(vehicle.plate_type, vehicle_age(vehicle))
+  end
+end
+
+class FacilityFactory
```
memcmahon, line 80: this class should be in its own file!!

lib/facility.rb
memcmahon: But make sure that you test **every** method - the vehicle tests are missing for these methods!!

lib/facility.rb
```diff
@@ -1,14 +1,89 @@
...
+  def self.create_facilities
```
memcmahon, line 81: This doesn't quite follow the interaction pattern, but it does successfully create facilities from the CO data.  It would not work though for the NY or MO data.

PR #588, opened 2024-12-14
Jaren Garman

JarenGarman:


lib/dmv.rb
```diff
@@ -9,8 +13,30 @@ def add_facility(facility)
...
+
+    @facilities.map(&:registered_vehicles).flatten.count { |vehicle| vehicle.year == year }
+  end
+
+  def most_popular_county
+    return if @facilities.map(&:registered_vehicles).empty?
+
+    @facilities.map(&:registered_vehicles).flatten.group_by(&:county).max_by do |_county, vehicle_array|
+      vehicle_array.length
+    end[0]
```
juliet-e, line 40: Nice use of chained enumerables! As a complimentary exercise, it's a good idea to practice talking through these solutions as they are more complex, and describing your thought process and approach to a solution can be great interview practice!

lib/facility_creator.rb
```diff
@@ -0,0 +1,62 @@
...
+      end.join(' ')
+    elsif facility.include?(:dmv_office)
+      facility[:dmv_office]
+    elsif facility.include?(:name)
+      facility[:name]
+    end
+  end
+
+  def parse_address(facility)
+    "#{unless facility[:street_address_line_1].nil?
```
juliet-e, line 30: This block of code is not very readable. I would much rather a more multi-line process that extracts each component as a variable and then maybe joins them at the end or uses string interpolation to combine the data together. Anytime you're using `\` to join long strings together, it generally could indicate unreadable string manipulation

PR #590, opened 2024-12-16
Rig Freyr

ontruster74:
This is my submission for Solo Project #2: the DMV. This repo contains completed versions of iterations 1-3 with full spec compliance. I also completed Option 1 of Iteration 4 and submitted a pull request to its respective repo. As always, please let me know if I am missing anything!


lib/facility.rb
```diff
@@ -1,14 +1,86 @@
 class Facility
-  attr_reader :name, :address, :phone, :services
+  attr_reader :name, :address, :phone, :services, :collected_fees, :registered_vehicles
 
-  def initialize(name, address, phone)
-    @name = name
-    @address = address
-    @phone = phone
+  def initialize(facility_details)
+    if facility_details.keys.include?(:dmv_office)
```
juliet-e, line 5: consider extracting these chunks of data-cleaning logic into helper methods that can be called from the initialize. We generally want to keep methods pretty short (8-10) lines so when methods move beyond that, using helper methods can be a great resource! For example, your initialize could be cleaned up like this:

```
def initialize(facility_details)
    @name = populate_name(facility_details)
    @address = populate_address(facility_details)
    ...
end

def populate_name(data)
    if data.keys.include?(:dmv_office)
         return facility_details[:dmv_office]
    etc...
end

    
```

lib/facility.rb
```diff
@@ -1,14 +1,86 @@
...
+
+  def renew_drivers_license(registrant)
+    if services.include?('Renew Drivers License')
+      if registrant.license_data[:license] == true
+        registrant.license_data[:renewed] = true
+      end
+    end
+    return registrant.license_data[:renewed]
+  end
+
```
juliet-e, line 85: Nice attention to complex logical conditions!

lib/vehicle.rb
```diff
@@ -6,13 +6,25 @@ class Vehicle
               :make,
               :model,
               :engine
+              
+  attr_accessor :plate_type, :registration_date
 
   def initialize(vehicle_details)
-    @vin = vehicle_details[:vin]
-    @year = vehicle_details[:year]
+    if vehicle_details.keys.include?(:vin_1_10)
```
juliet-e, line 13: same goes here -- we want our initialize especially to remain pretty simple, so extracting if statements/logic into helper methods is a great option for simplifying this (see Facility note for more examples)

lib/vehicle.rb
```diff
@@ -6,13 +6,25 @@ class Vehicle
               :make,
               :model,
               :engine
+              
+  attr_accessor :plate_type, :registration_date
```
juliet-e, line 10: For the purposes of this project, it is totally okay to use an attr_accessor. However, we often want to stay away from these and instead create methods for changing the value of our attributes. We want to be able to have control over these data changes, and maintain data integrity. Right now, someone could change the registration_date to banana and we have no control over that. But, if you added a "setter" method in this class that an outside class had to call, like set_registration_date(date) then you could add logic like checking that it's a valid date. This is a good rule of thumb for adhering to the principle of encapsulation in our our classes.

lib/facility.rb
ontruster74: That makes sense!

lib/facility.rb
ontruster74: Thanks! :)

spec/facility_factory_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+describe FacilityFactory do
+  describe '#create_facilities' do
+    it 'can create facilities from external data (Colorado)' do
+      factory = FacilityFactory.new
+      co_dmv_office_locations = DmvDataService.new.co_dmv_office_locations
+      colorado_facilities = factory.create_facilities(co_dmv_office_locations)
+
+      facility_zero = colorado_facilities[0]
+      expect(facility_zero).to be_a(Facility)
+      expect(facility_zero.name).to eq("DMV Tremont Branch")
```
juliet-e, line 13: These are very robust tests! The only risk here is that the facility data is coming from an external service (via the DmvDataService) so there's a slight chance that this data could change/update and cause your tests to fail even if there's nothing wrong with your code or the way you're handling their data. This is a risk that is *totally* fine for the project, but as you progress in Turing (especially Mod 3) you'll start to think about other options when you're testing something that uses a data set we can't control. Two other options are:
1. using more general assertions (i.e. expect(facility_zero.name).to be_a String)
2. using mock (dummy) data instead of actually getting data from the DmvDataService

This is food for thought for as you progress through the program. For now, these are great tests!

lib/vehicle.rb
ontruster74: Right, I’ll keep that in mind for next time. I’m very used to long elsif chains/switch-case statements but the helper methods would look much cleaner.

lib/vehicle.rb
ontruster74: Understood. I’ll err on the side of readers over accessors in the future.

spec/facility_factory_spec.rb
ontruster74: Thanks! I had been wondering how to deal with external database changes, but this clears some of that up for me for sure.

PR #598, opened 2024-12-17
Beverly Green - the_dmv

bevgreen:


lib/dmv_facilities.rb
```diff
@@ -0,0 +1,19 @@
+class DMVFacilities
```
memcmahon, line 1: I think it might make more sense to call this class `DMVFactory` to match the Vehicle Factory naming convention

lib/registrant.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    def permit?
+        @permit = permit
+    end
+
+    def earn_permit
+        @permit = true
+    end
+
+    def change_license_data(key)
```
memcmahon, line 26: Love this helper method! Just make sure that you are writing tests for any methods that you add 😉

lib/facility.rb
```diff
@@ -1,14 +1,73 @@
...
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+
+    if @services.include?('Vehicle Registration')
+      @registered_vehicles << vehicle
```
memcmahon, line 25: When a vehicle is registered, you probably also want to call the registration method on the vehicle so that the plate type is set and the registration date is updated!

PR #599, opened 2024-12-17
Jacob Inman The DMV Solo Proj 2

jinman14:


lib/facility.rb
```diff
@@ -1,14 +1,74 @@
...
+      end
+    end
+
+    @collected_fees
+  end
+
+  def administer_written_test(registrant)
+
+    if registrant.age >= 16 && registrant.permit? && @services.include?("Written Test")
+      registrant.license_data[:written] = true
```
memcmahon, line 53: This definitely works, but it would be a little better if there was a method on the registrant class that managed this change. It could either be explicitly exposing license_data as an attr_writer or attr_accessor, or you could create a method to `earn_written_license` that would change the value from inside the registrant.

lib/dmv_facilities.rb
```diff
@@ -0,0 +1,30 @@
+class DMVFacility
```
memcmahon, line 1: I think it might make more sense to call this class `FacilityFactory` to match the Vehicle Factory naming convention

spec/facility_spec.rb
```diff
@@ -2,25 +2,183 @@
 
 RSpec.describe Facility do
```
memcmahon, line 3: Great job testing all of the functionality - I think you _could_ combine some of your `it` blocks related to registering a vehicle (only if you wanted to make these files more succinct :) )

PR #601, opened 2024-12-17
Completed project 2 pull request

ChasePhillips89:
This pull request includes the entirety of iterations one, two, and three. All spec files and class files to include proper methods are included, and tests are functioning and passing.


lib/dmvfacility.rb
```diff
@@ -0,0 +1,22 @@
+class DMVFacility
```
memcmahon, line 1: I think it might make more sense to call this class `DMVFactory` or `FacilityFactory` to match the Vehicle Factory naming convention

spec/vehicle_spec.rb
```diff
@@ -33,4 +34,5 @@
       expect(@camaro.electric_vehicle?).to eq(false)
     end
   end
-end
+
```
memcmahon, line 37: You are missing a test for `plate_type`!

spec/vehicle_spec.rb
```diff
@@ -1,11 +1,12 @@
...
 
 RSpec.describe Vehicle do
   before(:each) do
     @cruz = Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice} )
     @bolt = Vehicle.new({vin: '987654321abcdefgh', year: 2019, make: 'Chevrolet', model: 'Bolt', engine: :ev} )
     @camaro = Vehicle.new({vin: '1a2b3c4d5e6f', year: 1969, make: 'Chevrolet', model: 'Camaro', engine: :ice} )
   end
+
   describe '#initialize' do
     it 'can initialize' do
```
memcmahon, line 11: Since you made the attributes of a vehicle attr_accessors, you would also want to test that functionality.  For example, you want to have a test for the method `registration_date =`

lib/facility.rb
```diff
@@ -1,14 +1,69 @@
...
+    if registrant.license_data[:written]
+      registrant.license_data[:license] = true 
+    else
+      "Registrant must pass the written test before taking the road test!!!"
+    end
+  end
+
+  def renew_drivers_license(registrant)
+    
+    return false unless services.include?('Renew License')
```
memcmahon, line 61: Nice use of guard statements!

PR #602, opened 2024-12-17
Mark Kendall

mkendall42:
Project is complete, iterations 1-4 implemented (option #3 chosen for iteration 4).
All RSpec test files working, tests pass as of submission.

### DIFF
--- a/lib/dmv.rb
+++ b/lib/dmv.rb
@@ -24,15 +24,7 @@ class Dmv
   def create_state_facilities(state, facilities_incoming_data)
     #Each state may have different API, and so we need a 'dispatching' method here
     #FOR NOW: just code this for Colorado, then can break it out from there for addt'l states
-    
-    # #Need to extract the following information:
-    # @name = facility_info[:name]
-    # @address = facility_info[:address]    #Gotta concatenate all of this too...ugh...
-    # @phone = facility_info[:phone]
-    # @services = []
-    # #Added in iteration 2
-    # @collected_fees = 0
-    # @registered_vehicles = []
+    #Also coded for NY and MO (that's all for this project)
     if state == "Colorado"
       facilities_incoming_data.each do |facility|
         facility_info = {
@@ -106,12 +98,6 @@ class Dmv
 
     if state == "Missouri"
       facilities_incoming_data.each do |facility|
-        #Pre-determine name, since this one is nasty, have to concatenate and then 'fix' capitlization the hard way (since each word is capitalized for a title usually)
-        
-        #Fuck this, I might write a helper method to just capitalize each word.
-        #I guess it will live in this class since it's not used anywhere else for this project
-        #(a case where OOO is kinda weird / less 'aligned')
-        #This could be tricky, though, because it needs to accept an arbitrary number of arguments (maybe pass the hash keys as symbols?).  Don't know how to do this yet...
         address_formatted = facility[:address1].delete_suffix(",")      #Certain (random?) entries end with it for some reason
         address_formatted = "#{address_formatted} #{facility[:city]} #{facility[:state]} #{facility[:zipcode]}"
 
--- a/lib/dmv.rb
+++ b/lib/dmv.rb
@@ -22,10 +22,6 @@ class Dmv
   #idiosyncracies to procedures and different dataset formats for different states.
 
   def create_state_facilities(state, facilities_incoming_data)
-    #Each state may have different API, and so we need a 'dispatching' method here
-    #FOR NOW: just code this for Colorado, then can break it out from there for addt'l states
-    #Also coded for NY and MO (that's all for this project)
-
     facilities_incoming_data.each do |facility_data|
       #Construct facility based on state, since dataset must be converted differently due to idiosyncracies in each case
       if state == "Colorado"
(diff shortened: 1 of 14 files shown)


lib/dmv.rb
```diff
@@ -9,8 +13,166 @@ def add_facility(facility)
...
+      end
+    end
+  end
+
+  def get_ev_registration_analytics(state, specified_year)
+    #Determine most popular make/model registered, the # registered for specified year, and the county with most registered vehicles
+    #NOTE: I kept method naming based on EV vehicles.  However, now that I've extended functionality to allow NY vehicles (later in iteration 4), this method is more general,
+    #and can analyze general NY vehicles.
+
+    vehicle_tally = {}
```
juliet-e, line 55: Consider breaking up some of this logic into well-named helper methods in order to shorten this method. We typically want methods to stay between 8-10 lines, so using a helper method like tally_vehicle could be a great way to extract some of this into a separate method.

lib/dmv.rb
```diff
@@ -9,8 +13,166 @@ def add_facility(facility)
...
+    vehicle_year_total = 0
+    @facilities.each do |facility|
+      if facility.state == state
+        vehicle_year_total += facility.registered_vehicles.count do |vehicle|
+          vehicle.registration_date.year == specified_year
+        end
+      end
+    end
+
+    #Find county which has the most vehicle registrations.  Note high machinery overlap with most popular model...a way to combine?
```
juliet-e, line 80: I love seeing your note around combining! Extracting a helper method that just uses a different keyword value to look for, and updates the correct key in the hash, could do the trick!

lib/dmv.rb
```diff
@@ -9,8 +13,166 @@ def add_facility(facility)
...
+    address_formatted = facility_ny_data[:street_address_line_1].split.map { |word| word.capitalize }
+    address_formatted << facility_ny_data[:city].split.map { |word| word.capitalize }
+    address_formatted << facility_ny_data[:state]
+    address_formatted << facility_ny_data[:zip_code]
+    address_formatted = address_formatted.join(" ")
+
+    #Hours are listed for each day by opening and closing.  For now, I'm going to brute-force this,
+    #since no exact format is expected.  This is a mess though:
+    #This is gross...some facilities don't have keys for all M-F.  So I need to run ifs now (otherwise problems with nil)
+    hours_formatted = ""
```
juliet-e, line 125: This data cleaning does get a bit icky. That's why extracting into a helper method that can be called such as `format_opening_times` could at least encapsulate the ickyness.

lib/facility.rb
```diff
@@ -1,14 +1,66 @@
...
+
+  def administer_road_test(registrant)
+    #Of course, this assumes the registrant actually passes the test (happy path)
+    registrant.license_data[:license] = include?("Road Test") && registrant.license_data[:written] == true
+  end
+
+  def renew_drivers_license(registrant)
+    registrant.license_data[:renewed] = include?("Renew License") && registrant.license_data[:license] == true
+  end
+
```
juliet-e, line 65: Nice logic here!

lib/vehicle.rb
```diff
@@ -5,14 +5,21 @@ class Vehicle
               :year,
               :make,
               :model,
-              :engine
+              :engine,
+              :registration_county
+  attr_accessor :registration_date, :plate_type
```
juliet-e, line 10: For the purposes of this project, it is totally okay to use an attr_accessor. However, we often want to stay away from these and instead create methods for changing the value of our attributes. We want to be able to have control over these data changes, and maintain data integrity. Right now, someone could change the registration_date to `banana` and we have no control over that. But, if you added a "setter" method in this class that an outside class had to call, like `set_registration_date(date)` then you could add logic like checking that it's a valid date. This is a good rule of thumb for adhering to the principle of encapsulation in our our classes.

spec/dmv_spec.rb
```diff
@@ -40,4 +43,183 @@
...
+      expect(@dmv.facilities[3].hours).to be_a(String)
+    end
+
+    it 'correctly create facilities array for Missouri based on API data' do
+      missouri_facilities = DmvDataService.new.mo_dmv_office_locations()
+      @dmv.create_state_facilities("Missouri", missouri_facilities)
+
+      # binding.pry
+      
+      #NOTE: this dataset list actually changed on me after I made tests; hence why I've moved to more 'permanent' testable aspects
```
juliet-e, line 91: Great observation. We'll talk more about this when you get to Mod 3, but when we're using external data sources we can either:
1. Make somewhat flexible assertions or only assert on the permanent aspects OR
2. Use some mocked data instead of calling the DMVDataService. You could, for example, create a hash in this test that includes 3 facilities as they appear in the MO dataset, and that is called a "stub". You can then test for exact equivalence of different attributes once the Facility objects are created. Your "fake" facilities list hash would replace line 86, and it would be that stubbed data you pass into the create_state_facilities method

I'd argue your tests here are just fine, but sometimes we want to be able to make more robust assertions so it's helpful to use stubbed data and control exactly what those attributes will be

spec/facility_spec.rb
```diff
@@ -2,25 +2,207 @@
...
+
+      expect(@registrant_1.license_data).to eq({written: false, license: false, renewed: false})
+
+      @facility_1.administer_written_test(@registrant_1)
+      @facility_1.administer_road_test(@registrant_1)
+      @facility_1.renew_drivers_license(@registrant_1)
+
+      max_upgrades_hash = {written: true, license: true, renewed: true}
+      expect(@registrant_1.license_data).to eq(max_upgrades_hash)
+    end
```
juliet-e, line 205: Excellent robust testing!

PR #603, opened 2024-12-17
Joseph Bloom

Jsphbloom:


lib/dmv_facilities.rb
```diff
@@ -0,0 +1,19 @@
+require './lib/facility'
+require './lib/dmv_data_service'
+
+
+class DmvFacilities
```
memcmahon, line 5: I think it might make more sense to call this class `DMVFactory` to match the Vehicle Factory naming convention

lib/facility.rb
```diff
@@ -1,14 +1,59 @@
...
+      @registered_vehicles << vehicle
+    else
+      @collected_fees += 100
+      @registered_vehicles << vehicle
+    end
+  end
+
+  def administer_written_test(registrant)
+    if @services.include?("Written Test") && registrant.permit? == true
+      registrant.license_data[:written] = true
```
memcmahon, line 41: This definitely works, but it would be a little better if there was a method on the registrant class that managed this change.  It could either be explicitly exposing `license_data` as an attr_writer or attr_accessor, or you could create a method to `earn_written_license` that would change the value from inside the registrant.

lib/facility.rb
```diff
@@ -1,14 +1,59 @@
 class Facility
   attr_reader :name, :address, :phone, :services
 
-  def initialize(name, address, phone)
-    @name = name
-    @address = address
-    @phone = phone
+  def initialize(facility_details)
+    @name = facility_details[:name] || facility_details[:dmv_office] || facility_details[:office_name]
```
memcmahon, line 5: Nice use of this OR logic 😉

spec/facility_spec.rb
```diff
@@ -23,4 +26,82 @@
       expect(@facility.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
     end
   end
+
+  describe '#register vehicle' do
+    it 'can register a vehicle' do
+      @facility.register_vehicle(@cruz)
+      expect(@facility.registered_vehicles).to eq([@cruz])
```
memcmahon, line 33: You also want to make sure that the registration date is set when this method is called!

lib/dmv_facilities.rb
Jsphbloom: I agree ^

lib/facility.rb
Jsphbloom: That makes sense! i thought it was a little weird that I was handling this from facility.

lib/facility.rb
Jsphbloom: Thank you B)

spec/facility_spec.rb
Jsphbloom: Ah yes gotcha! Missed that!

PR #605, opened 2024-12-17
Sebastian the_dmv

0nehundr3d:


lib/facility.rb
```diff
@@ -1,14 +1,63 @@
...
+
+    @registered_vehicles << vehicle
+  end
+
+  def administer_written_test(registrant)
+    return false if !services.include?("Written Test")
+    return false if !registrant.permit?
+    return false if registrant.age < 16
+
+    registrant.update_license_data(:written)
```
juliet-e, line 44: Great delegation to the class that is the best candidate to handle this logic! This is a great example of using encapsulation in your code. Same with delegating `register` to vehicle on line 25 :)

lib/registrant.rb
```diff
@@ -0,0 +1,26 @@
...
+    def permit?
+        return @permit
+    end
+
+    def earn_permit
+        @permit = true
+    end
+
+    def update_license_data(key)
+        @license_data[key] = true
```
juliet-e, line 24: Clever reduction of duplication with this method! Very flexible and compact!

lib/vehicle_factory.rb
```diff
@@ -0,0 +1,16 @@
+class VehicleFactory
+    def create_vehicles(dataset)
+        vehicles = []
+        dataset.each do |vehicle|
+            new_vehicle = Vehicle.new({
+                :vin=>vehicle[:vin_1_10],
+                :year=>vehicle[:model_year],
+                :make=>vehicle[:make],
+                :model=>vehicle[:model],
+                :engine=>vehicle.include?(:electric_range) ? :ev : :ice
```
juliet-e, line 10: Nice attention to detail with this difference

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+        it "can create vehicles" do
+            wa_ev_registrations = DmvDataService.new.wa_ev_registrations
+            vehicles = @factory.create_vehicles(wa_ev_registrations)
+
+            expect(vehicles[0]).to be_a(Vehicle)
+            expect(vehicles[0].engine).to eq(:ev)
+            expect(vehicles[0].make).to eq("BMW")
+            expect(vehicles[0].model).to eq("i3")
+            expect(vehicles[0].vin).to eq("WBY8P8C51K")
+            expect(vehicles[0].year).to eq("2019")
```
juliet-e, line 26: These are robust tests! The only risk here is that the facility data is coming from an external service (via the DmvDataService) so there's a slight chance that this data could change/update and cause your tests to fail even if there's nothing wrong with your code or the way you're handling their data. This is a risk that is totally fine for the project, but as you progress in Turing (especially Mod 3) you'll start to think about other options when you're testing something that uses a data set we can't control. Two other options are:

 1. using more general assertions (i.e. expect(facility_zero.name).to be_a String)
 2.  using mock (dummy) data instead of actually getting data from the DmvDataService

This is food for thought for as you progress through the program. For now, these are great tests!

PR #608, opened 2024-12-18
Patrick_Little

patchy-pants:
Javascript Iteration 4 https://github.com/little-Patrick/the_dmv_extension_js


lib/facility.rb
```diff
@@ -1,14 +1,60 @@
...
+      
+      car
+
+    else
+      nil
+    end
+  end
+
+  def administer_written_test(registrant)
+    services.include?('Written Test') && registrant.age >= 16 ? registrant.license_data[:written] = true : false
```
memcmahon, line 50: This definitely works, but it would be a little better if there was a method on the registrant class that managed this change. It could either be explicitly exposing license_data as an attr_writer or attr_accessor, or you could create a method to `earn_written_license` that would change the value from inside the registrant.

lib/store_facilities.rb
```diff
@@ -0,0 +1,35 @@
+
+
+class StoreFacilities
```
memcmahon, line 3: I think it might make more sense to call this class `FacilityFactory` to match the Vehicle Factory naming convention

spec/store_spec.rb
```diff
@@ -0,0 +1,37 @@
+require './spec/spec_helper'
+
```
memcmahon, line 2: For readability, you want to name this to match the class it is testing, so it should be `store_facilities_spec.rb`

PR #609, opened 2024-12-18
Logan Sauer

ldsauer:
Here is my project. Sorry I did not get nearly as far as I wanted to again. I was very confident last night that I would at least finish up iteration 2 and make all my tests pass but with my husband getting in the accident and this being due my stress levels went through the roof and I had a little menty b lol. I know that I can do better and I am really looking forward to this winter break to really buckle down and go through all our past lessons and really up my confidence and show that I can do this and know what I am doing!!


spec/registrant_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+            registrant_2 = Registrant.new('Penny', 15)
+            
+            expect(registrant_2.permit?).to eq(false)
+            
+            registrant_2.earn_permit
+            
+            expect(registrant_2.permit?).to eq(true)
+        end
+
+
```
memcmahon, line 67: Make sure that you add tests for all of the methods that you create - we are missing tests for `take_written_test` `administer_road_test` and `renew_drivers_license`

lib/facility.rb
```diff
@@ -1,14 +1,74 @@
...
+      #renew license
+    else
+      false
+    end
+  end
+
+
+# Helper methods
+
+  def determine_plate_type_and_fees(vehicle)
```
memcmahon, line 62: I really like this as a helper method!  Good separation of responsibilities.

PR #610, opened 2024-12-18
Will Fox: DMV

willfox0409:
All three iterations complete, all tests passing, and I made time for some refactoring to improve readability and efficiency!


lib/facility.rb
```diff
@@ -1,14 +1,63 @@
...
+        @collected_fees += 25
+      else
+        @collected_fees += 100
+      end
+    end
+  end
+
+  def administer_written_test(registrant)
+    if @services.include?('Written Test') && registrant.age >= 16 && registrant.permit?
+      registrant.license_data[:written] = true
```
memcmahon, line 42: This definitely works, but it would be a little better if there was a method on the registrant class that managed this change. It could either be explicitly exposing license_data as an attr_writer or attr_accessor, or you could create a method to `earn_written_license` that would change the value from inside the registrant.

lib/facility_factory.rb
```diff
@@ -0,0 +1,48 @@
...
+                name = entry[:dmv_office]
+                address = entry[:address_li]
+                phone = entry[:phone]
+
+                facility = Facility.new({
+                    name: name,
+                    address: address,
+                    phone: phone
+                })
+                facilities << facility
```
memcmahon, line 17: These lines are repeated 3 times - this is a good opportunity to refactor into a helper method!

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,46 @@
+require 'spec_helper'
+
+RSpec.describe VehicleFactory do 
+    before(:each) do
+        @factory = VehicleFactory.new
+        @mock_data = [
```
memcmahon, line 6: Love the use of mock data here!  You'll learn more about mocking API data in future mods, and this is a great instinct to start playing around with!

PR #611, opened 2024-12-18
Donte Handy

dontehandy:


dmv_ds_ip.rb
```diff
@@ -0,0 +1,120 @@
+require 'pry'
```
memcmahon, line 1: These automated scripts are really cool - a fun programming challenge, and you did a nice job with them.  However, this is a little redundant because we are using TDD - your tests are also showing what is or is not working in your application without having to write these scripts.

lib/facility.rb
```diff
@@ -1,14 +1,161 @@
...
+  end
+
+  #the above methods are business logic to administer the written test, road test, and renew a driver's license
+  #using conditional statements to determine if the services are available and if the registrant meets the requirements
+  #setting the keys in the license_data hash to true based on the services provided
+  #returning true if the services are available and the requirements are met
+  #returning false if the services are not available or the requirements are not met
+  #using the include? method to check if the services array includes a specific string
+
+  def self.create_from_data(data, state) #class method to create facility objects from data and state
```
memcmahon, line 87: This is a great use-case for a class method - to create instances of the class.  For this project, we did want to use the Factory pattern to do this work - similar to the Vehicle Factory in iteration 2!

spec/facility_spec.rb
```diff
@@ -1,26 +1,59 @@
...
+  describe '#renew_drivers_license' do
+    it 'can renew a driver\'s license' do
+      @facility.administer_written_test(@registrant) #calling the administer_written_test method on the @facility variable and passing in the @registrant variable as an argument
+      @facility.administer_road_test(@registrant) #calling the administer_road_test method on the @facility variable and passing in the @registrant variable as an argument
+      expect(@facility.renew_drivers_license(@registrant)).to be true #calling the renew_drivers_license method on the @facility variable and passing in the @registrant variable as an argument
+      expect(@registrant.license_data[:renewed]).to be true #checking if the renewed key in the license_data hash of the @registrant variable is true
     end
   end
-end
+end
```
memcmahon, line 59: You are missing a test for the class method that you  wrote: `self.create_from_data`

PR #612, opened 2024-12-18
Hannah Gray- Final Pull/Commit Messages

hanbanfan:


lib/dmv.rb
```diff
@@ -1,16 +1,40 @@
+<<<<<<< HEAD
```
juliet-e, line 1: Anytime you see <<<<< and ===== symbols, those indicate the presence of a merge conflict and technically the merge conflict is not resolved until these symbols are cleared. In its current state, this class is not stable/usable until that merge conflict is resolved. Notice that this top section is the initial starter code for this class.

lib/dmv.rb
```diff
@@ -1,16 +1,40 @@
...
+  attr_reader :facilities
 
+  # Initializes the Dmv object with an empty array of facilities.
+=======
+# The Dmv class represents the DMV system that manages facilities.
+class Dmv
+  attr_reader :facilities
+
+  # Initializes an empty array to store Facility objects.
+>>>>>>> feature/vehicle-class
```
juliet-e, line 14: Here are your additions to the class. You can remove the duplication so that it's just one single class with the necessary attr_readers, initialize and methods

lib/facility.rb
```diff
@@ -1,14 +1,51 @@
+<<<<<<< HEAD
```
juliet-e, line 1: Same goes for this class --> you're still in the midst of a merge conflict here and will need to merge the code accordingly and then remove the merge conflict symbols

lib/dmv_data_service.rb
```diff
@@ -1,25 +1,33 @@
...
+  # @return: Parsed JSON data
   def load_data(source)
     response = Faraday.get(source)
     JSON.parse(response.body, symbolize_names: true)
   end
 
+  # Retrieves Washington EV Registration data
   def wa_ev_registrations
-    @wa_ev_registrations ||= load_data('https://data.wa.gov/resource/rpr4-cgyd.json')
+    load_data('https://data.wa.gov/resource/rpr4-cgyd.json')
```
juliet-e, line 16: I wouldn't expect you to need to update these instance variables. Did you have an issue with reading them from another class?

lib/dmv.rb
juliet-e: There is a really helpful VSCode tool that can allow you to visualize what a merged class looks like. If you look at this branch in VSCode, you should see an option that says "View Code in Merge Editor" or something like that, and that view will be very helpful for solving these conflicts!

lib/facility_factory.rb
```diff
@@ -0,0 +1,25 @@
...
+>>>>>>> feature/facility-class
+require_relative 'facility'
+
+# The FacilityFactory class creates Facility objects from external data sources.
+class FacilityFactory
+  # Creates an array of Facility objects from external data.
+  # @param data [Array<Hash>] An array of facility details from the API.
+  # @return [Array<Facility>] An array of Facility objects.
+  def create_facilities(data)
+    data.map do |facility_data|
```
juliet-e, line 13: This looks good!

lib/facility_factory.rb
```diff
@@ -0,0 +1,25 @@
+<<<<<<< HEAD
```
juliet-e, line 1: This merge conflict should be a relatively easier merge fix, so I'd encourage you to use the VSCode merge editing tool with this one

lib/vehicle.rb
```diff
@@ -1,25 +1,106 @@
...
+  def electric?
     @engine == :ev
+=======
+  # Register the vehicle
+  def register
+    @registration_date = Date.today
+>>>>>>> feature/facility-class
+=======
+# The VehicleFactory class creates Vehicle objects from external data sources.
+class VehicleFactory
```
juliet-e, line 89: We'd want VehicleFactory to live in a separate file

spec/facility_spec.rb
```diff
@@ -1,26 +1,92 @@
...
 
-  describe '#add service' do
-    it 'can add available services' do
-      expect(@facility.services).to eq([])
-      @facility.add_service('New Drivers License')
-      @facility.add_service('Renew Drivers License')
-      @facility.add_service('Vehicle Registration')
-      expect(@facility.services).to eq(['New Drivers License', 'Renew Drivers License', 'Vehicle Registration'])
-    end
+  it 'can register a vehicle and collect fees' do
```
juliet-e, line 81: It seems like this test was added but is not implemented. Be sure to run all your tests to make sure things are green! It's also helpful to refer back to the interaction pattern in order to prevent missing functionality

spec/vehicle_factory_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'rspec'
+require './lib/vehicle_factory'
+require './lib/vehicle'
+
+RSpec.describe VehicleFactory do
+  before(:each) do
+    @factory = VehicleFactory.new
+    @data = [
+      { vin_1_10: '123ABC4567', model_year: '2012', make: 'Toyota', model: 'Prius' },
```
juliet-e, line 9: Nice used of stubbed (fake) data here!

PR #613, opened 2024-12-18
Zion Mekonnen

zionmekonnen:


lib/vehicle.rb
```diff
@@ -5,21 +5,36 @@ class Vehicle
               :year,
               :make,
               :model,
-              :engine
+              :engine,
+              :plate_type,
+              :antique,
+              :register_date
```
memcmahon, line 11: Make sure you name your attr_readers with the same name as your instance variable that it is reading.  In this case, it should be :registration_date

spec/vehicle_spec.rb
```diff
@@ -33,4 +33,11 @@
       expect(@camaro.electric_vehicle?).to eq(false)
     end
   end
+  it 'returns nil for cruz registration date' do
+    cruz = Vehicle.new({vin: '123456789abcdefgh', year: 2012, make: 'Chevrolet', model: 'Cruz', engine: :ice})
+    
+    expect(cruz.registration_date).to eq(nil)
+  end
+
+
```
memcmahon, line 42: You want to add tests for any methods that you create.  It looks like we are missing tests for `update_registration_date` and `update_plate_type`

spec/registrant_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    it 'returns attributes registrant 2' do
+        registrant_2 = Registrant.new('Penny', 15)
+
+        expect(registrant_2.name).to eq("Penny")
+        expect(registrant_2.age).to eq(15)
+        expect(registrant_2.license_data).to eq({:written=>false, :license=>false, :renewed=>false})
+    end
+
+    it "returns if permit is earned" do
```
memcmahon, line 30: This is a great test!

PR #624, opened 2025-02-18
Iteration 3 - Stephanie Guerrero 

Sguerrero0991:
This pull request includes adds and commits to complete iteration 3. The purpose of iteration 3 was to create DMV Facility objects with data from DMV Facilities data sets. These data sets were accessed to create appropriate tests and methods in order to create new facility objects through my FacilityFactory class. I used binding.pry to access each data set and pull all necessary info for each location. All tests were written and run one location at a time.


lib/facility.rb
```diff
@@ -1,14 +1,70 @@
...
+    @collected_fees = 0
   end
 
-  def add_services(service)
+  def add_service(service)
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    if @services.include?('Vehicle Registration')
```
memcmahon, line 23: Another option (guard): 

```
```

PR #625, opened 2025-02-18
Andrew Snyder

Andrew-Snyder627:
DMV solo project 2!


lib/facility.rb
```diff
@@ -1,14 +1,106 @@
...
-  def add_services(service)
+  def add_service(service) #simple typo error, resolved
     @services << service
   end
+
+  def register_vehicle(vehicle)
+    if @services.include?('Vehicle Registration') #checking if service is offered
+    else
+      return [] #should I change this to nil?
+    end
```
memcmahon, line 25: You could refactor to a 'guard clause'

```
```

PR #626, opened 2025-02-18
Katya Weicht Solo Project #2 Submission

KatyaW-01:


lib/facility.rb
```diff
@@ -1,14 +1,85 @@
...
+        @collected_fees += 200
+      else
+        vehicle.plate_type = :regular
+        @collected_fees += 100
+      end
+      @registered_vehicles << vehicle
+    end
+  end 
+
+  def administer_written_test(registrant)
```
memcmahon, line 45: You could use a guard statement:

```
    return false if !@services.include?("Written Test")
    
    if registrant.permit? == true && registrant.age >= 16
    else
      false
    end
 ```

PR #630, opened 2025-02-19
Maia Bard

MaiaKBard:


spec/registrant_spec.rb
mikedao, line 1: This is good, though for attributes I might have tested more than one example.

lib/registrant.rb
mikedao, line 1: Instead of making license data accessbile, I'd have made a method to change the data. It's dangerous to have things as accessors. I would have a method like this:

```
def take_written_test
  @license_data[:written] = true
end
```

and call this method from the method in the facility class.

mikedao (commented):
Overall well done. Code is clean. Like I said earlier, think about how you can separate out your code so that you're not relying on attr_accessors.

lib/facility_factory.rb
mikedao, line 1: Lambdas - FANCY!

PR #633, opened 2025-02-19
Calvin Suiter

cdsuit00:


lib/dmv.rb
```diff
@@ -1,6 +1,11 @@
 class Dmv
+  attr_reader :name, :address, :phone, :services, :facilities
 
-  def initialize
+  def initialize(details ={})
+    @name = details[:name]
+    @address = details[:address]
+    @phone = details[:phone]
+    @services = []
```
memcmahon, line 8: These attributes should exist on a Facility! A DMV only needs one attribute `@facilities`

lib/dmv.rb
```diff
@@ -9,7 +14,7 @@ def add_facility(facility)
   end
 
   def facilities_offering_service(service)
-    @facilities.find do |facility|
+    @facilities.select do |facility|
```
memcmahon, line 17: Great use of `select`!

lib/facility.rb
```diff
@@ -1,14 +1,14 @@
 class Facility
   attr_reader :name, :address, :phone, :services
 
-  def initialize(name, address, phone)
-    @name = name
-    @address = address
-    @phone = phone
+  def initialize(facility_details)
```
memcmahon, line 4: Yes!  Great job debugging this class :)

lib/registrant.rb
```diff
@@ -0,0 +1,19 @@
+class Registrant
+    attr_reader :name, :age, :permit, :license_data
+
+    def initialize(name, age, permit = false)
+    @name = name
+    @age = age
+    @permit = permit
+    @license_data = {written: false, license: false, renewed:false}
```
memcmahon, line 8: Keep an eye on your indentation here - we want code that is inside of a method to be indented one level:

```
  @name = name
  @age = age
  @permit = permit
  @license_data = {written: false, licence: false, renewed: false}
end
```

lib/vehicle.rb
```diff
@@ -5,14 +5,16 @@ class Vehicle
               :year,
               :make,
               :model,
-              :engine
+              :engine,
+              :registration_date
```
memcmahon, line 9: great!

spec/registrant_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    @registrant_1 = Registrant.new('Bruce', 18, true)
+    @registrant_2 = Registrant.new('Penny', 15)
+    end
+    
+    describe '#initialize' do
+      it 'can initialize' do
+      expect(@registrant_1).to be_a Registrant
+      expect(@registrant_2).to be_a Registrant
+      end
+  end
```
memcmahon, line 14: keep an eye on your indentation here.

lib/registrant.rb
```diff
@@ -0,0 +1,19 @@
...
+    end
+
+    def permit?
+        @permit
+    end
+
+    def earn_permit
+        @permit = true
+    end
```
memcmahon, line 17: These methods are exactly correct!

lib/registrant.rb
```diff
@@ -0,0 +1,19 @@
+class Registrant
+    attr_reader :name, :age, :permit, :license_data
```
memcmahon, line 2: Because we have a `permit?` method defined that is intended to return the value of our instance variable `@permit`, we don't need an attr_reader for `permit`.  The attr_reader accomplishes the same thing as the method we defined, so its a bit of duplication.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  it 'has the correct age' do
+    expect(@registrant_1.age).to eq(18)
+    expect(@registrant_2.age).to eq(15)
+  end
+
+  it 'has a permit' do
+    expect(@registrant_1.permit).to eq(true)
+    expect(@registrant_2.permit).to eq(false)
+  end
+
```
memcmahon, line 30: We also want to make sure we have a test for `permit?` - it would look like your 'it has a permit' test, but would verify that the `permit?` method is working.

spec/registrant_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  it 'has license_data' do
+    expect(@registrant_1.license_data).to eq({written: false, license: false, renewed:false})
+    expect(@registrant_2.license_data).to eq({written: false, license: false, renewed:false})
+  end
+
+  it 'can earn permit' do
+    expect(@registrant_2.earn_permit).to eq(true)
```
memcmahon, line 37: When we have a command method like `earn_permit` (a method that causes a change to our objects), we generally will test that method by using query methods to confirm the change.  It would look like:

```
it 'can earn permit' do
  @registrant_2.earn_permit
  
  expect(@registrant_2.permit?).to eq(true)
end
```

We could even go one step further and verify that the attribute starts with one value, and changes to another value:

```
it 'can earn permit' do
  expect(@registrant_2.permit?).to eq(false) #starting value
  
  @registrant_2.earn_permit #method that should change an attibute
  
  expect(@registrant_2.permit?).to eq(true) #ending value (after the method is called)
end
```

